PR #2, opened 2020-10-14
shramkoweb: add js-dom task

Shramkoweb:

### [js-dom task](https://github.com/kottans/frontend/blob/master/tasks/js-dom.md)

[Demo](https://shramkoweb.github.io/kottans-dom/) |
[Code base](https://github.com/Shramkoweb/kottans-dom)

Please, review.

### DIFF
--- a/submissions/shramkoweb/js-dom/index.html
+++ b/submissions/shramkoweb/js-dom/index.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="main.css">
+    <title>Dom - Kottans</title>
+  </head>
+  <body>
+    <main class="main">
+      <aside class="aside">
+        <ul class="frameworks-list"></ul>
+      </aside>
+
+      <section class="content"></section>
+    </main>
+    <script src="main.js"></script>
+  </body>
+</html>
--- a/submissions/shramkoweb/js-dom/main.css
+++ b/submissions/shramkoweb/js-dom/main.css
@@ -1,0 +1,73 @@
+html {
+    box-sizing: border-box;
+}
+
+*,
+*::before,
+*::after {
+    box-sizing: inherit;
+}
+
+body {
+    padding: 0;
+    font-size: 14px;
+    line-height: 20px;
+    font-family: "Open Sans", "Arial", sans-serif;
+}
+
+img {
+    max-width: 100%;
+    display: block;
+    height: auto;
+}
+
+.main {
+    display: grid;
+    grid-template-columns: 200px 1fr;
+}
+
+.content {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.frameworks-list {
+    display: flex;
+    flex-direction: column;
+    list-style: none;
+    padding: 0;
+}
+
+.content img {
+    max-height: 200px;
+}
+
+.frameworks-list li {
+    margin-bottom: 10px;
+}
+
+.button {
+    padding: 10px 20px;
+    background-color: transparent;
+    color: #1890ff;
+    border: 1px solid #1890ff;
+    border-radius: 2px;
+    box-shadow: 0 2px 0 rgba(0, 0, 0, .045);
+    outline: 0;
+    font-weight: 700;
+    width: 100%;
+    user-select: none;
+}
+
+.button.active {
+    color: #fff;
+    background: #096dd9;
+    border-color: #096dd9;
+}
+
+.button:focus {
+    color: #fff;
+    background: #40a9ff;
+    border-color: #40a9ff;
+}
--- a/submissions/shramkoweb/js-dom/main.js
+++ b/submissions/shramkoweb/js-dom/main.js
@@ -89,11 +89,17 @@ const init = () => {
 
     const handleListItemClick = (evt) => {
         const { target } = evt;
-        if (target.tagName !== 'BUTTON') {
+        const currentID = parseInt(target.id, 10);
+        const isNeedRerender = currentID !== state.activePageID;
+
+        if (
+            target.tagName !== 'BUTTON' ||
+            !isNeedRerender
+        ) {
             return;
         }
 
-        state.activePageID = parseInt(target.id, 10);
+        state.activePageID = currentID;
         render(state);
     }
     frameworksListElement.addEventListener('click', handleListItemClick);
--- a/submissions/shramkoweb/js-dom/main.js
+++ b/submissions/shramkoweb/js-dom/main.js
@@ -2,73 +2,41 @@ const frontendFrameworks = [
     {
         id: 0,
         name: 'React',
-    },
-    {
-        id: 1,
-        name: 'Vue',
-    },
-    {
-        id: 2,
-        name: 'Angular',
-    },
-    {
-        id: 3,
-        name: 'Django',
-
-    },
-    {
-        id: 4,
-        name: 'Ruby',
-
-    },
-    {
-        id: 5,
-        name: 'Flask',
-
-    },
-    {
-        id: 6,
-        name: 'Express',
-
-    },
-    {
-        id: 7,
-        name: 'JQuery',
-
-    },
-];
-
-const imageURLs = [
-    {
-        id: 0,
         url: 'https://upload.wikimedia.org/wikipedia/commons/a/a7/React-icon.svg',
     },
     {
         id: 1,
+        name: 'Vue',
         url: 'https://upload.wikimedia.org/wikipedia/commons/9/95/Vue.js_Logo_2.svg',
     },
     {
         id: 2,
+        name: 'Angular',
         url: 'https://upload.wikimedia.org/wikipedia/commons/c/cf/Angular_full_color_logo.svg',
     },
     {
         id: 3,
+        name: 'Django',
         url: 'https://upload.wikimedia.org/wikipedia/commons/7/75/Django_logo.svg',
     },
     {
         id: 4,
+        name: 'Ruby',
         url: 'https://upload.wikimedia.org/wikipedia/commons/f/f1/Ruby_logo.png',
     },
     {
         id: 5,
+        name: 'Flask',
         url: 'https://upload.wikimedia.org/wikipedia/commons/3/3c/Flask_logo.svg',
     },
     {
         id: 6,
+        name: 'Express',
         url: 'https://upload.wikimedia.org/wikipedia/commons/6/64/Expressjs.png',
     },
     {
         id: 7,
+        name: 'JQuery',
         url: 'https://upload.wikimedia.org/wikipedia/commons/8/81/JQuery_logo_text.svg',
     },
 ];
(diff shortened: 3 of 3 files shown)


madmaxWMFU:
Hi @Shramkoweb! Good job.
Please, tell me why you have two different objects `frontendFrameworks ` and `imageURLs`, maybe it is better to unite them because the field `id` is the same

Shramkoweb:
@madmaxWMFU I think it's better to have as flat data structure as possible, so it's easier to work with them.

Putting it all together, the data binds more strongly, so this option is considered worse. Something like normalized data.

MaxKalashnyk (changes requested):
The code looks good to me! It's cool that you operate the `state` of your app. But I've found some performance issues in your code. It will be nice if you fix them 🙂
Let me know if you have some ideas on how you can deal with it (I left a comment under your `render` function). I'm open to discussing it.

submissions/shramkoweb/js-dom/main.js
```diff
@@ -0,0 +1,104 @@
...
+    const { url } = getItemByID(frontendFrameworks, activePageID);
+    const contentElement = document.querySelector('.content');
+    const image = document.createElement('img');
+    image.src = url;
+
+    clearHTML(contentElement);
+    contentElement.append(image);
+};
+
+const render = (state) => {
```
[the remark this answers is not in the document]
Shramkoweb: When I didn't think about the performance. thanks =)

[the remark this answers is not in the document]
babayK0: Maybe @MaxKalashnyk talks about that renderListItems is redundant in render function. You need render list of buttons only once, when init your app. Then you must rerender only content (image in your  case)

submissions/shramkoweb/js-dom/main.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    const handleListItemClick = (evt) => {
+        const { target } = evt;
+        if (target.tagName !== 'BUTTON') {
+            return;
+        }
+
+        state.activePageID = parseInt(target.id, 10);
+        render(state);
```
babayK0, line 103: Here you can call just renderContent function i guess

submissions/shramkoweb/js-dom/main.js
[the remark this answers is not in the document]
Aldegid: I Agree with @babayK0 , i think you must call renderListItems only once (first render).

submissions/shramkoweb/js-dom/main.js
```diff
@@ -0,0 +1,104 @@
...
+    const frameworksListElement = document.querySelector('.frameworks-list');
+    clearHTML(frameworksListElement);
+    const fragment = document.createDocumentFragment();
+    items.forEach((item) => fragment.append(renderListItem(item, activeID)));
+
+    frameworksListElement.append(fragment);
+};
+
+const renderContent = (activePageID) => {
+    const { url } = getItemByID(frontendFrameworks, activePageID);
```
babayK0, line 69: What will be in case when your getItemByID function will return undefined value ?
I think you need fallback for unpredictable situations.
In this app you don't need to rewrite, but it is up to you.

submissions/shramkoweb/js-dom/main.js
```diff
@@ -0,0 +1,104 @@
...
+const getItemByID = (items, id) => items.find(item => item.id === id);
+const clearHTML = node => node.innerHTML = '';
+
+const renderListItem = ({ name, id }, activeID) => {
+    const item = document.createElement('li');
+    const button = document.createElement('button');
+    button.innerText = name;
+    button.type = 'button';
+    button.id = id;
+    button.className = activeID === id ? 'active button' : 'button';
```
babayK0, line 53: It may be better to separate render of buttons list and toggling of active class. It is also better to use [classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) instead of className

babayK0 (changes requested):

submissions/shramkoweb/js-dom/main.js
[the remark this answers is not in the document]
Shramkoweb: @babayK0 @Aldegid in this task i use MVC model.
And when my state(model) change i need rerender my app whit new state(new activePageID).

I think in this case its ok. When we move renderListItems out each render func its will be not MVC

submissions/shramkoweb/js-dom/main.js
Shramkoweb: Answered above

submissions/shramkoweb/js-dom/main.js
Shramkoweb: If you do the activation from the render, you will have to work with the DOM directly bypassing the state, which is not acceptable in MVC

Shramkoweb: Render knows only about the state, the state changes in the handlers (controller), rendering is based on the state, no information is stored or read in the HOUSE, otherwise the essence of MVC which I use here is lost

submissions/shramkoweb/js-dom/main.js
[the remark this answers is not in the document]
MaxKalashnyk: @Shramkoweb, it's really cool that you tried the MVC pattern. I respect your decision and the code you wrote. But I guess it will be better to use it in more complex apps. I recommend looking at this course:
https://www.udemy.com/course/the-complete-javascript-course/?admitad_uid=6151191e959be7d1516e751c1c354aae
(maybe you know about it as well)

It seems that using the MVC pattern here is quite overengineering.
This task is pretty similar to the tabs UI element and I'm not sure that the easiest way of how to design it will be the MVC pattern.
Unfortunately, I didn't get any of your suggestions on how to simplify your state and rendering logic.

Let's simplify it to the toggling of the ID of the framework and just toggle the src of the image of the selected framework.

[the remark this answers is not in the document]
Shramkoweb: U really think I need rewrite all my app, because rerender button list.
I think I correct make my app, and fix issues.
In MVC pattern normal to rerender some DOM on state change.
I agree it's overwhelming, but I try MVC and make it's correct.

submissions/shramkoweb/js-dom/index.html
```diff
@@ -0,0 +1,20 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
```
MaxKalashnyk, line 5: I guess `content="width=device-width"` will be enough here. HTML validators sometimes give warnings about it

submissions/shramkoweb/js-dom/index.html
```diff
@@ -0,0 +1,20 @@
...
+    <link rel="stylesheet" href="main.css">
+    <title>Dom - Kottans</title>
+  </head>
+  <body>
+    <main class="main">
+      <aside class="aside">
+        <ul class="frameworks-list"></ul>
+      </aside>
+
+      <section class="content"></section>
```
MaxKalashnyk, line 16: It seems your `section` doesn't have any title. It can cause warnings during the validation of your layout

submissions/shramkoweb/js-dom/main.css
```diff
@@ -0,0 +1,73 @@
+html {
```
MaxKalashnyk, line 1:
```suggestion
*,
*::before,
*::after {
     box-sizing: border-box;
}
```

submissions/shramkoweb/js-dom/main.css
```diff
@@ -0,0 +1,73 @@
...
+
+.content img {
+    max-height: 200px;
+}
+
+.frameworks-list li {
+    margin-bottom: 10px;
+}
+
+.button {
```
MaxKalashnyk, line 50: Why do you use `user-select: none;` here? I'd recommend adding `cursor: pointer;` for this button as well.

submissions/shramkoweb/js-dom/main.js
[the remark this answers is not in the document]
MaxKalashnyk: @Shramkoweb, it's up to you. This task doesn't require to build it using the MVC pattern. 
I'm here to help to simplify this code and make it more readable for other people.
Sorry, I can't approve this code only because you used the MVC. 
Thanks for your understanding.

submissions/shramkoweb/js-dom/main.css
Shramkoweb: Because I don't want select text in button. 
It about user select property

submissions/shramkoweb/js-dom/main.js
```diff
@@ -0,0 +1,110 @@
...
+const init = () => {
+    const frameworksListElement = document.querySelector('.frameworks-list');
+    const state = {
+        frameworks: frontendFrameworks,
+        activePageID: 0,
+    }
+
+    const handleListItemClick = (evt) => {
+        const { target } = evt;
+        const currentID = parseInt(target.id, 10);
```
MaxKalashnyk, line 92: Take into account the performance of the parsing strings into integer methods:
https://i.stack.imgur.com/jtTwP.png

submissions/shramkoweb/js-dom/index.html
Shramkoweb: Yes. It's not error. 
In u opinion I need add hidden title ? Or not use semantics correct in this case section ?

submissions/shramkoweb/js-dom/index.html
Shramkoweb: Can I show when I have problem with this case ?

submissions/shramkoweb/js-dom/index.html
MaxKalashnyk: https://validator.w3.org/nu/?doc=https%3A%2F%2Fshramkoweb.github.io%2Fkottans-dom%2F

submissions/shramkoweb/js-dom/main.css
Shramkoweb: I think my is better. Because inheritance is more flexible in feature.
If I wrong can u explain why you variant better ?

submissions/shramkoweb/js-dom/index.html
Shramkoweb: It's not error. Only warn.
In this case section is more semantics correct.

submissions/shramkoweb/js-dom/main.js
Shramkoweb: And so ? What is option u prefered ?

babayK0 (approved):

submissions/shramkoweb/js-dom/index.html
yeromin: let me add smt as well. 
I also wanted to highlight that semantically there's no sense in <section> if it doesn't have a Header. It's better to add a hidden title. 
Correct, it's not an error, it's a recommendation. But, for example, Vadim Makeev usually says "If you don't want to add a hidden title, let it be <div>, not <section>".
@Shramkoweb 

ref: http://html5doctor.com/avoiding-common-html5-mistakes/
Screenshot 2020-10-16 at 10 44 10

Shramkoweb: @yeromin ok, thx add hidden title =)

submissions/shramkoweb/js-dom/main.js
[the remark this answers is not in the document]
zonzujiro: I will resolve this one, if you don't mind.

@MaxKalashnyk thank you for detailed explanation and link to the course. This is a great peace of knowledge which should be considered in the future.
@Shramkoweb nicely done with separating of concerns between rendering and state management!

zonzujiro:
Ok. I believe this PR becomes a little bit bloated for such simple task :D

I will merge it.

Thank you all for great job! 💪

PR #31, opened 2020-10-27
Add: DOM API Task  from andrewklmn

andrewklmn:
### DOM API Task

[Demo](https://andrewklmn.github.io/frontend-2021-homeworks/submissions/andrewklmn/task_js_dom/) |
[Code base](https://github.com/andrewklmn/frontend-2021-homeworks/tree/task_js_dom/submissions/andrewklmn/task_js_dom)

Please, review.


submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  let elem = event.target;
+  let aside = document.querySelector('aside');
+  let wraper = document.querySelector('.wraper');
+  
+
+  if (elem.innerHTML=="«") {
+    elem.innerHTML = "☰";
+    elem.style.left = "0";
+    aside.style.width = "0";
+    wraper.style.marginLeft = "20px";
```
madmaxWMFU, line 65: You should create the rules in CSS and change their  
`classList.toggle()` or use `classList.add() / classList.remove()`

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+    a = document.createElement("a");
+    a.innerHTML = item.menu;
+    a.className = "menu-item";
+    a.href = "javascript:void(0)";
+    
+    if (index==state.activeIndex) {
+      a.classList.toggle("active");
+    };
+    
+    a.addEventListener('click',menuItemClickListener);
```
madmaxWMFU, line 114: You can add the event listener only for the main parent element. Use an [event delegation](https://www.educative.io/edpresso/what-is-event-delegation-in-javascript)

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+    method: document.querySelector('p.method').innerHTML,
+  });
+  state.activeIndex = 0;
+
+  renderPage(state);
+  
+  const vw = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  if (vw < 820) {
+    document.querySelector('.btn-toogle-menu').click(); 
+  }
```
madmaxWMFU, line 207: This code is duplicated, write a function for this.

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  // Add first menu item content from default HTML layout
+  state.items.unshift({
+    menu: "All types:",
+    header: document.querySelector('article > h2').innerHTML,
+    image: document.querySelector('article > img').src,
+    description: document.querySelector('p.description').innerHTML,
+    habitat: document.querySelector('span.habitat').innerHTML,
+    method: document.querySelector('p.method').innerHTML,
+  });
+  state.activeIndex = 0;
```
madmaxWMFU, line 200: You can write the value as constant in your object `state`

madmaxWMFU (changes requested):
Good job

See some comments below

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  renderMenu(state);
+  renderArticle(state.items[state.activeIndex]);
+
+};
+
+// start App when DOM is loaded
+document.addEventListener('DOMContentLoaded', (event)=>{
+
+  // Add first menu item content from default HTML layout
+  state.items.unshift({
```
madmaxWMFU, line 192: Why did you add this object to the array?
You can initially write everything to an array and display information by specifying the desired index. And not publish this information like as static in HTML

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+    ul.appendChild(li);
+  });
+
+  fragment.appendChild(ul);
+  prevUl.remove();
+  nav.appendChild(fragment);
+};
+
+const showTextContent = ()=>{
+  document.body.style.color = "black";
```
madmaxWMFU, line 125: Use style in CSS

madmaxWMFU (changes requested):

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+    let li = document.createElement("li");   
+    li = document.createElement("li");
+
+    let a = document.createElement("a");
+    li.id = index;
+      
+    a = document.createElement("a");
+    a.innerHTML = item.menu;
+    a.className = "menu-item";
+    a.href = "javascript:void(0)";
```
madmaxWMFU, line 108: Please, can you explain why you set this parameter?

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: I did it because Googlebot needs something to read for its search database. It can read it as a static HTML page without JS instead of an empty page. But I can move this content to the initial state object.
Should I move it?

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  return false;
+};
+
+const renderMenu = (state)=>{
+  
+  let nav = document.querySelector('nav');
+  let prevUl = document.querySelector('ul');
+  let fragment = document.createDocumentFragment();
+  let ul = document.createElement("ul");
+
```
mustbefail, line 97: Use class reference instead of tag reference. You can have several ul, li, nav, a etc. on the page.

submissions/andrewklmn/task_js_dom/app.js
MaxKalashnyk: @madmaxWMFU, it's one of the techniques for preventing page reload after clicking on this link. I think it's ok

submissions/andrewklmn/task_js_dom/app.css
```diff
@@ -0,0 +1,243 @@
+body {
+  padding: 0;
+  margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+ul {
+  display: flex;
```
mustbefail, line 8: Better to use classes than direct tag reference

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+      image: "img/digger.jpg",
+      description: "Sadly, this is one of the more common types of L.A. women, but luckily they’re easy to spot. The twist here is that you don’t necessarily have to have a lot of money, you just have to have the potential to make it. So if you’re a junior agent or work in the mailroom at an entertainment law firm, she’ll invest her time in you. But if you don’t come through, buddy, you are screwed. All she wants is the house on the hill and the fancy car, so if you’re just looking for a trophy wife, then take your pick.",
+      habitat: "Beverly Hills, Hollywood Hills, any Hills, really.",
+      method: "How to attract her: What are you, an idiot? Just flash your cash!",
+    },
+  ],
+  activeIndex: null,
+};
+
+const menuToogleClickListener = (event)=>{
```
MaxKalashnyk, line 55: This function name has a typo. Also, I'd recommend you to simplify the naming of it. Maybe, `toggleMenu` will ok.

mustbefail (commented):
Great job. See some comments below

submissions/andrewklmn/task_js_dom/app.js
madmaxWMFU: Agree, this is just for my understanding🙂

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  let elem = event.target;
+  let aside = document.querySelector('aside');
```
MaxKalashnyk, line 57: Let's use here `const` instead of `let` for these variables. See the difference [here](https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/).
Also, I'd recommend you to move these variables to the beginning of this file where you declared your constants.
Because now you querying your DOM and create a new variable on each click of the button. It can reduce the performance of your app

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  return false;
+};
+
+const menuItemClickListener = (event)=>{
+  let elem = event.target;
+  state.activeIndex = elem.parentNode.id;
+  renderPage(state);
+
+  const vw = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  if (vw < 820) {
```
MaxKalashnyk, line 83: Create for this 'magic number' some variable and move it to the constants

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+const renderMenu = (state)=>{
+  
+  let nav = document.querySelector('nav');
+  let prevUl = document.querySelector('ul');
+  let fragment = document.createDocumentFragment();
+  let ul = document.createElement("ul");
+ 
+ state.items.forEach((item,index)=>{
+    let li = document.createElement("li");   
+    li = document.createElement("li");
```
MaxKalashnyk, line 100: Duplicate of `li` creation

submissions/andrewklmn/task_js_dom/app.js
madmaxWMFU: It only creates this tag here

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+
+    let a = document.createElement("a");
+    li.id = index;
+      
+    a = document.createElement("a");
```
MaxKalashnyk, line 105: Why did you reassign `a` variable here?

submissions/andrewklmn/task_js_dom/app.js
mustbefail: This is my mistake, the comment is on lines 93 and 94. Referring to nav and ul.

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
+const state = {
+  items: [
+    {
+      menu: 'Ingénue',
```
MaxKalashnyk, line 4: It's not a blocker, but I'd recommend renaming some of the properties of the objects in the `items` array for making your data easy to understanding.
```suggestion
      menuTitle: 'Ingénue',
      title: "The Ingénue",
      image: "img/beautiful-blonde.jpg",
      description: "...",
      location: "Santa Monica pier",
      attractionMethod: "..."
```

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,209 @@
...
+  fragment.appendChild(ul);
+  prevUl.remove();
+  nav.appendChild(fragment);
+};
+
+const showTextContent = ()=>{
+  document.body.style.color = "black";
+}
+
+const renderArticle = (content)=> {
```
MaxKalashnyk, line 128: To my mind, there are too many DOM operations here. How about adding these all content to your HTML and then change only text content from the `state` object. Let me know if you got me. I'm ready for help to discuss it if you'll have some questions.

MaxKalashnyk (changes requested):
Nice work! Look at my comments, please

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: > Please, can you explain why you set this parameter?

The void operator can be used to return undefined and prevent load other content. 
I took it here: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)
But you are right. They recommend to use event hadlers...

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -22,20 +22,34 @@ const state = {
...
-  let ul = document.createElement("div");
+  let ul = document.createElement("ul");
+
   state.items.forEach((item,index)=>{
     let li = document.createElement("li");
-    li.innerHTML = item.menu;
+    let a = document.createElement("a");
     li.id = index;
+    a.innerHTML = item.menu;
+    a.className = "menu-item";
```
A-Ostrovnyy, line 103: Use `ClassList` API for manipulating classes [Read more](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList)

submissions/andrewklmn/task_js_dom/app.js
madmaxWMFU: You may be right, but there are sites that are completely generated on the server
And they use other techniques for the Googlebot

madmaxWMFU (changes requested):

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: I have created  function for getting viewport width [in this commit](https://github.com/kottans/frontend-2021-homeworks/pull/31/commits/08c4460b4dea94c5aeda59e156255b8505340855)

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,195 @@
...
+
+const menuContainer = document.querySelector('.menu-container');
+const wraper = document.querySelector('.wraper');
+const menuUl = document.querySelector('.menu');
+
+const getViewWidth = () => Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+
+const toggleMenu = (event)=>{
+  
+  event.target.innerHTML = (event.target.innerHTML == "«") ? "☰":"«";
```
madmaxWMFU, line 63: In this situation, I would use CSS.
```
.btn-toogle-menu:before {
  position: absolute;
  content: "«";
  top: 10px;
  left: 10px;
}

.btn-toogle-menu:after {
  content: "";
}

.burger:before {
  content: "";
}

.burger:after {
  content: "☰";
}
```

madmaxWMFU (changes requested):

submissions/andrewklmn/task_js_dom/app.js
madmaxWMFU: If you try this don't forget delete the line `a.innerHTML = "&laquo;";`

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,206 @@
...
+  elem.classList.toggle("burger");
+  wraper.classList.toggle("wide");
+
+  event.stopPropagation();
+  return false;
+};
+
+const menuItemClickListener = (event)=>{
+
+  if (event.target.tagName=='A') {
```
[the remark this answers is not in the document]
andrewklmn: It really works!...

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: Cool solution! 
Thanks!!!

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,206 @@
...
+  
+  aside.classList.toggle("opened") ;
+  elem.classList.toggle("burger");
+  wraper.classList.toggle("wide");
+
+  event.stopPropagation();
+  return false;
+};
+
+const menuItemClickListener = (event)=>{
```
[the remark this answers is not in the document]
madmaxWMFU: [Destructuring](https://learn.javascript.ru/destructuring)

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+      location: "Beverly Hills, Hollywood Hills, any Hills, really.",
+      attractionMethod: "How to attract her: What are you, an idiot? Just flash your cash!",
+    },
+  ],
+  activeIndex: 0,
+};
+
+const menuContainer = document.querySelector('.menu-container');
+const menuUl = document.querySelector('.menu');
+const wraper = document.querySelector('.wraper');
```
MaxKalashnyk, line 65: It's a typo here
```suggestion
const wrapper = document.querySelector('.wrapper');
```

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+const description = document.querySelector('.description');
+const locationArea = document.querySelector('.habitat');
+const attractionMethod = document.querySelector('.method');
+
+const getViewWidth = () => Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+const maxAutoHideMenuWidth = 820;
+
+const toggleMenu = (event)=>{
+  
+  event.target.classList.toggle("burger");
```
MaxKalashnyk, line 77: Maybe it's better to toggle `closed` class. Because today you have the 'burger' icon, tomorrow you'll decide to replace the burger icon with another. So let's make it more generic.
```suggestion
  event.target.classList.toggle("closed");
```

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+  
+  menuContainer.classList.toggle("opened") ;
+  wraper.classList.toggle("wide");
+
+  event.stopPropagation();
```
MaxKalashnyk, line 82: Why do you use `event.stopPropagation();` here?

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+  return false;
```
MaxKalashnyk, line 83: Do you really need this `return false;` here?

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+const menuItemClickListener = (event)=>{
+
+  if (event.target.matches('.menu-item')) {
+    state.activeIndex = event.target.parentNode.id;
+    renderPage(state);
+  
+    if (getViewWidth() < maxAutoHideMenuWidth) {
+      document.querySelector('.btn-toogle-menu').click(); 
+    }
+    event.stopPropagation();
```
MaxKalashnyk, line 95: The same as in the previous comment about the `event.stopPropagation();`

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+  };
+
+  return false;
```
MaxKalashnyk, line 98: See the previous comment about it

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+  });
+};
+
+const showTextContent = ()=>{
+  document.body.classList.remove("text-white");
+}
+
+const renderArticle = (content)=> {
+
+  document.body.classList.add("text-white");
```
MaxKalashnyk, line 129: Why do you need to add this class here? Is it an imitation of the loading data process?

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+  state.items.forEach((item,index)=>{
+    const li = document.createElement("li");   
+    li.id = index;
+
+    const a = document.createElement("a");
+    a.innerHTML = item.menuTitle;
+    a.classList.add("menu-item");
+    a.href = "javascript:void(0)";
+    
+    if (index==state.activeIndex) {
```
MaxKalashnyk, line 114: Better to use `===` operator for checking type `state.activeIndex` as well

submissions/andrewklmn/task_js_dom/app.js
```diff
@@ -0,0 +1,165 @@
...
+  a.addEventListener('click', toggleMenu);
+  header.parentNode.insertBefore(a,header);
+  
+
+  renderPage(state);
+
+  menuUl.addEventListener('click',menuItemClickListener);
+  image.addEventListener('load',showTextContent);
+
+  if (getViewWidth() < maxAutoHideMenuWidth) {
```
MaxKalashnyk, line 161: Maybe it's better to move this check to your styles? You can handle it in the media queries.

MaxKalashnyk (changes requested):
Good job! It looks much better! See my comments and we will ready to go ahead.

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: You are right!  I don't need it because I have no events on the bubbling stage to prevent their firing in this case...

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: I don't need it.  It is even a button now, and not an anchor with href...

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: You are right again! There is href="javascript: void (0)" which prevents page reload on click event ...

submissions/andrewklmn/task_js_dom/app.js
andrewklmn: Yes. This is just a little animation for the case when the image loading takes a longer time. The text disappears before and appears after the image loading.

MaxKalashnyk: Got it. Maybe it will be better to show/hide some loader instead of adding white color for the texts.

madmaxWMFU (approved):
You listened to my recommendations

MaxKalashnyk (approved):
LGTM! Thank you for your patience and for striving to improve your code.

PR #34, opened 2020-10-27
dom-api task done

ses3332002:
### DOM API 

[Demo](https://ses3332002.github.io/kottans-frontend/dom/) |
[Code base](https://github.com/ses3332002/kottans-frontend/tree/main/dom)

Please, review.


submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+      "bgColor": 33
+    }
+  ];
+
+  // definition of page elements
+  var doc = document.querySelector("body");
+  var navTab = document.querySelector(".navigation");
+  var infoTab = document.querySelector(".info");
+
+  // filling the menu block
```
zonzujiro, line 230: Instead of such comments you can create function `createMenuBlocks` and then use it.

Comments should describe *why* something is happening, not *what*.
What happening you will understand from code :)

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+        "Slave to the Rhythm",
+        "Do You Know Where Your Children Are",
+        "Blue Gangsta",
+        "Xscape"
+      ],
+      "bgColor": 33
+    }
+  ];
+
+  // definition of page elements
```
zonzujiro, line 225: Redundant comment

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+    navTabBtn.innerText = alboms[i].title;
+    navTab.append(navTabBtn);
+  };
+  var navTabBtns = document.querySelectorAll(".navigation_btn");
+  var navTabBtnsArray = Array.prototype.slice.call(navTabBtns);
+
+  // listening for click events for buttons
+  navTab.addEventListener("click", menuHandler);
+
+  function menuHandler(e) {
```
zonzujiro, line 245: Violation of naming convention. Name of functions should be verbs - `doSomething`, `createThat`, etc. This increases readability of the code

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+  navTab.innerHTML = "";
+  var navTabBtn;
+  for (var i = 0; i < alboms.length; i++) {
+    navTabBtn = document.createElement("button");
+    navTabBtn.classList.add("navigation_btn");
+    navTabBtn.innerText = alboms[i].title;
+    navTab.append(navTabBtn);
+  };
+  var navTabBtns = document.querySelectorAll(".navigation_btn");
+  var navTabBtnsArray = Array.prototype.slice.call(navTabBtns);
```
zonzujiro, line 240: `Array.prototype.slice.call` - no need for such dirty things.

You can use [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from).

And I don't see any reason of why you make copy. So, why do you need shallow copy of these elements?

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+
+  // listening for click events for buttons
+  navTab.addEventListener("click", menuHandler);
+
+  function menuHandler(e) {
+    var selectedBtnIndex = navTabBtnsArray.indexOf(e.target);
+    var infoList;
```
zonzujiro, line 247: Just a reminder:

- you should use `let`. It is a 2020 :)
- no need in declaring variables outside of the loop. Just declare them inside of it. 

You will not affect performance and will keep all code only in your loop.

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+    var infoLine;
+    infoTab.innerHTML = "";
+    doc.style.setProperty("--background_color_h", alboms[selectedBtnIndex].bgColor);
+    infoTab.style.backgroundImage = "url(img/" + alboms[selectedBtnIndex].cover + ".jpg)";
+    infoList = document.createElement("ul");
+    for (var i = 0; i < alboms[selectedBtnIndex].songs.length; i++) {
```
zonzujiro, line 253: Just use `Array#forEach`

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+    navTab.append(navTabBtn);
+  };
+  var navTabBtns = document.querySelectorAll(".navigation_btn");
+  var navTabBtnsArray = Array.prototype.slice.call(navTabBtns);
+
+  // listening for click events for buttons
+  navTab.addEventListener("click", menuHandler);
+
+  function menuHandler(e) {
+    var selectedBtnIndex = navTabBtnsArray.indexOf(e.target);
```
zonzujiro, line 246: Pls, dnt use shrtcuts in the nms of vrbls.

Write `Button` 😄

zonzujiro (changes requested):

submissions/ses3332002/dom/index.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <script src="script.js"></script>
```
MaxKalashnyk, line 8: It's better to place the `script` tag at the end of the `body` tag. It can help to load your page faster

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
+"use strict";
+window.onload = function() {
+
+// data for displaing
+  const alboms = [
```
MaxKalashnyk, line 5: Typo here. 
```suggestion
  const albums = [
```

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
+"use strict";
+window.onload = function() {
+
+// data for displaing
```
MaxKalashnyk, line 4: This comment is redundant here. Remove it, please

submissions/ses3332002/dom/index.html
MaxKalashnyk, line 8: It's better to place the `script` tag at the end of the `body` tag. It can help to load your page faster

submissions/ses3332002/dom/script.js
MaxKalashnyk, line 4: This comment is redundant here. Remove it, please

submissions/ses3332002/dom/script.js
MaxKalashnyk, line 5: It's a typo here.
```suggestion
  const albums = [
```

MaxKalashnyk (changes requested):
Good job! See my comments below, please

submissions/ses3332002/dom/script.js
ses3332002: Ok. I'll remake this

submissions/ses3332002/dom/script.js
ses3332002: I need to define which button is clicked while I listen to "click" event on parent block, so I found "Array.prototype.slice.call" to convert HTML-collection into array and use array's method "indexOf".
I'll remake it with Array.from.
I can make this code shorter, but I afraid it will be less "readable" for review. Is it OK?

submissions/ses3332002/dom/index.html
ses3332002: Well, this page is nearly empty :) and to give this page some content I need to JS load as fast as it can. So, I placed it to <head>.
But I can remake it if this is necessary.

submissions/ses3332002/dom/script.js
ses3332002: My mistake :)
I'll remake it.

zonzujiro: BTW, in order to avoid such mistakes, you can install a spell checker.
As an example for VSC you can use this - https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,260 @@
...
+        "Do You Know Where Your Children Are",
+        "Blue Gangsta",
+        "Xscape"
+      ],
+      "bgColor": 33
+    }
+  ];
+
+  // definition of page elements
+  var doc = document.querySelector("body");
```
zonzujiro, line 226: Just use `document.body`

submissions/ses3332002/dom/script.js
zonzujiro: `Array.from` will be more readable then `Array.prototype.slice.call`.

Also, you should avoid this `prototype.{method}.call` things in the future. Most such hacks are outdated and you can use other methods or build your code using another logic in which such things will be not needed.

ses3332002: Well, it was the first answer that Google gave to me, and it works :)

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: removed

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/index.html
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,257 @@
...
+      "Slave to the Rhythm",
+      "Do You Know Where Your Children Are",
+      "Blue Gangsta",
+      "Xscape"
+    ],
+    "bgColor": 33
+  }
+];
+
+let doc = document.body;
```
MaxKalashnyk, line 223: Why did you use `let` here?

MaxKalashnyk, line 223: Why did you use `let` here?

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,257 @@
...
+let navTab = document.querySelector(".navigation");
+let infoTab = document.querySelector(".info");
+let navTabButtonsArray;
+
+function fillNavTab() {
+  navTab.innerHTML = "";
+  let navTabButton;
```
MaxKalashnyk, line 230: No need to reassign this variable. Just use `const` and use it inside `forEach`

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,257 @@
...
+  for (let i = 0; i < albums.length; i++) {
+    navTabButton = document.createElement("button");
+    navTabButton.classList.add("navigation_button");
+    navTabButton.innerText = albums[i].title;
+    navTab.append(navTabButton);
```
MaxKalashnyk, line 235: I recommend you using `DocumentFragment` here. It helps to improve the performance of your rendering.
More info [here](https://developer.mozilla.org/ru/docs/Web/API/DocumentFragment).
You can create this fragment before the loop, then you can append `navTabButton` here. And then use `append` or `innerHTML` for placing fragment content into the `navTab`. It helps you make only one insert to the DOM element, but not `albums.length` times.

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,257 @@
...
+  };
+  navTabButtonsArray = Array.from(document.querySelectorAll(".navigation_button"));
+};
+
+function fillInfoTab(e) {
+  let selectedButtonIndex = navTabButtonsArray.indexOf(e.target);
+  let infoList;
```
MaxKalashnyk, line 242: The same variable reassigns as in the previous comment.

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,257 @@
...
+  let infoLine;
+  infoTab.innerHTML = "";
+  doc.style.setProperty("--background_color_h", albums[selectedButtonIndex].bgColor);
+  infoTab.style.backgroundImage = "url(img/" + albums[selectedButtonIndex].cover + ".jpg)";
+  infoList = document.createElement("ul");
```
MaxKalashnyk, line 247: You can move this `ul` to the HTML and generate `li` elements here. Also, I recommend using `DocumentFragment` here

MaxKalashnyk (changes requested):
Good job! See my comments below

submissions/ses3332002/dom/script.js
ses3332002: Do you mind "why let but not const"?
Nothing criminal, just a habit.
Well, I can remake this.

submissions/ses3332002/dom/script.js
ses3332002: Don't understand about "reassign". 
navTabButton is a variable that I use to create DOM elements. Even if I use "forEach", not "for" - I need to put different content in it, so it can't be a const.

submissions/ses3332002/dom/script.js
ses3332002: I' ll remake it to do only one DOM insertion.

submissions/ses3332002/dom/script.js
ses3332002: Sorry, still don't understand you.

submissions/ses3332002/dom/script.js
MaxKalashnyk: It seems I didn't explain my thoughts exactly.
I meant no need to create this variable outside the `for` loop. 
```suggestion
for (let i = 0; i < albums.length; i++) {
    const navTabButton = document.createElement("button");
```

submissions/ses3332002/dom/script.js
ses3332002: :) I use "ul" here exactly for making only one DOM insertion.
OK, I can remake this with DocumentFragment.

submissions/ses3332002/dom/script.js
MaxKalashnyk: I had the same habit)) but it's better to use their exact purpose

submissions/ses3332002/dom/script.js
MaxKalashnyk: I replied to you in the previous commit. Let me know if you still have some misunderstandings.

submissions/ses3332002/dom/script.js
ses3332002: Well, it's whole new treatment for me: instead of creating a variable one time and reuse it many times (let's call it "classic way") we re-create a constant for many times. Is it right? I've never seen such a treatment.

MaxKalashnyk: I think it's no big deal. You can leave it as you wrote before. It will work fine. But!
As you use the `navTabButton` variable only inside the loop, it's better to use `const`. It helps you to prevent accidental reassignment of this variable. Also, it helps to keep your code clean and easy to read.

ses3332002: Then I'll keep it in "classic way". 
May be, I'm too old :)
I promise not to reassign my variables, even accidentally.

zonzujiro: @ses3332002 I will add my opinion. It is not only about reassigning. In low level languages, as C/C++, this can be source of problems, if you will start counting assembler instructions, etc. But this is a JavaScript, which is quite high-level and compiler very smart - we have "warm up functions", we have JIT and all other magic necessary to load YouTube in 4k under 200 ms.

Also, you will not have a lot of data here. And in frontend in general. If you "really" will need to loop thru 10k elements, or 100 elements but 10-100 times, then "maybe" such approach will be reasonable. Especially for mobile phones. Or at backend.

But you should keep in mind, that you must avoid thing called "premature optimization". This approach in most cases usually creates bad code... and thats all it's result. In order to optimize something, you need to gather analytics in order to understand bottle necks. And then solve this bottle necks.

This is a just a variable and not a big problem. But try not to run before the train. :) Readability, clean code comes first.

submissions/ses3332002/dom/script.js
zonzujiro: Just a good practice. Usually you will have linter rule which will tell you to use `const` if you will not reassign variable.

submissions/ses3332002/dom/script.js
zonzujiro: You can use ES6 strings also. Something like:

```javascript
const htmlContent = `
  <div class="my-class">
    ${albums[i].title}
  </div>
```

Just more readable

submissions/ses3332002/dom/script.js
```diff
@@ -0,0 +1,257 @@
...
+  for (let i = 0; i < albums.length; i++) {
+    navTabButton = document.createElement("button");
+    navTabButton.classList.add("navigation_button");
+    navTabButton.innerText = albums[i].title;
+    navTab.append(navTabButton);
+  };
+  navTabButtonsArray = Array.from(document.querySelectorAll(".navigation_button"));
+};
+
+function fillInfoTab(e) {
```
zonzujiro, line 240:
```suggestion
function fillInfoTab({ target }) {
```

You сan use ES6 destructuring here. It's 2020!

zonzujiro (approved):
Ok. LGTM!

I believe we can merge it. Nothing critical here as for learning project.

submissions/ses3332002/dom/script.js
ses3332002: I'ii remake it.

submissions/ses3332002/dom/script.js
ses3332002: Well, I can use DocumentFragment, fill it in the loop and make only one DOM insertion in parent block.
Or I can use ES6 strings in loop like someParentBlock.innerHTML+=`<button class="navigation_button">${albums[i].title}</button>`.
But I can't do it together, 'cos DocumentFragment does not have innerHTML.
:) There are many "right" ways, so which one do you want me to go?

MaxKalashnyk: You can use `append` for adding `navTabButton` to the fragment on each loop iteration or use template strings(without fragement):
```JavaScript
let someContent = '';
loop {
   someContent+= '<button class="navigation_button">${albums[i].title}</button>'
}
navTab.insertAdjacentHTML('afterbegin', someContent)
// or
navTab.innerHTML = someContent
```

submissions/ses3332002/dom/script.js
ses3332002: ready

submissions/ses3332002/dom/script.js
ses3332002: Ready. I used DocumentFragment to insert in DOM only once.

MaxKalashnyk (approved):
Approved! Perfect work!

submissions/ses3332002/dom/script.js
ses3332002: Ready. I removed "ul" to HTML and used DocumentFragment to insert in DOM only once.
But as a result my HTML became bigger, my CSS became bigger ('cos I needed to style both "ul" and "li") and my JS became bigger. Honestly: I hardly understand such improvements :)

submissions/ses3332002/dom/script.js
ses3332002: Ready

submissions/ses3332002/dom/script.js
MaxKalashnyk: 1. HTML became bigger per one element? 
2. Why did your CSS become bigger? How did changes in JS affect your styles?
3. Your code became more clear and easy to read.

ses3332002: 1. Yes, HTML became bigger per one element, which is absolutely not needed in "initial" screen (while no button pressed).
2. Because of 1 I needed to remake styles - make this "ul" clear and transparent and give background and paddings to "li" (and reset ugly defaults for both of them).
3. Ok. But I'm afraid that some third reviewer could find another problems, like unnessessery variables, used only for DOM manipulations :)

MaxKalashnyk: @ses3332002 Look, your decision was totally fine. And I appreciate all work you've done! Your code was nice and it worked perfectly.
My idea was to move all elements to HTML (if the situation allows) and to decrease DOM operations in your javascript because as we know, DOM operations cost the most. Maybe, it caused some changes in CSS. But making changes is CSS is no big deal, to my mind. 

BTW, I don't urge you to follow all of my comments and make these changes. There are just recommendations that I learned during my work experience. You can't agree with all of them. It's ok. It's totally up to you. I open to discussions if we need it.

ses3332002: It's Ok, I'm just grumble :)

PR #35, opened 2020-10-28
Dom api task

5Mountains:
### dom-api-task

[Demo](https://5mountains.github.io/dom-api-task/) |
[Code base](https://github.com/5Mountains/dom-api-task)

Please, review.


submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,118 @@
...
+        animation();
+    })
+    .catch(reason => console.log(reason.message));
+
+function createHtml(data) {
+    const menuItems = document.createDocumentFragment();
+    const heroItems = document.createDocumentFragment();
+    data.pancakes.forEach((elem, index) => {
+        //nav
+        //nav item (li)
```
zonzujiro, line 21: What the meaning of comment? :) Looks like redundant for me

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,118 @@
...
+        drawPage(details);
+        toggleTabs();
+        animation();
+    })
+    .catch(reason => console.log(reason.message));
+
+function createHtml(data) {
+    const menuItems = document.createDocumentFragment();
+    const heroItems = document.createDocumentFragment();
+    data.pancakes.forEach((elem, index) => {
```
zonzujiro, line 19: This function looks WAAAAAY to big :D

Pls, split it into smaller parts. For this, you will need to move parts of code into separate functions, like `createImageWrapper`, etc. and call these functions in your callback.

BTW, instead of calling DOM API you can use template strings and write something like:

```javascript
const htmlContent = `
  <figure src={${src}}
    <a href=${href} /> 
  </figure>`;
```

etc. In such a way it will be more readable and understandable what DOM structure you are building.

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,118 @@
...
+            document.querySelectorAll('.hero__item').forEach(item => item.classList.remove('hero__item--active'));
+    
+            elem.classList.add('menu__link--active'); 
+            document.getElementById(index).classList.add('hero__item--active');
+        });
+    });
+    document.querySelector('.menu__link').click();
+}
+
+////animation
```
zonzujiro, line 92: Redundant commit. You have function named `animation`, no need in the comment in such a case.

General rule - your comments should describe *why* something is happening, not *what*. Last one should be clear from code itself.

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,118 @@
...
+function animation(){
+    const heroImgHolders = document.querySelectorAll(".hero__image");
+    heroImgHolders.forEach(item => {
+        item.addEventListener("mousemove", (e) => {
+            let xAxis = (window.innerWidth / 2 - e.pageX) / 25;
+            let yAxis = (window.innerHeight / 2 - e.pageY) / 25;
+            item.style.transform = `rotateY(${xAxis}deg) rotateX(${yAxis}deg)`;
+        });
+    });
+    heroImgHolders.forEach(item => {
```
zonzujiro, line 102: Can you handle this animation with CSS and `:first-child`, `:last-child` selectors?

zonzujiro (changes requested):

submissions/5mountains/dom-api-task/js/main.js
5Mountains: My bad! I forgot to remove all comments and console.log in this folder))) I'll fix it!

submissions/5mountains/dom-api-task/js/main.js
5Mountains: I will! I left them out of this t folder *( but removed them all in the project folder (github pages) (:

submissions/5mountains/dom-api-task/js/main.js
5Mountains: I need some time to do this (:

submissions/5mountains/dom-api-task/js/main.js
5Mountains: I guess yes)  I think I could handle this via :hover. I'll try it

5Mountains:
How about this project version (after some improvements/optimizations) ? :)

submissions/5mountains/dom-api-task/js/main.js
zonzujiro: Still looks messy. Please, separate into functions :)

zonzujiro:
Good work!

One thing from my side :)

submissions/5mountains/dom-api-task/js/main.js
5Mountains: I feel that this time is gonna be OK. I split up maximum (in my point of view) into functions which, in their turn, call to other functions (:

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,104 @@
...
+        animation();
+    })
+    .catch(reason => console.log(reason.message));
+
+function createHtml(data) {
+    const menuItems = document.createDocumentFragment();
+    const heroItems = document.createDocumentFragment();
+    data.pancakes.forEach((elem, index) => {
+        const element = elem;
+        const token = index;
```
zonzujiro, line 32: Why are you need this? O_o

submissions/5mountains/dom-api-task/js/main.js
zonzujiro: Yeah. Looks better now. You can resolve the conversation

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,104 @@
...
+        menuItems.appendChild(navContent);
+        heroItems.appendChild(heroContent);
+    });
+    return {
+        menuItems,
+        heroItems
+    };
+}
+
+function generateNavContent(el, i) {
```
zonzujiro, line 44: Please, give a full name to the variable `i` :)

Such naming, with one letter, is antipattern.

zonzujiro (changes requested):

submissions/5mountains/dom-api-task/js/main.js
5Mountains: While writing the code, the functions did not want to work without additional variables! But, now they do))) mysticism... so, I'll remove it

submissions/5mountains/dom-api-task/js/main.js
5Mountains: okay, good notice!

submissions/5mountains/dom-api-task/js/main.js
zonzujiro: Please resolve this one when you remove them

5Mountains: done (:

submissions/5mountains/dom-api-task/js/main.js
5Mountains: done

submissions/5mountains/dom-api-task/js/main.js
5Mountains: also done

submissions/5mountains/dom-api-task/js/main.js
5Mountains: This is also completed.

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+    heroItem: 'hero__item',
+    heroImg: 'hero__image',
+    itemIcon: 'item__icon',
+    itemTitle: 'item__caption',
+    itemPrice: 'item__price',
+    itemInfo: 'item__description'
+};
+
+async function fetchAsync () {
+    let response = await fetch('./data/data.json');
```
MaxKalashnyk, line 13: Use `const` instead of `let` in these cases. And change it in other places

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+    return item;
+}
+
+function drawPage(details) {
+    document.querySelector('.nav__menu').appendChild(details.menuItems);
+    document.querySelector('.hero').appendChild(details.heroItems);
+}
+
+function toggleTabs(){
+    document.querySelectorAll('.menu__link').forEach(elem => {
```
MaxKalashnyk, line 76: Use one event listener on the parent element, please. More info [here](https://learn.javascript.ru/event-delegation)

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+            document.getElementById(index).classList.add('hero__item--active');
+        });
+    });
+    document.querySelector('.menu__link').click();
+}
+
+function animation(){
+    const heroImgHolders = document.querySelectorAll(".hero__image");
+    heroImgHolders.forEach(item => {
+        item.addEventListener("mousemove", (e) => {
```
MaxKalashnyk, line 92: The same situation with the event listener here. Change it to adding one event listener to the parent element

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+    document.querySelector('.nav__menu').appendChild(details.menuItems);
+    document.querySelector('.hero').appendChild(details.heroItems);
+}
+
+function toggleTabs(){
+    document.querySelectorAll('.menu__link').forEach(elem => {
+        elem.addEventListener('click', (e) => {
+            e.preventDefault();
+            const index = e.target.getAttribute('id').replace('link', 'tab');
+            document.querySelectorAll('.menu__link').forEach(item => item.classList.remove('menu__link--active'));
```
MaxKalashnyk, line 80: As I see, you have only one element with the selector '.menu__link' which has an `active` class. So why do you query multiple elements and iterate them? I think you can find only one element and remove the active class from it

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+            document.querySelectorAll('.hero__item').forEach(item => item.classList.remove('hero__item--active'));
```
MaxKalashnyk, line 81: The same situation as in a previous comment.

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+        menuItems.appendChild(navContent);
+        heroItems.appendChild(heroContent);
+    });
+    return {
+        menuItems,
+        heroItems
+    };
+}
+
+function generateNavContent(elem, index) {
```
MaxKalashnyk, line 42: Why do you create elements with different approaches? I see you generated `li` elements by using `document.createElement` and links you created by using template strings.
I recommend you to use only one approach. For example, template strings.

MaxKalashnyk (changes requested):
Good job! See my comments below

submissions/5mountains/dom-api-task/js/main.js
5Mountains: My idea was to show that I understand and can work with different approaches. But okay, I will bring everything to the one kind.

submissions/5mountains/dom-api-task/js/main.js
5Mountains: done

submissions/5mountains/dom-api-task/js/main.js
5Mountains: ready

submissions/5mountains/dom-api-task/js/main.js
5Mountains: it also has already been done

submissions/5mountains/dom-api-task/js/main.js
5Mountains: I created another path to get this element)

submissions/5mountains/dom-api-task/js/main.js
5Mountains: I optimized these iterations (:

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+        heroContent += heroItems;
+    });
+    return {
+        navContent,
+        heroContent
+    };
+}
+
+function generateNavContent(elem, index) {
+    const navLink = `<a class=${styles.menuLink} href='#' id=link${index+1}>${elem.title}</a>`;
```
MaxKalashnyk, line 49: Let's return the template string without creating variables. It will be more readable, to my mind.
Also, don't forget about checking the input parameters of this function for truthy. In this case, it's acceptable, but in the future keep this in mind.
```suggestion
    return `<li class=${styles.menuItem}><a class=${styles.menuLink} href='#' id=link${index+1}>${elem.title}</a></li>`;
```

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+    const navElem = `<li class=${styles.menuItem}>${navLink}</li>`;
+    return navElem;
+}
+
+function generateHeroContent(elem, index) {
+    const heroPrice = `<div class=${styles.itemPrice}>Price for portion ${elem.price} $</div>`;
```
MaxKalashnyk, line 55: The same as in the previous comment.

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function highlightHeroItem(target) {
+    if (selectHeroItem) selectHeroItem.classList.remove(styles.heroItemActive);
+    selectHeroItem = document.getElementById(target.getAttribute('id').replace('link', 'tab'));
+    selectHeroItem.classList.add(styles.heroItemActive);
+    animation(selectHeroItem);
+}
+
+function animation(elem){
```
MaxKalashnyk, line 91: Each function should start with a verb. You can rename it to `animateSmth`, where `Smth` is something that you want to animate

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+let selectMenuItem;
+let selectHeroItem;
+
+async function fetchAsync () {
+    const response = await fetch(url);
+    const data = await response.json();
+    return data;
+}   
+  
+fetchAsync()
```
MaxKalashnyk, line 25: In order to increase the readability of your code, let's move this function call to the bottom after all your function declarations.
Also, you can create a new function and place there calling `fetchAsync`. Then add some event listener for `DOMContentLoaded`
```suggestion

function initApp() {
fetchAsync().then(data => {
        const details = createHtml(data);
        drawPage(details);
        toggleTabs();
    })
    .catch(reason => console.log(reason.message));
}
document.addEventListener('DOMContentLoaded', init = () => {
    initApp();
    document.removeEventListener('DOMContentLoaded', init);
});
```

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,102 @@
...
+    .then(data => {
+        const details = createHtml(data);
+        drawPage(details);
+        toggleTabs();
+    })
+    .catch(reason => console.log(reason.message));
```
MaxKalashnyk, line 31: It's not a blocker but for displaying errors you can use `console.error`

MaxKalashnyk (changes requested):
Nice work! Look at my comments below, please

submissions/5mountains/dom-api-task/js/main.js
5Mountains: Hmm, okay, in this case, it's a good idea.

submissions/5mountains/dom-api-task/js/main.js
5Mountains: But in this case - it will be a really large line of the template. Or is it okay?)

MaxKalashnyk: Yes, it's ok. Don't forget that you can use the Enter key for moving to another line and just write HTML there

5Mountains: unfortunately, I cannot! Because in my case, when I check the parent's children in my functions, "Enter" shows as an empty string, and break program logic.

MaxKalashnyk: Hm... It's weird. Could you please provide some screenshot or the link to some sandbox with your code?

5Mountains: without Enters - https://user-images.githubusercontent.com/29441499/97614051-22ccf600-1a22-11eb-8f1f-86c363e2e997.jpg

with Enters - https://user-images.githubusercontent.com/29441499/97614046-206a9c00-1a22-11eb-98eb-a049dc76b494.jpg

and space or tab key also have influence...

submissions/5mountains/dom-api-task/js/main.js
5Mountains: It's clear

submissions/5mountains/dom-api-task/js/main.js
zonzujiro: > Because in my case, when I check the parent's children in my functions,

Actually this is a good example, why you should not rely on DOM structure and check parents or children of some element.

In one day DOM structure can be changed - you will have additional button or element will be moved elsewhere in general. Instead of `childNodes` use `querySelector` and instead of parents use [`closest`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

5Mountains: >  use `querySelector` 

I used querySelectorAll, I found solution with this approach, but then I had to found different and I came up with this example of code...

zonzujiro: >  this example of code...

Can you be more specific? :)

submissions/5mountains/dom-api-task/js/main.js
5Mountains: Honestly, I'm a little confused about the second part) I should create one more function, put their async function fetchAsync, and then call it in addEventListener('DOMContentLoaded') ? Could you please explain more in details...

submissions/5mountains/dom-api-task/js/main.js
5Mountains: In this case, I use path > parent.firstChild.nextSibling in order  to dig to the proper element end make animation. You could check it in 92 lines of this JS code. https://user-images.githubusercontent.com/29441499/97616671-8147a380-1a25-11eb-8485-5475ecf791ba.jpg

5Mountains: It is a long road. First, I found the clicked menu item through the parent, then I found the hero's item according to the clicked menu item through comparison/replacement of the ID part, and finally by this hero's item - I go down via children to the item that should be animated ...

MaxKalashnyk: @5Mountains , maybe @zonzujiro means that you can use `parent.querySelector('.hero__image')` insead of `parent.firstChild.nextSibling`. In this case you'll not depend on children and other DOM nodes.

5Mountains: @MaxKalashnyk OMG))) Great Option! I did not know about it... I lack of experience yet...

submissions/5mountains/dom-api-task/js/main.js
5Mountains: Now I got it, there was a little misunderstanding from my side! All is clear!

submissions/5mountains/dom-api-task/js/main.js
5Mountains: Changed it too!

5Mountains:
All requests as well as blockers were done.

submissions/5mountains/dom-api-task/js/main.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function drawPage(details) {
+    sectionMenu.innerHTML = details.navContent;
+    sectionHero.innerHTML = details.heroContent;
+}
+
+function toggleTabs(){
+    sectionMenu.addEventListener('click', e => {
+        e.preventDefault();
```
MaxKalashnyk, line 64: One more thing I noticed. You can handle the `click` event by doing this. And it will look better.
```suggestion
        if (e.target.matches('.menu__link')) {
           highlightMenuItem(e.target);
           highlightHeroItem(e.target);
        }
```
It helps you to remove `selectMenuItem` variable and maybe you'll notice how you can optimize the `highlightHeroItem` function.

MaxKalashnyk (changes requested):
Awesome! You did a great job! See my last! comment and I beg you I'll approve it.

submissions/5mountains/dom-api-task/js/main.js
5Mountains: Hmm... is it really essential?  All my logic that's tied up in a click will break again... but okay, I'll try...

MaxKalashnyk: Yes, it's an essential way to handle click events. Ping me if you'll need any help.

5Mountains: I tried to figure out, but without result) So, I just need to remove e.prevent and put your proposal code (it is too easy) or I need to remove all from toggle func and put there you proposal code and then try to come up with some ideas?

MaxKalashnyk: Yes, you need to remove `e.preventDefault();` and add this check. It will be enough.
In the `highlightMenuItem` you can remove `selectMenuItem` variable and just use `target`. It helps to simplify your code

MaxKalashnyk: And let me know if this approach is clear for you. If you got it now, you can use it with confidence in the future

5Mountains: Oh, and I started thinking in a very different direction, to remove all from func and start to think from what element I should found matched().

MaxKalashnyk: No-no, it's already working fine. Just small optimization :)

MaxKalashnyk: Sorry, you need to leave `e.preventDefault();` but place it inside this `if` statement. It's my fault to recommend you to remove it. I forgot that you use `a` tags in the menu.

5Mountains: No problem! But I need more time, I cannot come up with an idea yet *(

5Mountains: sorry guys, but I need help, I haven't any idea how to tackle this task with target and without extra variable...

MaxKalashnyk: Maybe it'll help you
``` JavaScript
function highlightMenuItem(target) {
    const activeMenuLink = document.querySelector(styles.menuLinkActive);
    if (activeMenuLink) {
        activeMenuLink.classList.remove(styles.menuLinkActive);
    }
    target.classList.add(styles.menuLinkActive);
}
```

5Mountains: Well, I was spinning around this solution, but I still didn't see anything... Thank you :)

MaxKalashnyk (approved):
Approved!
TBH I'm a big fan of pancakes. Maybe it's the best American food! 🥞
Good job! You rock! Continue working with such big enthusiasm!

submissions/5mountains/dom-api-task/js/main.js
5Mountains: Thanks again! Everything is work, but with a slight difference ```const activeMenuLink = document.querySelector('.menu__link--active');``` I don't know why, but it didn't wanna work with variable...

MaxKalashnyk: I think it doesn't work because maybe you missed '.' symbol in the selector. But it's no big deal
``` JavaScript
const activeMenuLink = document.querySelector(`.${menuLinkActive}`)
```

5Mountains: Yep, you're right! Style property was taking from global obj styles, the path was style.menuLinkActive but, in querySelector method i put it without "." I guess it's too late and I'm sleepy)))

MaxKalashnyk: Sure! Take some rest after this work.

zonzujiro (approved):
Lgtm

PR #68, opened 2020-11-04
Document Object Model - done

evgenii-del:

### Document Object Model

[Demo](http://nauseating-disease.surge.sh/) | [Code base](https://github.com/evgenii-del/kottans-menu)

Please, review.


submissions/evgenii-del/dom/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+        icon: 'fas fa-sun',
+        url: 'https://images.pexels.com/photos/979247/pexels-photo-979247.jpeg?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=800&q=60',
+    }
+]
+
+const main = document.querySelector('.js-main');
+const menu = document.querySelector('.js-menu');
+
+const renderCats = (position) => {
+    let cat = cats[position];
```
MaxKalashnyk, line 33: Better to use `const` instead of `let` here

submissions/evgenii-del/dom/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+    main.innerHTML = `
+        <img id="img" src=${cat.url} alt=${cat.name}>`;
+}
+
+const renderItem = (cat, index) => {
+    let block = document.createElement('li');
```
MaxKalashnyk, line 39: the same here

submissions/evgenii-del/dom/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+        li.querySelector('a').className = "";
+    })
+}
+
+window.onload = () => {
+    cats.forEach((item, index) => {
+        renderItem(item, index);
+    })
+
+    document.querySelectorAll('li').forEach((li) => {
```
MaxKalashnyk, line 62: It's better to use event delegation in this case. More info [here](https://learn.javascript.ru/event-delegation)

submissions/evgenii-del/dom/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+        li.addEventListener('click', ({target}) => {
+            removeActive();
+            target.className = 'active';
+            renderCats(target.getAttribute("data-id"));
+        });
+    })
+
+    renderCats(0);
+    document.querySelector('a').className = 'active';
```
MaxKalashnyk, line 71: Let's add some className for the `a` element, because it can broke your app when you add some new `a` tag before this element.

submissions/evgenii-del/dom/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const removeActive = () => {
+    document.querySelectorAll('li').forEach((li) => {
+        li.querySelector('a').className = "";
+    })
+}
+
+window.onload = () => {
+    cats.forEach((item, index) => {
```
MaxKalashnyk, line 58: You decision is totally fine for the case when we have only 5 cat objects for rendering, but when you'll have a lot of data to render it's better to use [fragment](https://developer.mozilla.org/ru/docs/Web/API/DocumentFragment) and append it to some element in your app.

submissions/evgenii-del/dom/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+            <span class="icon">
+                <i class="${cat.icon}"></i>
+            </span>
+            <p class="text">${cat.name}</p>
+        </a>
+    `;
+    menu.appendChild(block);
+}
+
+const removeActive = () => {
```
MaxKalashnyk, line 51: Your decision about removing class is ok. But as we have only one element with the `active` class, let's make it simpler. I just recommend finding this element by 
``` javascript
const activeElement = document.querySelector('a.active')
```
And then you can use `activeElement.classList.remove('active')`
Also, it's better to add some class for these links

submissions/evgenii-del/dom/css/style.css
```diff
@@ -0,0 +1,120 @@
...
+.section__body {
+    width: 100%;
+    padding: 25px;
+}
+
+.main img {
+    width: 100%;
+}
+
+.main .item {
```
MaxKalashnyk, line 86: I couldn't find any element with the `item` class in your app.
Remove this selector when you don't need it

MaxKalashnyk (changes requested):
Nice work! Look at my comments below

submissions/evgenii-del/dom/index.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./css/style.css">
+    <link rel="shortcut icon" type="image/x-icon" href="./img/favicon.ico">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap" rel="stylesheet">
+    <title>Kottans side-menu</title>
+    <script src="https://kit.fontawesome.com/b6d85b8ecc.js" crossorigin="anonymous"></script>
```
MaxKalashnyk, line 10: Using external scripts for icons is ok for this task, but avoid using it in the future if you have only 5 icons. Using SVG icons will be better

MaxKalashnyk (approved):
LGTM! Nice work!

PR #69, opened 2020-11-04
Dom task

AnnaStoyano:
### DOM

[Demo](https://annastoyano.github.io/DOM/index) |
[Code base](https://github.com/AnnaStoyano/DOM)

Please, review.


submissions/AnnaStoyano/js-dom/index.html
```diff
@@ -0,0 +1,37 @@
...
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="style.css">
+    <title>DOM</title>
+</head>
+
+<body>
+    <div class="conteiner">
```
MaxKalashnyk, line 12: It's a typo here
```suggestion
    <div class="container">
```

submissions/AnnaStoyano/js-dom/index.html
```diff
@@ -0,0 +1,37 @@
...
+            </main>
+        </div>
+        <footer class="footer">
+            Created by Anna Stoyanova
+        </footer>
+    </div>
+    <script src="script.js"></script>
+</body>
+
+</html>
```
MaxKalashnyk, line 37: Add an empty line here

submissions/AnnaStoyano/js-dom/style.css
```diff
@@ -0,0 +1,222 @@
...
+
+    .menu-icon.active + .menu-wrapper{
+        transform: translate(0,0px);
+    }
+
+    .menu{
+        margin-top: 55px;
+    }
+    
+}
```
MaxKalashnyk, line 222: Add empty line here as well

submissions/AnnaStoyano/js-dom/script.js
```diff
@@ -0,0 +1,106 @@
...
+            return data;
+        })
+        .then(data=>{
+            menuHandler(data);
+            menuIconHandler();
+        })
+        
+}
+
+download();
```
MaxKalashnyk, line 16: I'd recommend you using [DOMContentLoaded](https://developer.mozilla.org/uk/docs/Web/Events/DOMContentLoaded) event listener. You can place there your function `download`. 
Also, I'd recommend renaming this function, because to my mind, this name is too generic.

submissions/AnnaStoyano/js-dom/script.js
```diff
@@ -0,0 +1,106 @@
...
+
+let addMenuItems = function (arr) {
```
MaxKalashnyk, line 18: Let's use `let` and `const` in appropriate cases. More info about it [here](https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/)

MaxKalashnyk (changes requested):
Awesome work! I really appreciate it!
Look at my comments below, please

AnnaStoyano:
Thank you for your advices)

submissions/AnnaStoyano/js-dom/script.js
```diff
@@ -0,0 +1,105 @@
...
+            menuHandler(data);
+            menuIconHandler();
+        })      
+}
+
+document.addEventListener('DOMContentLoaded',downloadInfoFromJson);
+
+const addMenuItems = function (arr) {
+    let menu = document.querySelector('.menu');
+    let menuFr = document.createDocumentFragment();
```
madmaxWMFU, line 19: Use a more informative value names

submissions/AnnaStoyano/js-dom/script.js
```diff
@@ -0,0 +1,105 @@
+const downloadInfoFromJson = function () {
+    let promise = fetch('./information.json');
+    promise.then(res => res.json())
+        .then(data => {
+            addMenuItems(data.faculties);
+            addContentItems(data.faculties);
+            return data;
+        })
+        .then(data=>{
+            menuHandler(data);
```
madmaxWMFU, line 10: Why you used twice the method `then`?
Think about constructive code
This function only returns data. You can write data to a variable and call a function. Calling all four functions outside of this function
Also, think about network conflicts and use the catch method to inform users of the problem.

madmaxWMFU (changes requested):
Nice work!
See some comments below

submissions/AnnaStoyano/js-dom/script.js
AnnaStoyano: Do you mean move this functions to one .then()? I thought that fetch is async, so before we call event handler we should wait until all data have downloaded, that's why I decided to use 2 different .then().

madmaxWMFU: The method `then` is already working with the loaded data. You can read about [async/wait](https://developer.mozilla.org/ru/docs/Learn/JavaScript/Asynchronous/Async_await) and use this in final projects.

submissions/AnnaStoyano/js-dom/script.js
```diff
@@ -0,0 +1,105 @@
...
+        if (index === 0) {
+            contentItem.classList.add('active');
+        }
+        contentFr.appendChild(contentItem);
+    })
+
+    content.appendChild(contentFr);
+}
+
+const menuIconHandler = function () {
```
MaxKalashnyk, line 62: It's better to start function name from a verb. More info you can find in this [article](https://medium.com/javascript-in-plain-english/javascript-naming-convention-best-practices-b2065694b7d) in a section about the functions

MaxKalashnyk (changes requested):
Good job!
Look at my comments below

AnnaStoyano:
Thank you for your advices) Hope, that now everything is ok =)

MaxKalashnyk (commented):
Good job! 
I think we can approve your code. I noticed the empty space in the bottom on the mobile devices, but it's up to you to fix that.
I see that you can work confidently with the DOM API. But let's imagine that we'll have 100500 menu items. In this case, your suggested approach won't work properly because of the performance issues. 

What can cause potential problems:
1. Rendering all these elements and adding unique class names to them. 
2. Your `checkFaculty` function would be quite enormous.
3. Hiding all these elements by using `display: none;` inside the `main` tag.
4. Using class names for handling logic stuff. It's better to use `data-attributes`.

What can we do to solve these problems:
1. Try to think about the incoming data at first: Will they scale in the future? If yes, how does our app would react to these changes? Do we need to change our code when the data is changed?
2. Try to render only the necessary content at this moment (it means that you can render your block in the HTML and just toggle some attributes inside them according to the data changes).
3. Consider adding some `id` attribute to the each object of data and try to operate by this id (you can use the `data-attributes`).

madmaxWMFU (approved):

mkalashnyk-asp (approved):

MaxKalashnyk (approved):

PR #71, opened 2020-11-04
Oop(Tiny js world advanced)

mustbefail:
### OOP Task

[Demo](https://mustbefail.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/mustbefail/a-tiny-JS-world)

Please, review.


OleksiyRudenko (changes requested):
Pretty nice job!
Please:
- use `super` to initialize parent object properties.
- do not use `Object.assign` to modify object properties.
- pass named parameters to constructors rather than an object comprising all object properties.
- avoid assigning properties that are defined in a parent class from a child class (that's what `Object.assign` does in fact) as child class shouldn't know how parent class internals are constructed (it is often not just plain assignment)
- consider having a superclass that holds properties that are common for all classes (yet you may keep `Animal` and `Human` as interim superclasses)

mustbefail:
> Pretty nice job!
> Please:
> 
> * use `super` to initialize parent object properties.
> * do not use `Object.assign` to modify object properties.
> * pass named parameters to constructors rather than an object comprising all object properties.
> * avoid assigning properties that are defined in a parent class from a child class (that's what `Object.assign` does in fact) as child class shouldn't know how parent class internals are constructed (it is often not just plain assignment)
> * consider having a superclass that holds properties that are common for all classes (yet you may keep `Animal` and `Human` as interim superclasses)

So I could not make a catWomen with class syntax. I tried to do it through composition, could you please check if I understood this technique correctly. [composition](https://github.com/mustbefail/a-tiny-JS-world/blob/composition-over-inheritance/index.js)

OleksiyRudenko:
> So I could not make a catWomen with class syntax. I tried to do it through composition, could you please check if I understood this technique correctly. [composition](https://github.com/mustbefail/a-tiny-JS-world/blob/composition-over-inheritance/index.js)

There is still room for minor improvements, but yes, you got the essence of composition pattern right.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(species, name, gender, hands, legs, friends, phrase) {
+    this.friends = friends;
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.hands = hands;
```
OleksiyRudenko, line 16: The point of OOP is that:
- a super class has only shared properties
- a child class introduces and handles properties unqiue to that child class

I.e. `hands` is smth unique to humans. Cats or dogs shouldn't have any hands among their props while they inherit them as a property from class `Inhabitant`.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,88 @@
...
+    this.phrase = phrase;
+  }
+  saying() {
+    return this.phrase;
+  }
+}
+
+class Cat extends Inhabitant {
+  constructor(name, gender, friends, phrase) {
+    super('cat', name, gender, 0, 4, friends, phrase);
```
OleksiyRudenko, line 27: This way we do not ask a developer to specify species when constructing a cat and the species is hardcoded.
It's a good approach.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+class Human extends Inhabitant {
+  constructor(name, gender, friends, phrase) {
+    super('human', name, gender, 2, 2, friends, phrase);
+  }
+}
+
+const dog = new Dog('Maki', 'male', ['Alfred', 'Prokhor'], 'Bark-bark!');
```
OleksiyRudenko, line 43: Can you make friends real inhabitants rather than arbitrary names of virtual friends?

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,88 @@
...
+    this.legs = legs;
+    this.phrase = phrase;
+  }
+  saying() {
+    return this.phrase;
+  }
+}
+
+class Cat extends Inhabitant {
+  constructor(name, gender, friends, phrase) {
```
OleksiyRudenko, line 26: `phrase` is something that every inhabitant has. They may not have friends.
This how the task description puts it.
Please make friends an optional property.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,88 @@
...
+    super('cat', name, gender, 0, 4, friends, phrase);
+  }
+}
+
+class Dog extends Inhabitant {
+  constructor(name, gender, friends, phrase) {
+    super('dog', name, gender, 0, 4, friends, phrase);
```
OleksiyRudenko, line 33: This says 'create a dog with 0 hands and etc' but dogs do not have any hands by design.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,88 @@
...
+  Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+  Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+  However, please, REFRAIN from improving visuals at least until your code is reviewed
+  so code reviewers might focus on a single file that is index.js.
+*/
+
+const inhabitants = [cat, man, woman, dog];
+
+const formatOutput = (habitant) => {
```
OleksiyRudenko, line 64: Why not make this a method of a super class?
Consider naming the method `toString`. It's a magic method that is called by default in string context.
For example `[Object object]` is the output from implicit `toString` method that every object has. You may override this method.

OleksiyRudenko (changes requested):
@mustbefail Nice job!
Let's improve the implementation in a couple of iterations.
You will find comments re iteration 1 below.

BTW, did you have a chance to study [extra materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) from the original repo? Those are helpful in better understanding of OOP as a set of principles, not only syntax.

mustbefail:
> BTW, did you have a chance to study [extra materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) from the original repo? Those are helpful in better understanding of OOP as a set of principles, not only syntax.

I am trying to understand OOP. I read the materials, books and watch videos. When you see and read, it seems everything is clear. When you start doing, you understand that it is more difficult than it seems.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,86 @@
...
+// Define your objects here
+class Inhabitant {
+  constructor(name, gender, phrase, friends, legs) {
+    this.friends = friends ? friends : ['no friends'];
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.phrase = phrase;
+  }
+  saying() {
```
OleksiyRudenko, line 18: Why do you need this method?

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,86 @@
...
+    return this.phrase;
+  }
+
+  toString() {
+    return `Hello! I am <strong>${this.species}</strong>, my name is <strong>${
+      this.name
+    }</strong>, my gender is <strong>${this.gender}</strong>, I have <strong>${
+      this.hands
```
OleksiyRudenko, line 26: This class knows nothing about child classes properties (and naturally shouldn't know). In theory you may have armless inhabitants and output will break.
Consider method overloading in child classes calling `super.parentClassMethod()` to method outcome.

submissions/mustbefail/oop/index.js
```diff
@@ -0,0 +1,86 @@
...
+  constructor(name, gender, phrase, friends, legs = 4) {
+    super(name, gender, phrase, friends, legs);
+    this.species = 'cat';
+  }
+}
+
+class Dog extends Inhabitant {
+  constructor(name, gender, phrase, friends, legs = 4) {
+    super(name, gender, phrase, friends, legs);
+    this.species = 'dog';
```
OleksiyRudenko, line 45: Isn't species a shared property worth defining in super class?
You had a nice solution for this before. Not every comment is about fixing something :)

OleksiyRudenko (changes requested):
@mustbefail pretty well done!
Let's take iteration 2. If you implement things right we're gonna be done.
Please update your demo so it reflects all the changes in your code.
This helps to see if the output is as expected. I suppose it is not yet.

submissions/mustbefail/oop/index.js
mustbefail: > This way we do not ask a developer to specify species when constructing a cat and the species is hardcoded.
> It's a good approach.

Misunderstood

submissions/mustbefail/oop/index.js
mustbefail: I use logic: an action is a function

submissions/mustbefail/oop/index.js
mustbefail: I hope I understood correctly what needs to be done

submissions/mustbefail/oop/index.js
OleksiyRudenko: Logic is good. Then function name should be a verb or start with a verb in imperative mood. E.g. `getProperties`, `sendMessage`, `jump`, `say`.
You are not required to fix this. Fix if you want use this project as a part of your portfolio to show to potential employers.

submissions/mustbefail/oop/index.js
OleksiyRudenko: Upd: Consider method overloading in child classes calling `super.parentClassMethod()` **to supplement child class method outcome**.

mustbefail: done)

submissions/mustbefail/oop/index.js
mustbefail: Fixed

OleksiyRudenko (approved):
@mustbefail Great job! Approved.

A couple of topicc for you to think about:
`friends` in your implementation is just a list of names. 

Besides it is `Array.toString` that nicely produces a comma-separated list of an array elements.

What if we wanted a list of friends to be effectively an array of objects representing inhabitants? How would you change your code and how would you approach friends list output (presuming we still want to see just their names)?

mustbefail:
> @mustbefail Great job! Approved.
> 
> A couple of topicc for you to think about:
> `friends` in your implementation is just a list of names.
> 
> Besides it is `Array.toString` that nicely produces a comma-separated list of an array elements.
> 
> What if we wanted a list of friends to be effectively an array of objects representing inhabitants? How would you change your code and how would you approach friends list output (presuming we still want to see just their names)?

The obvious answer is to override method` toString` of the `Array` class, but this is not very good as far as I read.

OleksiyRudenko:
> The obvious answer is to override method toString of the Array class, but this is not very good as far as I read.
No it is not, indeed. It was just for your awareness that `toString` is an implicit method that every object in JS has.

With regards to managing friends you'd rather need something like
```
class A {
  constructor(friends = []) {
    this.friends = [];
    this.addFriends(friends);
  }
  addFriends() {
    this.friends = [...this.friends, ...friends];
 }
  listFriendNames() {
    return this.friends.map(friend => friend.name).join(", ");
  }
  toString() {
    return 'whatever' + this.listFriendNames();
  }
}

const human1 = new A();
const human2 = new A([human1]);
human1.addFriends([human2]);
```
The above is oversimplified. Just try to gain major ideas and concepts from it.

PR #78, opened 2020-11-06
Task js DOM added

BlueLamantine:
### JS DOM

[Demo](https://bluelamantine.github.io/kottans-frontend/task_js_dom/) |
[Code base](https://github.com/BlueLamantine/kottans-frontend/tree/main/task_js_dom) |

Please, review.


submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+	},
+	paragraph: 'A shot of espresso in steamed milk lightly topped with foam.',
+}];
+const MENU = document.getElementById('mainMenu');
+const CUP = document.getElementById('coffeeCups');
+const ART = document.getElementById('articles');
+
+function render() {
+	document.getElementById('topHeader').innerHTML = `
+    <h1 class="mainHeader">Cat & C<span class="o"></span>ffee</h1>
```
MaxKalashnyk, line 57: You can move this title to the html page as it doesn't continue any dynamic content

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+    `;
+	COFFEE.forEach(element => {
+		MENU.appendChild(listNode(element));
+		CUP.appendChild(cupNode(element));
+		ART.appendChild(contentNode(element));
+	});
+	document.getElementById('mainFooter').innerHTML = `
```
MaxKalashnyk, line 64: The same with the footer - move it to the html

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+		}
+	});
+}
+
+document.addEventListener('DOMContentLoaded', (load = () => {
+	render();
+	toggleMenu();
+	responsiveMenu();
+	document.removeEventListener('DOMContentLoaded', load);
+}));
```
MaxKalashnyk, line 154: Add an empty line, please

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+
+function toggleContent(idEL, arr) {
+	const setHidden = arr.find(el => !el.hasAttribute('hidden'));
+	setHidden.hidden = true;
+	const removeHidden = arr.find(el => el.id.indexOf(idEL) == 0);
+	removeHidden.hidden = false;
+}
+
+function responsiveMenu() {
+	document.getElementById('menu-btn').addEventListener('click', e => {
```
MaxKalashnyk, line 137: Let's use tag `button` with the type `button` for this element `document.getElementById('menu-btn')`. It will be more semantically.

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+};
+
+function toggleContent(idEL, arr) {
+	const setHidden = arr.find(el => !el.hasAttribute('hidden'));
+	setHidden.hidden = true;
+	const removeHidden = arr.find(el => el.id.indexOf(idEL) == 0);
+	removeHidden.hidden = false;
+}
+
+function responsiveMenu() {
```
MaxKalashnyk, line 136: I'd recommend to rename this function because now it's not clear. Try to start naming it from a verb and try to describe what it will do. In this case this function adds the `click` handler. So the appropriate name for it would be better to connect to the click adding

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+	document.getElementById('menu-btn').addEventListener('click', e => {
+		e.preventDefault();
+		e.target.classList.toggle("fa-bars");
+		e.target.classList.toggle("fa-times");
+		if (MENU.className === "menu") {
```
MaxKalashnyk, line 141: It's better to use `classList.add` or `classList.remove`

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
+const COFFEE = [{
```
MaxKalashnyk, line 1: I appreciate you approach with creating this `COFFEE` array, but I'd recommend splitting your data from the user actions. It won't be connected. I mean the `state` attribute. It's better to move this state outside from your data, because it isn't consistent enough

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,154 @@
...
+	return cup;
+}
+
+function contentNode(el) {
+	const articleContent = document.createElement('article');
+	const headerContent = document.createElement('h2');
+	const textContent = document.createElement('p');
+	articleContent.id = `${el.id}-article`;
+	articleContent.className = 'description';
+	if (!el.state) {
```
MaxKalashnyk, line 101: Maybe, it's better to add some `hidden` class here and add some styles in your css for it

MaxKalashnyk (changes requested):
Good job! I'm a big fan of coffee, so special respect for choosing coffee topic ;)
Look at my comments below, please

BlueLamantine:
@MaxKalashnyk thanks a lot for attention to my work! I followed your recommendations.
P.S. Coffee is my "perpetual motion machine".

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: :ok_hand: Done!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: :ok_hand: Done!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: :ok_hand: Done!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: Thanks for the comment !
With ```<button>``` I don't need anymore extra ```<div>``` wrapper in that chunk of html!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: :thinking: I was thinking of this, so this function supposed to be named toggleMenu.
This function does visible/hidden menu by click.

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: :ok_hand: Done!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: Let me explain. This attribute is a marker for main elements visibility, when DOM generated at first time. As you recommended I removed it from data and instead of ```state``` I use attribute ```id``` for this role (main element visible after reload page).

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: There are ways to hide all unnecessary content with css, but I prefer to use HTML boolean attribute ```hidden``` in this task, with this method I hide other elements.

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,144 @@
...
+		articles: [...document.getElementsByTagName('article')],
+		cups: [...document.getElementsByClassName('cup')]
+	}
+	MENU.addEventListener('click', e => {
+		e.preventDefault();
+		if (e.target.matches('.item__link')) {
+			const arrLi = [...MENU.children];
+			const isActive = arrLi.find(li => li.classList.contains("active"));
+			isActive.classList.remove('active');
+			e.target.parentNode.classList.add('active');
```
MaxKalashnyk, line 115: Be attentive with using `e.target.parentNode` because it strongly depends on your layout. It's better to use `closest`. More info [here](https://developer.mozilla.org/ru/docs/Web/API/Element/closest)

submissions/BlueLamantine/js-dom/index.html
```diff
@@ -0,0 +1,35 @@
...
+    </header>
+    <main class="content">
+      <nav>
+        <button id="menu-btn" type="button" class="icon">
+          <i class="fa fa-bars fa-2x"></i>
+        </button>
+        <ul id="mainMenu" class="menu"></ul>
+      </nav>
+      <div class="coffee">
+        <div id="coffeeCups" class="coffe-content"></div>
```
MaxKalashnyk, line 25: I'd recommend combining these two blocks into one. It'll simplify your rendering logic and reduce the number of layout blocks.

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,144 @@
...
+};
+
+function toggleMenuItem() {
+	const content = {
+		articles: [...document.getElementsByTagName('article')],
+		cups: [...document.getElementsByClassName('cup')]
+	}
+	MENU.addEventListener('click', e => {
+		e.preventDefault();
+		if (e.target.matches('.item__link')) {
```
MaxKalashnyk, line 111: you can create some variable for storing `e.target` before this condition.

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,144 @@
...
+			const arrLi = [...MENU.children];
```
MaxKalashnyk, line 112: It's better to use `document.querySelector('.someElement.active')` for finding active element

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,144 @@
...
+	const removeHidden = arr.find(el => el.id.indexOf(idEL) == 0);
+	removeHidden.hidden = false;
+};
+
+function toggleMenu() {
+	document.getElementById('menu-btn').addEventListener('click', e => {
+		e.preventDefault();
+		e.target.classList.toggle("fa-bars");
+		e.target.classList.toggle("fa-times");
+		MENU.className === "menu" ?
```
MaxKalashnyk, line 134: It's better to use `classList.contains` here

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,144 @@
...
+	}
+	headerContent.innerHTML = el.title;
+	textContent.innerHTML = el.paragraph;
+	articleContent.appendChild(headerContent);
+	articleContent.appendChild(textContent);
+	return articleContent;
+};
+
+function toggleMenuItem() {
+	const content = {
```
MaxKalashnyk, line 105: You can combine the rendering of this content into one function. No need to separate them, as cups are connected to the texts in articles. It allows you to get rid of calling one of the `toggleContent` functions.

MaxKalashnyk (changes requested):
Good job! Look at my comments below

submissions/BlueLamantine/js-dom/index.html
BlueLamantine: @MaxKalashnyk Please, can you specify which exactly blocks need to be combined ? All this blocks ```<div>``` for providing a more efficient way to lay out.  Responsive lay out is implement with flex-boxes.

MaxKalashnyk: It's not so critical in this task. But I mean why did you separate rendering logic of cups and text information in different sections. Maybe I was wrong that I left this comment in the layout section. I left a comment about it [here](https://github.com/kottans/frontend-2021-homeworks/pull/78/files#r519664676)

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: Ok, thanks for a good tip!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: My agreements, I use ```e.target``` many times below, so now it is const ```selectedMenuItem```

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: OK, done!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: Ok, done!

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: In my plan was to do function ```toggleContent``` more multi purpose, because render of content cups & articles uses the same code, but ok, let it be as you want.

submissions/BlueLamantine/js-dom/index.html
BlueLamantine: Now I got it. In my plan was to do function ```toggleContent``` more multi purpose, because render of content cups & articles uses the same code, but ok, let it be as you want.

submissions/BlueLamantine/js-dom/dom.js
BlueLamantine: I've combined.

submissions/BlueLamantine/js-dom/index.html
BlueLamantine: Excuse for this add : I thought that I've written the same part of code for two actions (renders) and I had to provide a function to avoid a duplicate of code.

submissions/BlueLamantine/js-dom/dom.js
```diff
@@ -0,0 +1,148 @@
...
+			const currentMenuItem = document.querySelector('.menu-item.active');
+			currentMenuItem.classList.remove('active');
+			selectedMenuItem.closest('li').classList.add('active');
+			toggleContent(selectedMenuItem.id);
+		}
+	});
+};
+
+function toggleContent(idEL, arr) {
+	const content = {
```
MaxKalashnyk, line 118: It seems we didn't get ourselves in the previous comments
Look, now you're trying to hide two sections which depend on one cup object.
Why do we separate information about ONE type of coffee into TWO separate blocks? Let's make it simpler. You can query only the wrap of article and cup and make it visible and hidden.
It helps you to reduce the amount of similar code in this function.
Let me know if you got me.

BlueLamantine: OK. Now it is clear! :grin:
Now I have wrapper of cups and articles and hide it with class ```.hidden{dislpay:none}```, not separately like it was earlier. 
And DOM creates this wraps-blocks (blocks are does not exist in initial html file).

MaxKalashnyk (approved):
Awesome job! Approved!

MaxKalashnyk (commented):
It looks much better now!
Also, I'd like to add that you can have a lot of types of coffee (f.e. 1000) and your approach will not work properly enough because you'll need to render and hide a lot of elements. It can drastically reduce the performance of your app. In this case, you should render only the visible part of coffee information and just toggle image, and text on each click to the menu item (no need to render all of the elements, only one section and change content there). Keep this thing in your mind for future tasks and challenges. 
For the current task requirements, your approach is totally ok.

PR #81, opened 2020-11-06
added initial code

Iakow:
## DOM

[Demo](https://iakow.github.io/kottans-js-dom/) |
[Code base](https://github.com/Iakow/frontend-2021-homeworks/tree/js-dom/submissions/iakow/js-dom)

Please, review.


submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,169 @@
...
+  CARS[navLink.id].descr.forEach(item => {
+    const p = document.createElement('p');
+    p.innerHTML = item;
+    carDescription.append(p);
+  });
+
+  document.querySelector('#burger-ctrl').checked = false;
+}
+
+document.querySelectorAll('.nav-link').forEach(item => item.addEventListener("click", handler));
```
MaxKalashnyk, line 169: Don't forget about using events [delegation](https://learn.javascript.ru/event-delegation) here

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,169 @@
...
+const carPhoto = document.querySelector('.photo');
+const carName = document.querySelector('h2');
+const carYears = document.querySelector('.years');
+const carDesign = document.querySelector('.design');
+const carWins = document.querySelector('.wins');
+const carDescription = document.querySelector('.descr');
+
+let currentCarIndex = "0";
+
+const handler = (e) => {
```
MaxKalashnyk, line 143: You can destrucure the `event` here and just pass `{target}`.
It helps you to get rid of creating a new variable `navLink`, you can just use `target` for it

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,169 @@
...
+      "Ground-breaking design? Check. Iconic livery? Check. Uncluttered lines? Check. Beautiful engineering? Check.",
+      "The Lotus 79, which refined the ground-effect concept and dominated the 1978 F1 season, ticks all the boxes for this list. It's also even more impressive close up in the flesh than it is in the famous period images of Mario Andretti and Ronnie Peterson.",
+      "Lower, neater and sleeker than the Lotus 78 that preceded it, the 79 won on its debut in the Belgian GP in the hands of Andretti and finished 1-2 next time out in Spain. It was only beaten to pole once during the campaign, though the odd reliability issue and jump-start penalty for Andretti at Monza meant it only won six of the 11 championship rounds it started in 1978.",
+      "By 1979 the opposition had caught up in the ground-effect stakes and the Lotus 79 was limited to four podiums. Even then it looked impressive in its green Martini livery, but it's the JPS-liveried original that deservedly earns our top spot."
+    ]
+  },
+]
+
+const carPhoto = document.querySelector('.photo');
+const carName = document.querySelector('h2');
```
MaxKalashnyk, line 135: You can add some class for this element and make querying by this class because adding a new `h2`element before current element can break your app.

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,169 @@
...
+  const navLink = e.target;
+
+  if (navLink.id === currentCarIndex) return;
+  currentCarIndex = navLink.id;
+
+  document.querySelector('.active').classList.toggle('active');
+  navLink.classList.toggle('active');
+
+  carPhoto.setAttribute('src', CARS[navLink.id].pic);
+  carName.innerText = CARS[navLink.id].name;
```
MaxKalashnyk, line 153: Maybe it's bette to use `textContent` here. See the difference [here](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext)

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,169 @@
...
+  carYears.innerText = CARS[navLink.id].years;
+  carDesign.innerText = CARS[navLink.id].design;
+  carWins.innerText = CARS[navLink.id].wins;
+
+  document.querySelectorAll('.descr > p').forEach(item => item.remove());
```
MaxKalashnyk, line 158: You don't need to remove these nodes from your DOM. Just use `innerHTML` for inserting a new content there. It'll overwrite the previous content inside this element.

MaxKalashnyk (changes requested):
Good job! Maybe it's better to make the whole your app data-driven. I mean you can generate all the content in your js code. I see you do it partially now. But i's up to you.
Look at my comments below as well.

madmaxWMFU (changes requested):
Good job!
Think about how to do the initial loading of information on the page.

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,169 @@
...
+const carName = document.querySelector('.car-name');
+const carYears = document.querySelector('.years');
+const carDesign = document.querySelector('.design');
+const carWins = document.querySelector('.wins');
+const carDescription = document.querySelector('.descr');
+
+let currentCarIndex = "0";
+
+const handler = (e) => {
+  const { target } = e;
```
madmaxWMFU, line 144: Another type of adding is `const handler = ( {target} ) => {`

submissions/iakow/js-dom/main.js
Iakow: There are different numbers of paragraphs in different cases. So I can't just inject new content into paragraphs.

Or did you mean I should use innerHTML for the parent? But this way I still delete these nodes. (Or not?). Аnd this option will complicate the logic.

I'm not sure if I understand you correctly. Please clarify.

submissions/iakow/js-dom/main.js
Iakow: Wow! I didn't know that. Very cool! Thanks)

Iakow:
> Good job! Maybe it's better to make the whole your app data-driven. I mean you can generate all the content in your js code. I see you do it partially now. But i's up to you.

I started the task with this idea. But after a chat about "ability to load some predefined html content" I settled on this option. Since this is up to me, I will leave it that way, but only because I panic because of I won't have enough time for all tasks ))

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,167 @@
...
+  carName.textContent = CARS[target.id].name;
+  carYears.textContent = CARS[target.id].years;
+  carDesign.textContent = CARS[target.id].design;
+  carWins.textContent = CARS[target.id].wins;
+
+  document.querySelectorAll('.descr > p').forEach(item => item.remove());
+
+  CARS[target.id].descr.forEach(item => {
+    const p = document.createElement('p');
+    p.innerHTML = item;
```
MaxKalashnyk, line 160: It's better to use `textContent` here because you insert only text to the DOM node.

submissions/iakow/js-dom/main.js
MaxKalashnyk: I mean you can use the `innerHTML` when you insert content for description into the `carDescription` node.
You can remove `carDescription.append(p);` (line 161) and rebuilt your logic with the descriptions using `Document.createDocumentFragment()`. More info [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment).
It will help you to get rid of the line 156 and improve performance a little bit with adding description nodes.
Let me know if it's clear for you

submissions/iakow/js-dom/main.js
```diff
@@ -0,0 +1,167 @@
...
+const carWins = document.querySelector('.wins');
+const carDescription = document.querySelector('.descr');
+
+let currentCarIndex = "0";
+
+const handler = ({ target }) => {
+  if (target.id === currentCarIndex) return;
+  currentCarIndex = target.id;
+
+  document.querySelector('.active').classList.toggle('active');
```
MaxKalashnyk, line 147: I'd recommend to move all these DOM manipulations to the separate function. It will be more readable.

MaxKalashnyk (changes requested):
Good job! Look at my comments below.

submissions/iakow/js-dom/main.js
Iakow: Got it, thanks!

submissions/iakow/js-dom/main.js
Iakow: I am often worried about the question in such cases: should I declare it inside the handler or outside?

MaxKalashnyk: It' better to place it outside. It'll help you to avoid creating this new function on each click.

Iakow: Oh, of course)) Thank you!

MaxKalashnyk (approved):
Nice work! LGTM!

Iakow:
Thank you very much! Brought out some cool new things.

PR #88, opened 2020-11-09
Pre-oop task

HannaSyn:
### Tiny JS world

[Demo](https://hannasyn.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/HannaSyn/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/HannaSyn/preOop/index.js
+++ b/submissions/HannaSyn/preOop/index.js
@@ -1,0 +1,77 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitans {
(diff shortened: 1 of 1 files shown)


OleksiyRudenko (commented):
@HannaSyn good job!
This qualifies as pre-OOP.
You will revert to this task after a while, where current implementation won't be sufficiently qualifying.
You will need to learn more about OOP (OOP is more about structure, hierarchy rather than just about syntax in JS). As an example: the point of having a super class and child classes is that the superclass holds only properties that are common for all child classes, and child classes implement what is unique in them.

My proposal is that you improve your code based on learnings from https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up when you reach [OOP Exercise](https://github.com/kottans/frontend/blob/master/tasks/js-post-oop.md)

Please let me know if you have questions at this stage or will revert to this topic later?

HannaSyn:
Thank you for review! I think, I understand, what you mean. I will improve my code on OOP task.

OleksiyRudenko:
Thank you for acknowledgement.
We're done here.
Please improve your code when you're back to your tiny js world later.

PR #89, opened 2020-11-09
pre-opp

AnnaStoyano:
### Pre-oop

[Demo](https://annastoyano.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/AnnaStoyano/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/AnnaStoyano/pre-oop/index.js
+++ b/submissions/AnnaStoyano/pre-oop/index.js
@@ -59,14 +59,15 @@ const catWoman = {
 
 
 const inhabitants = [man,woman,cat,dog,catWoman];
-let output = '';
-inhabitants.forEach(item=>{
-   for(let key in item){
-      output+=`<b>${key}</b>: ${item[key]}\t`
-   }
-   output+='\n'
+
+const properties = ['name','species','gender','legs','hands','saying','friends'];
+let output = inhabitants.map(inhabitant=>{
+   let match = properties.map(property=>{
+      return `<b>${property}:</b> ${inhabitant[property]}`
+   });
+   return match.join('\t')
 })
-print(output);
+print(output.join('\n'));
 // ======== OUTPUT ========
 /* Use print(message) for output.
    Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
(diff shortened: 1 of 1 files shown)


submissions/AnnaStoyano/pre-oop/index.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+
+const inhabitants = [man,woman,cat,dog,catWoman];
+let output = '';
+inhabitants.forEach(item=>{
+   for(let key in item){
+      output+=`<b>${key}</b>: ${item[key]}\t`
+   }
+   output+='\n'
```
OleksiyRudenko, line 67: This fragment relies on:
1. order of object properties (which is not guaranteed)
2. there are no extra properties we do not want in output

Both issues can be resolved by specifying the list of keys and mapping it into `<b>${key}</b>: ${item[key]}\t`.

Besides, after having keys mapped into strings they can be joined with `\t`, and strings per object then subsequently can be joined with `\n`.

In other words you might use `.map`, `.join` also nested (map+join per each object, map+join per key list for each object).

OleksiyRudenko (changes requested):
@AnnaStoyano Well done!
Please let me know if you want to take a challenge as per comment below.
The purpose is to improve code scalability and maintainability by using `Array` methods (in this particular case).

AnnaStoyano:
Thank you for your advice)

submissions/AnnaStoyano/pre-oop/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+const inhabitants = [man,woman,cat,dog,catWoman];
+
+const properties = ['name','species','gender','legs','hands','saying','friends'];
+let output = inhabitants.map(inhabitant=>{
+   let match = properties.map(property=>{
+      return `<b>${property}:</b> ${inhabitant[property]}`
+   });
+   return match.join('\t')
+})
```
OleksiyRudenko, line 69: Chaining:
```
let a = arr.map(...);
return a.join(...);

// versus
return arr.map(...).join(...);
```

Arrow functions:
```
arr.map(e => {
  let a = e*2;
  return a;
})

// versus
arr.map(e => e*2);
```
Please let me know if the above gives you hints on how to make the code even more concise.

OleksiyRudenko (changes requested):
Good job!
One more iteration to make it great

OleksiyRudenko (approved):
@AnnaStoyano Great job!

PR #94, opened 2020-11-09
Oop

usides:
### OOP

[Demo](link-to-webpage-on-github-pages) |
[Code base](https://usides.github.io/a-tiny-JS-world/)

Please, review.


OleksiyRudenko (approved):
@usides Brilliant work!
You got the idea of composition pattern perfectly well.

You do not need to fix anything. The following is just FYI.

Passing a partial object to constructor is a solution that offers flexibility (e.i. one may pass a value for species if willing so), but quite often we do not want to grant that much of flexibility and want to keep some properties 'hard-coded' (e.g. pass a scalar to `createHuman` rather then default value from arguments list). Keep this in mind when you employ composition in production.

usides:
> @usides Brilliant work!
> You got the idea of composition pattern perfectly well.
> 
> You do not need to fix anything. The following is just FYI.
> 
> Passing a partial object to constructor is a solution that offers flexibility (e.i. one may pass a value for species if willing so), but quite often we do not want to grant that much of flexibility and want to keep some properties 'hard-coded' (e.g. pass a scalar to `createHuman` rather then default value from arguments list). Keep this in mind when you employ composition in production.

That's a good idea! Thanks!

PR #98, opened 2020-11-09
added a tiny-js-world-opp my versions of the task

5Mountains:
### OOP exercise (tiny-js-world-oop)

[Demo](https://5mountains.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/5Mountains/a-tiny-JS-world/blob/master/index.js)

Please, review.


submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,105 @@
...
+  introducing() {
+    return `Wonderful creature - <b>${
+      this.species
+    }</b>, whose name is <b>${
+      this.name
+    }</b>! Usually the greeting is <b>${
+      this.saying
+    }</b></i> and friends of this inhabitant are <b>${
+      this.friends.join(', ')
+    }</b>. By the wat, this member is owner of`;
```
OleksiyRudenko, line 27: It is not a really good practice for a superclass to assume what child classes would do (add something about their limbs in this case).
Imagine you have a really vast hierarchy of class including limbless creatures like worms. It wouldn't be wise to force every child class to be an owner of anything.

(Yet we assume our inhabitans may have 0 or more friends).

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,105 @@
...
+// Define your objects here
+class Resident {
+  constructor(species, name, gender, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+  }
+  introducing() {
```
OleksiyRudenko, line 18: If you replace `introducing` with magic `toString` you won't need to call this method explicitly in string context. E.g. `print(inhabitant)` would do, `toString` will be called implicitly.

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,105 @@
...
+    super('cat', name, gender, 'meow!');
+    this.friends = ['Tom', 'Jenny', 'Selina'];
+  }
+}
+
+class CatWoman extends Human {
+  constructor(name, gender) {
+    super(name, gender);
+    this.friends = ['Tom', 'Felix'];
+    this.saying = new Cat().saying;
```
OleksiyRudenko, line 73: Nice workaround. Works fine when a class we borrow a property from guarantees the property is set. This particular task implementation is the case.

**You do not need to fix anything in this regards**. Just the the below into consideration.

However, be careful. In general case you would want to borrow a class static property/method rather than a property of an individual `Cat`. Think of it as if you created a cat, took its saying and... well... discarded the cat itself. Doesn't look nice.

Yet another approach is to implement an 'independent' function or variable `meow` and use it to define a relevant property for cats and catwomen.

Yet another approach is to implement composition pattern. Extra materials in the task's repo refers to an article on this topic.

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,105 @@
...
+    'Rex',
+    'Selina',
+  ]),
+  women = new Human('Jenny', 'female', 'Hello Tom!', ['Tom', 'Rex', 'Felix']),
+  dog = new Dog('Rex', 'male'),
+  cat = new Cat('Felix', 'male'),
+  catWoman = new CatWoman('Selina', 'female'),
+  residents = [man, women, cat, dog, catWoman];
+
+residents.map((resident) => print(resident.introducing(), 'p'));
```
OleksiyRudenko, line 96: Do we really need `.map` here? Doesn't seem we use outcome from the mapping. Array methods offer a more appriate method for this use case.

OleksiyRudenko (changes requested):
@5Mountains Good job! 
A few improvements and we're done.

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: Aha, so the best idea that superclass contains only a few major features as name, gender, and species. And other characteristics should be added with the constructor in each extended class?)

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: Okey, I'll try it.

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: About the "compositional" approach, I tried to understand it, but I discovered and started to explore JS OOP through a "Class" pattern and haven't fully covered it yet, so I want to fully understand it first and after other options (:

5Mountains:
Could you please review the new changes?)

submissions/5mountains/tiny-js-world-oop/index.js
OleksiyRudenko: @5Mountains exactly

submissions/5mountains/tiny-js-world-oop/index.js
OleksiyRudenko: @5Mountains ok. Then just leave this part as it is

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: I replaced `map()` method on `forEach()`, it's more appropriate in this case (:

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,115 @@
...
+    this.saying =  'woof-woof!';
+    this.friends = ['Tom', 'Jenny'];
+  }
+}
+
+class Cat extends Animal {
+  constructor(name, gender, saying, friends) {
+    super('cat', name, gender, saying, friends);
+    this.saying = 'meow!';
+    this.friends =  ['Tom', 'Jenny', 'Selina'];
```
OleksiyRudenko, line 80: This says every cat will have Tom, Jenny and Selina as their friends. Not realistic scenario.

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+class Human extends Resident {
+  constructor(name, gender, saying, friends = 'no one') {
+    super('human', name, gender);
+    this.legs = 2;
+    this.hands = 2;
+    this.saying = saying;
+    this.friends = friends;
```
OleksiyRudenko, line 33: Looks like `friends` is a property shared by all inhabitants. Worth having it and handle it at the super class level.

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,115 @@
...
+class Cat extends Animal {
+  constructor(name, gender, saying, friends) {
+    super('cat', name, gender, saying, friends);
+    this.saying = 'meow!';
+    this.friends =  ['Tom', 'Jenny', 'Selina'];
+  }
+}
+
+class CatWoman extends Human {
+  constructor(name, gender, saying) {
```
OleksiyRudenko, line 85: Looks like CatWoman **actually** has no choice but being a female.

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,115 @@
...
+}
+// ======== OUTPUT ========
+/* Use print(message) for output.
+Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+However, please, REFRAIN from improving visuals at least until your code is reviewed
+so code reviewers might focus on a single file that is index.js.
+*/
+const man = new Human('Tom', 'male', 'Hello Jenny!', ['Tom','Jenny','Rex', 'Selina']),
```
OleksiyRudenko, line 99: Friends list here is just a list of names, i.e. strings. How about referring real inhabitan objects?
Note that you may assign (or reassign) friends after an object is being created. Just like in real life.

Also consider using empty array `[]` instead of string `'no friends'`. Imagine you need traverse inhabitant's friends for multiple purposes. You'd need then use logic `if (Array.isArray(this.friends))...` or `if (this.friends !== 'no friends')...` everywhere. Instead, you will have special handling of an empty array in a single method - `toString`.

OleksiyRudenko (changes requested):
@5Mountains well done!
There is still room for improvement.
Whenever you define or implement anything ask yourself: what if I need to create hundreds of inhabitants? What if I need to add species? How costly would that be?

In other words think of scalability and maintainability.

OleksiyRudenko:
@5Mountains do you need any help or support on this?

5Mountains:
Greetings! Is there still an opportunity for me to participate and finish the assignment? There were unexpected family circumstances and another project that caused me to lose so much...

OleksiyRudenko:
No worries. Keep going.

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: I think I fixed it :)

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: Done

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: remove this method from the constructor and set it by default (:

5Mountains:
Updated solution. Improved (as it seems to me) inheritance and classes.

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: Honestly, I did not understand the implementation of the link directly to objects *( About the empty array - I'll change (:

submissions/5mountains/tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,102 @@
...
+class Cat extends Animal {
+  constructor(name, gender, saying, friends) {
+    super('cat', name, gender, saying, friends);
+  }
+}
+
+class CatWoman extends Human {
+  constructor(name, saying, friends = ['Tom', 'Felix']) {
+    super(name, 'female', saying, friends);
+    this.saying = CatWoman.saying();
```
OleksiyRudenko, line 69: `saying` received by the class constructor is sent over its super and then overwritten.
1. Do we need to CatWoman constructor to expect any saying at all? Imagine a peer dev who does `new CatWoman('Kate', 'Meohow!')` and this brings no effect.
2. If super assign saying why not sending value for it instead of overwriting the property?

`callFunction(5, varName, someFunction())` is pretty valid syntax, `callFunction` will receive `5`, value of `varName` and whatever `someFunction()` returns.

OleksiyRudenko (changes requested):
@5Mountains pretty well done!
Let's polish it.

submissions/5mountains/tiny-js-world-oop/index.js
5Mountains: Aha, I think I got it, in this case, I should totally remove this.saying property inside CatWoman class and send the way of calling this static method into parent constructor via super() ? :)

OleksiyRudenko: try it

5Mountains: It's alive! I feel like a mad scientist!

OleksiyRudenko (approved):
@5Mountains well done!
Approved with a reservation.
I see how did you tricked a catwoman to have friends omitting useless saying at construction time in your past version of implementation.
This wasn't efficient approach as there are no good reasons for every cat-woman have default friends. At least it is not consistent with other inhabitants behaviour.
The learning for you to take from here: when solution to a problem requires a trick then the solution is not good.

5Mountains:
I completely agree with you that this is a tricky move! I have tried different approaches on how to correctly build CatWoman object since I inherit everything from Human - the constructor expects me to pass into it, including - gender and greeting properties. Due to the fact, that I don't pass these parameters it all breaks and when I pass friends it appears in parameter gender *( so I used such a trick...

PR #99, opened 2020-11-09
Tiny JS World (pre-OOP)

Iakow:
### A Tiny JS World (pre-OOP)
[Demo](https://iakow.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Iakow/frontend-2021-homeworks/blob/tiny-js-world/submissions/iakow/a-tiny-js-work/index.js)

Please, review.

### DIFF
--- a/submissions/iakow/a-tiny-js-work/index.js
+++ b/submissions/iakow/a-tiny-js-work/index.js
@@ -48,7 +48,7 @@ woman.friends = [man, cat];
 dog.friends = [man, woman, catWoman, cat];
 catWoman.friends = [cat, woman];
 
-const printInhabitant = obj => {
+const formHabitantСertificate = obj => {
    const props = Object.keys(obj).map(key => {
       if (key === 'friends') {
          return obj[key].map(friend => friend.name).join(', ');
@@ -57,7 +57,7 @@ const printInhabitant = obj => {
       return obj[key];
    })
 
-   print(props.join('; '));
+   return props.join('; ');
 }
 
-[man, woman, cat, dog, catWoman].forEach(printInhabitant);
+[man, woman, cat, dog, catWoman].forEach((obj) => print(formHabitantСertificate(obj)));
--- a/submissions/iakow/a-tiny-js-work/index.js
+++ b/submissions/iakow/a-tiny-js-work/index.js
@@ -1,0 +1,63 @@
+const man = {
+   species: 'human',
+   name: 'Donald',
+   gender: 'male',
+   legs: 2,
+   hands: 2,
+   saying: 'Aloha!',
+};
+
+const woman = {
+   species: 'human',
+   name: 'Eva',
+   gender: 'female',
+   legs: 2,
+   hands: 2,
+   saying: 'Hello',
+};
+
+const cat = {
+   species: 'cat',
+   name: 'Murka',
+   gender: 'female',
+   legs: 4,
+   hands: 0,
+   saying: 'Meow...',
+};
+
+const dog = {
+   species: 'dog',
+   name: 'Gektor',
+   gender: 'male',
+   legs: 4,
+   hands: 0,
+   saying: 'WOFFF!',
+};
+
+const catWoman = {
+   species: 'humacat',
+   name: 'Akrum',
+   gender: 'female',
+   legs: 2,
+   hands: 2,
+   saying: cat.saying,
+};
+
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@Iakow please edit the PR title so that it would be clear what task it is about from the list of the PRs.

Iakow:
Sorry, I overlooked.  Now it is better?

OleksiyRudenko:
I meant this
image

Please check how it is different from other PR titles in a list. It is always good to see a book's title before you take a look inside it.
image

submissions/iakow/a-tiny-js-work/index.js
```diff
@@ -0,0 +1,63 @@
...
+   const props = Object.keys(obj).map(key => {
+      if (key === 'friends') {
+         return obj[key].map(friend => friend.name).join(', ');
+      }
+
+      return obj[key];
+   })
+
+   print(props.join('; '));
+}
```
OleksiyRudenko, line 61: `printInhabitant` does two major things: (1) constructs an inhabitant's presentation string AND (2) prints it.
Following [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) makes code scalable, maintainable and less error prone.
Please refactor this.

submissions/iakow/a-tiny-js-work/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+[man, woman, cat, dog, catWoman].forEach(printInhabitant);
```
OleksiyRudenko, line 63: `.forEach(printInhabitant)` is a nice shortcut if compared with `.forEach(obj => printInhabitant(obj))`
You might keep it that concise if you introduce `toString` method to every object (effectively bounding `printInhabitant` function to `toString` object method).
This, however will affect they way you enumerate object's own properties.

**You do not need to implement this now.** You will have a chance yet after the subsequent task.

Getting familiar with [a built-in `toString` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString) will be sufficient at this stage (check **Examples** section).

OleksiyRudenko (changes requested):
@Iakow Good job!
Making code SRP-compliant will make it great.
Please check the comments below.

submissions/iakow/a-tiny-js-work/index.js
Iakow: I think, done.

submissions/iakow/a-tiny-js-work/index.js
Iakow: Thanks. Looked at examples and I have one question, if it's convenient for you. Why is a named function rather than an anonymous one used when overriding toString()?

Iakow:
> I meant this...

Felt dumb) Done.

submissions/iakow/a-tiny-js-work/index.js
OleksiyRudenko: Good question. Have no idea why they used a named function in the examples.
The only valid, to my knowledge, use case of a named function on the right side of an assignment is when that function needs to call itself recursively.
E.g. `const func = function rfunc(n){if (n=0) return; rfunc(n-1);};`
`func` is not defined yet and one needs a defined name to call the function from inside itself.

submissions/iakow/a-tiny-js-work/index.js
```diff
@@ -0,0 +1,63 @@
...
+         return obj[key].map(friend => friend.name).join(', ');
+      }
+
+      return obj[key];
+   })
+
+   return props.join('; ');
+}
+
+[man, woman, cat, dog, catWoman].forEach((obj) => print(formHabitantСertificate(obj)));
```
OleksiyRudenko, line 63: With arrow function you don't need to have a single param parenthesised.
I.e. `.forEach(obj => print(...))` would work.
Not a strict rule though. Just FYI
While zero arguments and 2+ arguments require parentheses.

OleksiyRudenko (approved):
@Iakow well done!

OleksiyRudenko:
> > I meant this...
> 
> Done.

No problem. You've just learnt something new about giving more useful and specific information :)

submissions/iakow/a-tiny-js-work/index.js
Iakow: Thanks for the answer!

PR #101, opened 2020-11-09
Add: memory-pair-game task by andrewklmn

andrewklmn:
### Memory Pair Game

[Demo](https://andrewklmn.github.io/memory-pair-game/) |
[Code base](https://github.com/andrewklmn/frontend-2021-homeworks/tree/memory-pair-game/submissions/andrewklmn/memory-pair-game)

Please, review.


OleksiyRudenko (changes requested):
@andrewklmn a visually appealing implementation.
However, this PR contains commits from different features. Please fix.

andrewklmn:
> @andrewklmn a visually appealing implementation.
> However, this PR contains commits from different features. Please fix.

Ups!  My bad... I have rebased this branch without commits from the previous task now...

OleksiyRudenko:
@andrewklmn please try avoiding force pushing if it is not absiolutely required to fix major issues.
It is not really good practice. Why?
Beside other no less improtant good reasons in context of code review/PR it affects UX for both reviewers and students. E.g. code comments may go lost. When one sees a force push it may be hard to say if there were any comments that might have gone lost.

andrewklmn:
@OleksiyRudenko  Got it!... I will check my code twice before every 'push' from this moment.
No more  'push -f'!...


> @andrewklmn please try avoiding force pushing if it is not absiolutely required to fix major issues.
> It is not really good practice. Why?
> Beside other no less improtant good reasons in context of code review/PR it affects UX for both reviewers and students. E.g. code comments may go lost. When one sees a force push it may be hard to say if there were any comments that might have gone lost.

submissions/andrewklmn/memory-pair-game/app.js
```diff
@@ -0,0 +1,189 @@
...
+const spreadCards = function () {
+  
+  const backImages = gameBoard.querySelectorAll('.back > img');
+
+  imageNames.sort(cardsShuffler);
+
+  backImages.forEach((image,i) => {
+    image.src = `img/${imageNames[i]}`;
+    image.parentNode.parentNode.parentNode.id = imageNames[i];
+  });
```
madmaxWMFU, line 92: In my opinion it would be better to shuffle the cards before drawing and not do double work.

madmaxWMFU (changes requested):
Nice job!
See comments below

submissions/andrewklmn/memory-pair-game/app.js
andrewklmn: > In my opinion it would be better to shuffle the cards before drawing and not do double work.

@madmaxWMFU  You are right.  Shuffle function at the beginning can be called twice.
The first time in the initialization function, and the second time in the Start button listener.

There are two ways to start a new game after page opening:
1) you can just click any closed card and the game will be started immediately without preview.
2) you can press the Start button with the same result, but with (or without) a preview before the game.

So, cards must be already shuffled in the first case... 
That is why I put the shuffling call in the initialization function.

It is needed only for the first new game which can be started just by clicking any of the closed cards.
But, after this first game, cards will be shuffled only when the Start button will be pressed just before a new game.

andrewklmn:
@madmaxWMFU  
> In my opinion it would be better to shuffle the cards before drawing and not do double work.

Do you mean to move shuffling call to initBoard function before drawing? Like this: 
https://github.com/kottans/frontend-2021-homeworks/pull/101/commits/d34e7f31f183bdaa09d568158bad52578a04c723

submissions/andrewklmn/memory-pair-game/app.js
andrewklmn: It seems I understand what you mean... )))

andrewklmn:
@madmaxWMFU
> In my opinion it would be better to shuffle the cards before drawing and not do double work.

And what do you think about this way? 
https://github.com/kottans/frontend-2021-homeworks/pull/101/commits/66f7021355f93cedf10be51039c29d9741fe8877

submissions/andrewklmn/memory-pair-game/app.css
```diff
@@ -0,0 +1,123 @@
...
+  width:100%;
+  cursor: pointer;
+}
+
+.guessed {
+  opacity: 0.3;
+}
+
+.hide {
+  opacity: 0;
```
madmaxWMFU, line 45: If we hide the control wrapper, we can find it with the cursor and reset the game again.
Use `display: none` and paste this rule after `.control`

submissions/andrewklmn/memory-pair-game/index.html
```diff
@@ -0,0 +1,24 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <link rel="stylesheet" href="app.css">
+  <title>Memory Pair Game</title>
+</head>
+<body>
+  <div class="info">
+      <span class="info-line">Enable Javascript!</span>
+  </div>
+  <div class="gameboard">
+    <img class="card" src="img/empty.png" alt="Back">
```
madmaxWMFU, line 14: Do you need this tag here?

submissions/andrewklmn/memory-pair-game/app.js
```diff
@@ -0,0 +1,190 @@
...
+
+  gameBoard.innerHTML = '';
+
+  imageNames.forEach((image)=>{
+    
+    const fragment = document.createDocumentFragment();
+
+    const flipContainer = document.createElement('div');
+    flipContainer.classList.add('flip-container')
+    flipContainer.id = image;
```
madmaxWMFU, line 51: It is not a good name for id, use `dataset`

submissions/andrewklmn/memory-pair-game/app.js
```diff
@@ -0,0 +1,207 @@
...
+  if (openedCards.length == 2) {
+    if (openedCards[0].dataset.label == openedCards[1].dataset.label) {
+      setTimeout(()=>openedCards.forEach(c=>{
+        c.classList.add('guessed');
+      }),500);
+      setTimeout(() => {
+        if(getGuessedCards().length==16) {
+          showInfo(`You won! Number of fails: ${numberOfFails}`);
+          openAllCards();
+          document.querySelector('.control').classList.remove('hide');
```
madmaxWMFU, line 145: You also have similar content, `controlDiv`

madmaxWMFU (changes requested):
Add full naming of values in cycles `forEach((card, index)=>{})`

submissions/andrewklmn/memory-pair-game/app.js
```diff
@@ -0,0 +1,207 @@
...
+const getGuessedCards = function() {
+  return gameBoard.querySelectorAll('.guessed');
+}
+
+const showInfo = (text) => {
+  infoDiv.innerHTML = text;
+}
+
+const cardClickListener = function(e) { 
+  if (getOpenedCards().length == 16) return;
```
madmaxWMFU, line 121: A magic number, use `maxOpenCards`

madmaxWMFU:
As for me, there are many `setTimeout()` functions but everything works fine.
Change the numeric values ​​to a constant in all the remaining functions `setTimeout()`. Then make an approve.

madmaxWMFU (approved):
LGTM 👍🏻

andrewklmn:
> LGTM 👍🏻

@madmaxWMFU 
Thank you for your help! )))

OleksiyRudenko (approved):
@andrewklmn great job!

PR #103, opened 2020-11-10
Frogger game

krylenger:
### Frogger game

[Demo](https://krylenger.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/krylenger/frontend-nanodegree-arcade-game/tree/frogger)

Appreciate your review in advance!
Have fun :)


submissions/krylenger/frogger/app.js
```diff
@@ -0,0 +1,246 @@
...
+
+const Enemy = function() {
+    this.width = 70;
+    this.height = 83;
+    this.speed = randomSpeed();
+};
+
+Enemy.prototype.handleGameOver = function () {
+  if (
+    this.x + this.width > player.x &&
```
AMashoshyna, line 30: Classes should not refer to any global variables, like global variable `player`, which is an instance of Player class
(referring to global constants and globals provided by the gaming platform like Resources is OK);
Hint: pass Player instance as an argument to every enemy

submissions/krylenger/frogger/app.js
```diff
@@ -0,0 +1,246 @@
...
+    player.x + player.width > this.x &&
+    player.y > this.y - this.height &&
+    player.y - player.height < this.y
+  ) {
+    setTimeout(() => {
+      player.lives -= 1;
```
AMashoshyna, line 36: This code to me looks too long to be `setTimeout` inline handler. How about extracting it into a function (or maybe several functions)?

submissions/krylenger/frogger/app.js
```diff
@@ -0,0 +1,246 @@
...
+
+const tileDimensions = {
+    width: 100,
+    height: 83
+}
+
+const score = document.querySelector('.score');
+const lives = document.querySelector('.lives');
+
+const randomSpeed = () => {
```
AMashoshyna, line 16: Usually function names should contain a verb (this way it is easier to identify them when reading code). How about `getRandomSpeed`?

submissions/krylenger/frogger/app.js
```diff
@@ -0,0 +1,246 @@
...
+    this.y = y;
+    this.sprite = 'images/char-horn-girl.png'
+}
+
+EnemyGirl.prototype = Object.create(Enemy.prototype);
+EnemyGirl.prototype.constructor = EnemyGirl;
+
+EnemyGirl.prototype.update = function(dt) {
+    this.x += this.speed * dt;
+    if (this.x > 500) {
```
AMashoshyna, line 92: Those numbers (500, 150) look a little magical. I would put them into properly named constants.

submissions/krylenger/frogger/app.js
```diff
@@ -0,0 +1,246 @@
...
+        }, 13);
+    }
+}
+
+let star = new Star();
+let heart = new Heart();
+
+
+document.addEventListener('keyup', function(e) {
+    var allowedKeys = {
```
AMashoshyna, line 238: Let's avoid mixing `var` with `let/const` with not special reason

AMashoshyna (changes requested):
Very good job! Please see my comments.

submissions/krylenger/frogger/app.js
krylenger: ✅  Done

submissions/krylenger/frogger/app.js
krylenger: ✅ Done

submissions/krylenger/frogger/app.js
krylenger: ✅ Done

submissions/krylenger/frogger/app.js
krylenger: ✅ Done

submissions/krylenger/frogger/app.js
krylenger: Thank you for your advices so much!
This helped me to understand the key principles of OOP pattern better and to see some weak and non logical parts in my code (such as some methods of Player class were attached to Enemy class).
What about this part: 
`Player.prototype.resetGame = function() {
  alert(`You lose :( Your score is ${this.score} Another try?`);
  this.lives = 3;
  this.score = 0;
  score.textContent = `Score: ${this.score}`;
  lives.textContent = `Lives: ${this.lives}`;
}`
is it fine to update score and lives global variables which are DOM elements from inside the Class?

AMashoshyna: This is a very good question. I think this is out of scope of this particular task (your solution looks fine to me), but in a general case you would probably have a dedicated entity for score board with methods like `setScore` and `setLivesCount`, which would update DOM. Speaking about DOM updates: this is a heavy operation so in modern applications you want to make as little DOM manipulations as possible, so after score/life update you would probably re-render the score board in one go (not score and then lives separately).

AMashoshyna (approved):

PR #106, opened 2020-11-11
a-tiny-JS-world  post_OOP

krylenger:
### a-tiny-JS-world  post_OOP

[Demo](https://krylenger.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/krylenger/a-tiny-JS-world/tree/postOOP)

I hope that I got the main idea of composition. 
Thanks for your review in advance!


submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,140 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details */
+
+const dog = {
+    type: "animal",
+    legs: 4,
+    hands: 0,
+    gender: "male",
+    friends: ["Anna", "Eugene"],
```
OleksiyRudenko, line 8: So, all dogs in this world will all be males and will all have Anna and Eugene as their friends?
Doesn't look really realistic. 
Can we allow more life-like customization?
There is no much point of defining class templates to create just one instance of each inhabitant class.

Please apply the same to all classes.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,140 @@
...
+  },
+  cat = {
+    type: "animal",
```
OleksiyRudenko, line 11: Task requires to expose species of each inhabitant. "animal" is not a species.

Please apply the same to all classes.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,140 @@
...
+    legs: 4,
+    hands: 0,
```
OleksiyRudenko, line 13: Here we say a typical cat has 4 legs and 0 hands. In fact cats do not have hands at all.
The beaty of OOP concepts is that in class hierarchy allows super (parent) classes to have properties shared by all child classes and child classes to have properties specific to them.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,140 @@
...
+      self.type
+    }</em>. 
+      woof woof <em>${self.gender}</em>. woof <em>${self.legs}</em> woof <em>${
+      self.hands
+    }</em> woof. 
+      Woof woof woof: <em>${self.friends.join(", ")}</em>.<em> 
+      </em><br><br>`;
+  },
+});
+
```
OleksiyRudenko, line 93: Quite a creative idea to have distinct communications for humans, dogs, and cats.
There is a pattern in `catSays` and `dogSays`. How about having a single yet customizable method for animals? This way adding more species of animals cheaper.

In other words how would you change your code in this section if you were asked to introduce other species of animals?

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,140 @@
...
+const womanAnna = createWoman("Anna");
+const catWomanSofia = createCatWoman("Sofia");
+
+const introductions = [
+  manEugene.introduction(),
+  womanAnna.introduction(),
+  catBarsik.introduction(),
+  catWomanSofia.introduction(),
+  dogSharik.introduction(),
+];
```
OleksiyRudenko, line 136: There is some code repetition here. We might save notable efforts if reduced repetition of `.introduce()`.
Hint: `.map()`

OleksiyRudenko (changes requested):
@krylenger Well done!
Normally we'd expect a contributor to employ es6 class syntax. You took a challenge to employ composition pattern. Respects.
Composition pattern poses its own challenges. Yet we need to follow OOP, SOLID and DRY principles (e.g. add properties where they are truly needed, reduce code repetition etc).

Let's improve the code that the solution becomes great!

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: ✅ Done

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: ✅ Done

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: ✅ Done
How do you find this implementation?
I mean the part of modifying the carnivoraPetSays function for catWoman. 
My concern that the part with `if if (self.species === 'catWoman')` doesn't look very nice but I need to use the same introduction method for catWoman and she doesn't have paws.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: ✅ Done

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: ✅ Done

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
OleksiyRudenko: @krylenger well, yes, looks not really nice. The problem is that it is an all-in-one function versus [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle). Split it.

First, split out methods to assign sayings to an inhabitants of this type (animals ans animal-like).

Then have a method for output creation. Obviously CatWoman should employ a method for humans in these regards.

The beauty of compositional design pattern is that one can combine small or even tiny chunks of properties and methods in order to achieve the goals yet keeping code simple and concise.

Those chunks should stay as abstract (agnostic of a specific species in your case) as possible. Whenever a composable method needs to know what entity class method or property it will be attributed to it means that the implementation is poor.

Do you want to take this challenge? We might get finished with what's already in place.

OleksiyRudenko (commented):
@krylenger the composition is a challenge at times.
Do you want to take it?
Please check the comment below.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: I really appreciate your feedback!
Certainly I want to take this challenge to understand the key principles better.
After your comments I refactored my classes and methods on them to follow composition design pattern. 
Now comparing to the previous version I can clearly see the weak parts of my previous code.
What about now? How do you find it? Are there some more work to be done to improve this code? Thanks in advance.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,126 @@
...
+const manEugene = createMan("Eugene", ['Anna', 'Eugene']);
+const womanAnna = createWoman("Anna", ['Anna', 'Eugene']);
+const catWomanSofia = createCatWoman("Sofia", ['Anna', 'Eugene']);
+
+const inhabitants = [manEugene, womanAnna, catBarsik, catWomanSofia, dogSharik];
+const introductions = inhabitants.map(inhabitant => inhabitant.introduction())
+
+introductions.forEach((introduction) => {
+  print(introduction, "pre");
+});
```
OleksiyRudenko, line 126: Let's chain this up like
```
inhabitans
  .map(...)
  .forEach(...);
```

BTW, if you rename `introduction` into `toString` then in string context object reference will `toString` implicitly and will make it possible to `inhabitans.forEach(inhabitant => print(inhabitant))` 

If `print` declaration matched `forEach`'s callback specification we even could make it even shorter: `inhabitants.forEach(print)`. Unfortunately, `print` has the second param that doesn't match the specs.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+const createWoman = (name, friends) => {
+  const self = createHuman(woman);
+  self.gender = "female";
+  return Object.assign(self, {name, friends});
+};
+
+const createCatWoman = (name, friends) => {
+  const self = createWoman(catWoman);
+  self.species = "catWoman";
```
OleksiyRudenko, line 111: Why `const self = createWoman(catWoman);` is not sufficient?

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
```diff
@@ -0,0 +1,126 @@
...
+  return Object.assign(self, {name, friends});
+};
```
OleksiyRudenko, line 113: Does cat-woman speak cattish?
You might introduce `meower` and use it when composing a Cat and a CatWoman instead of having a hardcoded `self.word` in `createCat`.

OleksiyRudenko (changes requested):
@krylenger the code reads now much better!
Let's make the code great by fixing a few inconsistencies.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: Got it! 
✅ Done

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: It's no sufficient because I instantiate from Woman class and don't have access to methods of Cat (e.g. output method) 
I suggest that solution is to compose catWoman from Inhabitant class and add all required methods and parameters. 
Is it a correct approach?

krylenger: Or even instantiate catWoman from Woman but also compose meower() and carnivoraPetSays to have all required properties and methods.

submissions/krylenger/a-tiny-JS-world_post_OOP/index.js
krylenger: I created meower function to assign the parameter 'word' to our brand new created inhabitants.
And now I have only 2 lines of code while creating my inhabitant classes! wow, looks great! thanks for that!  

Also I've changed the output method for catWoman and she speaks cattish now.

OleksiyRudenko (approved):
@krylenger Great job! Approved.

P.S. You do not need to fix this now, but it might be important in your next submissions.
GitHub marked your file with **No newline at end of file**
image
Visually, there is a newline. This might be a signal that your editor and/or git settings have wrong character sequence for newlines. By default a Linux newline should be committed (LF), you might have something different (CR if you're working on a Mac, or CRLF if you're working on Windows).

Recommended approach is to have LF in your IDE settings and `core.autocrlf=input` in your global git config. Check [this](https://jenyay.net/Git/Autocrlf) out.

PR #119, opened 2020-11-13
OOP Exercise

HannaSyn:
### OOP Exercise

[Demo](https://hannasyn.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/HannaSyn/a-tiny-JS-world/blob/post_oop/index.js)

Please, review.

### DIFF
--- a/submissions/HannaSyn/oop-exercise/index.js
+++ b/submissions/HannaSyn/oop-exercise/index.js
@@ -16,7 +16,7 @@ class Inhabitant {
       this.friends = friends;
       this.say = say
    }
-   speak() {
+   toString() {
       return `${this.say} I'm ${this.biologicalSpecies}! My name is ${this.name}, my gender is ${this.gender}, and I have friends: ${this.friends}.`;
    }
 }
@@ -24,47 +24,49 @@ class Inhabitant {
 class Human extends Inhabitant {
    constructor( name, gender, friends, say, legs, hands ) {
       super( name, gender, friends, say, 'human');
-      this.legs = legs;
-      this.hands = hands;
+      this.legs = 2 || legs;
+      this.hands = 2 || hands;
    } 
-   speak() {
-      return super.speak() + ` I have ${this.legs} legs and ${this.hands} hands.`
+   toString() {
+      return super.toString() + ` I have ${this.legs} legs and ${this.hands} hands.`
    }
 }
 
 class Animal extends Inhabitant {
    constructor( name, gender, friends, say, biologicalSpecies, paws) {
       super( name, gender, friends, say, biologicalSpecies )
-      this.paws = paws;
+      this.paws = 4 || paws;
    }
-   speak() {
-      return super.speak() + ` I have ${this.paws} paws.`
+   toString() {
+      return super.toString() + ` I have ${this.paws} paws.`
    }
 }
 
 class Cat extends Animal {
    constructor( name, gender, friends, say, paws) {
       super( name, gender, friends, say, 'cat', paws );
+      this.say = 'Meow!' || say;
    } 
 }
 
 class Dog extends Animal {
-   constructor( name, gender, friends, say, paws, Human ) {
+   constructor( name, gender, friends, Human, say, paws) {
       super( name, gender, friends, say, 'dog', paws);
       this.master = Human;
+      this.say = 'Woof!' || say;
    } 
-   speak() {
-      return super.speak() + ` I love my master ${this.master.name} very much!`
+   toString() {
+      return super.toString() + ` I love my master ${this.master.name} very much!`
    }
 }
 
 
-const woman = new Human('Hanna', 'female', ['Vlad', 'Ernest', 'Emilia'], 'Hi!', 2, 2);
-const man = new Human('Vlad', 'male', ['Hanna', 'Ernest', 'Emilia'], 'Hello!', 2, 2);
-const cat = new Cat('Ernest', 'male', ['Vlad', 'Hanna', 'Emilia'], 'Meow!', 4);
-const dog = new Dog('Emilia', 'female', ['Vlad', 'Ernest'], 'Wooof!', 4, woman);
+const woman = new Human('Hanna', 'female', ['Vlad', 'Ernest', 'Emilia'], 'Hi!');
+const man = new Human('Vlad', 'male', ['Hanna', 'Ernest', 'Emilia'], 'Hello!');
+const cat = new Cat('Ernest', 'male', ['Vlad', 'Hanna', 'Emilia']);
+const dog = new Dog('Emilia', 'female', ['Vlad', 'Ernest'], woman);
 
-[woman, man, cat, dog].forEach(inhabitant => print(inhabitant.speak()));
+[woman, man, cat, dog].forEach(inhabitant => print(inhabitant));
 
 // ======== OUTPUT ========
 /* Use print(message) for output.
--- a/submissions/HannaSyn/oop-exercise/index.js
+++ b/submissions/HannaSyn/oop-exercise/index.js
@@ -1,0 +1,87 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitans {
+   constructor ( name, gender, friends, say, biologicalSpecies ){
+      this.name = name;
+      this.biologicalSpecies = biologicalSpecies;
+      this.gender = gender;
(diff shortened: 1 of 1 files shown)


submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitans {
```
OleksiyRudenko, line 11: Normally, plural form is used to denote collections. This class creates a single instance of an inhabitant.

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+class Human extends Inhabitans {
+   constructor( name, gender, friends, say, legs, hands ) {
+      super( name, gender, friends, say, 'human');
+      this.legs = legs;
+      this.hands = hands;
+   } 
+
+   speak() {
+      return ` ${this.say} I'm ${this.biologicalSpecies}! My name is ${this.name}, my gender is ${this.gender}, I have ${this.legs} legs and ${this.hands} hands and I have friends: ${this.friends}.`;
+   }
```
OleksiyRudenko, line 31: This child class method handles properties of its super class. Let the super class do it itself.
You may re-use methods from superclass, the pattern is `super.methodName()`

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+   purr() {
+      return ` ${this.say}  I'm ${this.biologicalSpecies}! My name is ${this.name}, my gender is ${this.gender}, I have ${this.paws} paws and I have friends: ${this.friends}.` 
+   }
+}
+
+class Dog extends Inhabitans {
+   constructor( name, gender, friends, say, paws, Human ) {
+      super( name, gender, friends, say, 'dog' );
+      this.paws = paws;
```
OleksiyRudenko, line 49: Dogs and cats have a shared property `paws`. A clear sign that you need a superclass for them to handle the shared property. Please implement.

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+   
+}
+
+class Cat extends Inhabitans {
+   constructor( name, gender, friends, say, paws) {
+      super( name, gender, friends, say, 'cat' );
+      this.paws = paws;
+   } 
+
+   purr() {
```
OleksiyRudenko, line 41: It's nice to have a clearly specific method.
However for the purposes of code conciseness add a `speak` method (which may call `super.speak()`). `purr` may then just call `this.speak()`. 
This will make every inhabitant handling similar later on.

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+const woman = new Human('Hanna', 'female', ['Vlad', 'Ernest', 'Emilia'], 'Hi!', 2, 2);
+const man = new Human('Vlad', 'male', ['Hanna', 'Ernest', 'Emilia'], 'Hello!', 2, 2);
+const cat = new Cat('Ernest', 'male', ['Vlad', 'Hanna', 'Emilia'], 'Meow!', 4);
+const dog = new Dog('Emilia', 'female', ['Vlad', 'Ernest'], 'Wooof!', 4, woman);
+
+print(woman.speak());
+print(man.speak());
+print(cat.purr());
+print(dog.bark());
```
OleksiyRudenko, line 66: Having all classes defined `speak` method you code would be here as short as `[woman, man, dog, cat].forEach(inhabitant => print(inhabitant.speak()));`
Imagine, you have hundreds of inhabitants what would code look like if you stick to the current approach.

OleksiyRudenko (changes requested):
@HannaSyn well done!
Nice hirarchy of classes.
However there is still room for improvement.
Think of scalability challenges (hundreds of inhabitants, dozens of species etc)

submissions/HannaSyn/oop-exercise/index.js
HannaSyn: Thank you for suggestions! I improved my code follow your advices.

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,88 @@
...
+      super( name, gender, friends, say, biologicalSpecies )
+      this.paws = paws;
+   }
+   speak() {
+      return super.speak() + ` I have ${this.paws} paws.`
+   }
+}
+
+class Cat extends Animal {
+   constructor( name, gender, friends, say, paws) {
```
OleksiyRudenko, line 46: How about default values for `paws` and e.g. `say`?
Same might apply for other animals.
Humans might have default hands and legs count.
This approach would allow a developer to define non-standard counts of limbs at construction time, hopefully saving their effort.
This might require you to reconsider order of arguments for dogs.

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,88 @@
...
+   speak() {
+      return super.speak() + ` I love my master ${this.master.name} very much!`
+   }
+}
+
+
+const woman = new Human('Hanna', 'female', ['Vlad', 'Ernest', 'Emilia'], 'Hi!', 2, 2);
+const man = new Human('Vlad', 'male', ['Hanna', 'Ernest', 'Emilia'], 'Hello!', 2, 2);
+const cat = new Cat('Ernest', 'male', ['Vlad', 'Hanna', 'Emilia'], 'Meow!', 4);
+const dog = new Dog('Emilia', 'female', ['Vlad', 'Ernest'], 'Wooof!', 4, woman);
```
OleksiyRudenko, line 65: Well, here friends list is just a set strings. What if want to send over real inhabitants as friends? This would give us liberty to describe the friends as we want them not limiting ourselves to names only.
Just like you did with a master for a dog.

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+[woman, man, cat, dog].forEach(inhabitant => print(inhabitant.speak()));
```
OleksiyRudenko, line 67: Now you have all inhabitants a `speak` method. Good.
If you rename `speak` to `toString` it would be implictly called when referring an object in string context.
I.e. `print(inhabitant)` will be equivalent to `print(inhabitant.toString)`. It's a magic method for all object. This method by default produces `[Object object]` when `alert(object)`

OleksiyRudenko (commented):
@HannaSyn great job!
I am ready to approve this unless you take a challenge outside the requirements.
Challenging matters are described below.
Please let me know if you want to take it. If not then I will aprove the code and you'd have more time to complete the course tasks that follow.

Alternative: we may approve this submission and you'd revert to this task with a new PR later if and when you have time.

submissions/HannaSyn/oop-exercise/index.js
HannaSyn: Thank you! I've really surprised `toString` method.

OleksiyRudenko: > I fixed all except friends

Hint: template strings allow not only variables but any valid expressions
 `` `.... ${this.friends.map(friend => frind.name).join(", ")}` ``

Not necessarily a good example in terms of readability but at least FYI

submissions/HannaSyn/oop-exercise/index.js
```diff
@@ -0,0 +1,90 @@
...
+   } 
+   toString() {
+      return super.toString() + ` I have ${this.legs} legs and ${this.hands} hands.`
+   }
+}
+
+class Animal extends Inhabitant {
+   constructor( name, gender, friends, say, biologicalSpecies, paws) {
+      super( name, gender, friends, say, biologicalSpecies )
+      this.paws = 4 || paws;
```
OleksiyRudenko, line 38: This is a valid approach.
However do not forget that `constructor` is a function so it allows for default values `constructor(a, b, c = 5) { ... }`
Well, the latter approach doesn't work well when you have dafaults for the arguments in the middle. This means that all args with defaults should be in the end of args list.
You may also combine both approaches.
Just FYI

HannaSyn: Should I fix this?

HannaSyn:
@OleksiyRudenko tell me, please, what part I should fix to get merge this PR?

OleksiyRudenko:
@HannaSyn assuming this is the answer to my question from this [comment](https://github.com/kottans/frontend-2021-homeworks/pull/119#pullrequestreview-531287595) :)
Approved.

OleksiyRudenko (approved):
Well done!

PR #120, opened 2020-11-13
A Tiny JS World

7SergSmith7:

### task name

[Demo](https://7sergsmith7.github.io/a-tiny-JS-world/index.html) |
[Code base](https://github.com/7SergSmith7/a-tiny-JS-world)

Please, review.


OleksiyRudenko:
@7SergSmith7 thank you for your submission.
In this task we need only `index.js` to conduct a review. Can you please have that file only in the PR?

OleksiyRudenko:
We've got a confusing label here. And placeholder **task name** in initial comment is not replaced with real task name.

Is your submission a result from completion of [Building a Tiny JS World](https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md) or [OOP exercise](https://github.com/kottans/frontend/blob/master/tasks/js-post-oop.md)?

7SergSmith7:
> We've got a confusing label here that doesn't really matches the code submitted. And placeholder **task name** in initial comment is not replaced with real task name.
> 
> Is your submission a result from completion of [Building a Tiny JS World](https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md) or [OOP exercise](https://github.com/kottans/frontend/blob/master/tasks/js-post-oop.md)?

I did the first task, but I wanted it to be really right away for the second task.

OleksiyRudenko:
Clear, thanks. Will take this into consideration.

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+    this.name = name;
+    this.gender = gender;
+    this.phrase = phrase;
+    this.friends = friends;
+    //this.addFriends(friends);
+  }
+
+  addFriends(newFriend) {
+    this.friends.push(newFriend);
+  }
```
OleksiyRudenko, line 24: How about making this method so it could accept both a single `friend` and an array of `friends` as its parameter?
This way you really could call it from the constructor for consistency. Yet you'd need to initialize the property with an empty array, again for consistency.

Hints:
`Array.isArray()`
`typeof`

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Habitant {
+  constructor(species, name, gender, phrase, friends = [], legs, hands) {
+    this.species = species;
+    this.legs = legs;
+    this.hands = hands;
```
OleksiyRudenko, line 14: Not every Habitant has hands.
The beuaty of OOP approach is that each super (parent) class defines and handles only shared properties, and each child class introduces and handles properties unique to that child class.
Imagine, you'd need add birds, worms etc. What super class would look like then?
Let's however stick to mammals.

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  toString() {
+    return `Hi! I am <strong>${this.species}</strong>, my name is <strong>${
+      this.name
+    }</strong>, my gender is <strong>${this.gender}</strong>, I have <strong>${
+      this.legs
+    }</strong> legs and <strong>${this.hands}</strong> hands. 
+     My friends: <strong>${this.listFriends()}</strong> and my phrase: <strong>${this.sayPhrase()}</strong>`;
+  }
```
OleksiyRudenko, line 41: 1. Too much code repetition, and there is a pattern for each property. Use list of props and `.map` to wrap each prop value in `<strong>`.

2. If Habitant has no hands then `toString` method shouldn't handle it at all. You have overloaded `toString` in the child classes for this.

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+  sayPhrase() {
+    return super.sayPhrase();
+  }
+  addFriends(newFriend) {
+    super.addFriends(newFriend);
+  }
+  listFriends() {
+    return super.listFriends();
+  }
```
OleksiyRudenko, line 59: You don't need to re-define any methods defined in a super class, they are inherited and automatically available to the child classes however deep hierarchy is.

Yet you can still redefine `toString` to add value unique to a child class, in this task.

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+print(dog);
+print(cat);
+print(man);
+print(woman);
```
OleksiyRudenko, line 125: Too much code repetition.
Imagine you have 20 inhabitants.

Hint `[obj1, obj2].forEach()`

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+  addFriends(newFriend) {
+    super.addFriends(newFriend);
+  }
+  listFriends() {
+    return super.listFriends();
+  }
+}
+class Cat extends Habitant {
+  constructor(name, gender, phrase, friends, legs = 4, hands = 0) {
```
OleksiyRudenko, line 62: If cats do not have hands by design why would we define any at all?

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+const dog = new Dog("Snoop", "male", "Woof!", ["Ban", "July"]);
+const cat = new Cat("Fluffy", "female", "Meow!", ["July"]);
+const man = new Human("Ban", "male", "Hello everybody.");
+const woman = new Human("July", "female", "Blah blah blah.", [
+  "Ban",
+  "Fluffy",
+  "Snoop",
+]);
```
OleksiyRudenko, line 102: You have a tool (`addFriends`) that is an enabler to add real inhabitants rather than their names as strings.
This is an extra task.
Will you accept this challenge now or when you revert to this shiny world at OOP Exercise?

OleksiyRudenko (changes requested):
@7SergSmith7 Great job!
Let's make it perfect

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

submissions/7SergSmith7/Tiny JS World/index.js
OleksiyRudenko: 👍  Code became shorter and concise )

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: done

OleksiyRudenko: Well, not exactly. You need to resolve (1) while keep (2).
Item 1 was a bit overdone. You had explicit props used, now you have no explicit list and base class manages all properties. What if a child class introduces a property it doesn't want to be a part of an output?

Hint: `['propA', 'propB'].map(propName => this[propName])`

OleksiyRudenko (changes requested):
@7SergSmith7 almost there!
But a bit overdone. We want to have the code DRY yet we don't want to deprive child classes of their natural right to manage their own properties, and base class shouldn't manage children's props neither explicitly, nor implicitly.

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: maybe done)

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,105 @@
...
+      ? this.friends.map((friend) => friend.name)
+      : "I am loner";
+  }
+
+  sayPhrase() {
+    return this.phrase;
+  }
+
+  toString() {
+    this.props = this.props.map((prop) => `<strong>${this[prop]}</strong>`);
```
OleksiyRudenko, line 40: (1) When we call `toString` for any object for the 2nd time this part will break. 
Why? At the second call `this.prop` will contain smth like `['<strong>cat</strong>',....]` so while being mapped will fail at `this[prop]` as it would return `undefined` (as there is no property named `'<strong>cat</strong>'`

(2) This fragment still implicitly manages properties from child classes as it traverses `.props` array populated by a child class (the latter is basically legal though).

Solution:
List properties' names explicitly.
Manage strongization of only properties known to that particular class. You may want to use a helper function like ``function makeStringStrong(s) { return `<strong>${s}....` } `` to be used across all classes for consistency.

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+    return `Hi! I am ${this.props[0]}, my name is ${
+      this.props[2]
+    }, my gender is ${this.props[3]}.
+    My phrase: ${
+      this.props[4]
+    } Friends: <strong>${this.listFriends()}</strong>. I have ${
+      this.props[1]
+    } legs `;
```
OleksiyRudenko, line 49: Referring to the props by indices looks cryptic. Imagine, the class has dozens of props. Easy to mess up with indices.

Deconstruction at our rescue
Hint: ``[var1, var2, var3] = ['var1', 'var2', 'var3'].map(varName => `<strong>${this[varName]}</strong>`) ``

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,105 @@
...
+    super("human", name, gender, phrase, friends, legs, hands, props);
+    this.hands = hands;
+    // this.props.push("hands");
+  }
+  addProps() {
+    super.addProps();
+    this.props.push("hands");
+  }
+  toString() {
+    return super.toString() + `and  have ${this.props[5]} hands`;
```
OleksiyRudenko, line 74: `props[5]` will definitely break when we change hierarchy, add a property e.g. `ears`.
Imagine you have dozens of classes and inherited properties.

OleksiyRudenko (changes requested):
@7SergSmith7 good job. But the problem is only half-solved.
Check the comments below.
Let's make code more straightforward.
I suppose we will also get rid of `addProps` definitions and calls.

Do not hesitate asking questions when got any.

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: Please look at this option

submissions/7SergSmith7/Tiny JS World/index.js
```diff
@@ -0,0 +1,115 @@
...
+    [
+      this.speciesProp,
+      this.nameProp,
+      this.genderProp,
+      this.phraseProp,
+      this.legsProp,
+      this.handsProp,
+    ] = ["species", "name", "gender", "phrase", "legs", "hands"].map(
+      (prop) => `<strong>${this[prop]}</strong>`
+    );
```
OleksiyRudenko, line 80: As `Human` extends `Habitant` the `Habitant`'s constructor gets called (actually thanks to the `super` call on the top of `Human`'s constructor. No need to construct `this.speciesProp` twice.

Besides this (and similar in the super class) fragment adds extra properties that are in fact only needed in `toString`.

Besides these `*Prop` props are carved in stone at construction time? What if we change any core prop like e.g. `this.hands` somewhere in the course of inhabitant's life and try to represent them? We will still have the "old" property value as output.

Suggested approach:
1. Create formatted values right where they are needed (i.e. in `toString`)
2. Do not store them as props but rather as temporary values every time the `toString` gets called.

OleksiyRudenko (changes requested):
@7SergSmith7 almost done.
There is a flaw that might fire at us back when we try to expand the world and make inhabitants change.

submissions/7SergSmith7/Tiny JS World/index.js
7SergSmith7: Done =)

OleksiyRudenko (approved):
@7SergSmith7 great job done!

PR #124, opened 2020-11-14
tiny js with classes

Nemooochka:
### Tiny js with classes

[Demo](https://nemooochka.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Nemooochka/a-tiny-JS-world)

Please, review.


submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,121 @@
...
+
+class Cat extends Animal {
+    species = 'cat';
+    saying = 'Meeeow!';
+    constructor(name, gender, friends) {
+        super(name, gender, friends);
+    }
+}
+
+class Human extends Inhabitant {
```
OleksiyRudenko, line 58: Seems like humans do not have any legs. Doesn't look realistic enough.

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,121 @@
...
+    sayHi() {
+        return(
+            super.sayHi() + ` By the way, I have ${this.hands} hands.`
+        )
+    }
+}
+
+class Man extends Human {
+    gender = 'male';
+    saying = 'How you doing?';
```
OleksiyRudenko, line 76: So, all men will have the same saying. In real life humans are way more customizable compared with animals.
Please allow humans to have individual greetings.

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,121 @@
...
+    constructor(species, name, gender, saying, friends) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.saying = saying;
+        this.friends = friends;
+    }
+
+    sayHi() {
+        return( `${this.saying} I'm a ${this.species} with the name ${this.name}, which means I'm ${this.gender}. I want to say hi to <b>${this.friends}</b>, you are the best!`);
```
OleksiyRudenko, line 21: Well, hardly having a specific name necessarily defines a gender.
"My name is Alex, which means I'm... errm... well..."

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,121 @@
...
+    constructor(name, friends) {
+        super(name, friends);
+    }
+}
+
+const dog = new Dog('Ivan', 'male',  ['Ira']);
+const cat = new Cat('Pushok', 'male',  ['Sharik', 'Ponchik']);
+const catWoman = new Cat('Murka', 'female',  ['Pushok']);
+const man = new Man('Andre', ['Vasya', 'Mitya']);
+const woman = new Woman('Natali', ['Madona']);
```
OleksiyRudenko, line 96: Looks like there are more inhabitants in this world, Ponchik, Vasya, Madona to name a few.
They are rather virtual though.
Can you make it so that friends list comprises real inhabitants?

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+class Animal extends Inhabitant {
+    legs = 4;
+    constructor(name, gender, friends) {
+        super();
+        this.name = name;
+        this.gender = gender;
+        this.friends = friends;
```
OleksiyRudenko, line 31: The fragment above establishes properties that actually should be initialized via super as own properties of class `Animal`. This makes superclass pointless.

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,121 @@
...
+    sayHi() {
+        return(
+            super.sayHi() + ` By the way, I have ${this.legs} legs.`
+        )
+    }
+}
+
+
+class Dog extends Animal {
+    species = 'dog';
```
OleksiyRudenko, line 43: `species` here is re-introduced as own `Dog`'s class property while it should be passed over all the way up to `Inhabitant` constructor.
`super()` calls immediate parent constructor (i.e. its equivalent to `Animal.constructor()` in this class). Take advantage of this fact.

OleksiyRudenko (changes requested):
@Nemooochka you did very well!
You created a quite consistent hierarchy of classes, where superclasses (parent classes) define shared properties while child classes introduce properties specific to each class.
You employed `super.methodName` to overload class methods. Very well.

However, there is still some room for improvement. Let's make your implementation great.

Nemooochka:
Hi @OleksiyRudenko ,

Thank you for the detailed review! 
Pls, re-check. I hope, I get right the idea of using superclass.

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+class Inhabitant {
+    constructor(species, name, gender, legs, saying, friends = []) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.legs = legs;
+        this.saying = saying;
+        this.friends = friends;
+        this.addFriends(friends);
```
OleksiyRudenko, line 19: Here we got direct assignment of `.friends` with the value from constructor params, then again assignment to `.friends` by calling `.addFriends`.
Looks a bit redundant. A vote for assignment via the method as it is created for purpose and encapsulates all required logic to properly handle the received argument.

But if we just remove direct `.friends` assignment then the logic inside `.addFriends` may break.

Two options with this problem:
(1) either assign `.friends` with empty array in constructor
 or
(2) handle creation of `.friends` in `.addFriends` if the property doesn't exist yet

It is up to you to decide what approach is cleaner, SOLID, DRY etc

Yet your

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    addFriends(friends) {
+        if(Array.isArray(friends)) this.friends = [...this.friends, ...friends];
+        else this.friends.push(friends);
+    }
+
+    getFriendsName() {
```
OleksiyRudenko, line 27: This method returns 1+ names. So the plural form (`Names`) in the method's name will better reflect the method implementation.

submissions/Nemooochka/tiny-js-upd/index.js
```diff
@@ -0,0 +1,126 @@
...
+womanNatali.friends = [cat, catWoman];
+
+// Andre met Natali, they are friends now!
+manAndre.addFriends(womanNatali);
+womanNatali.addFriends(manAndre);
+
+
+const inhabitants = [dog, cat, catWoman, manAndre, manVasya, manMitya, womanNatali];
+
+inhabitants.forEach(inhabitant => print(inhabitant.sayHi()));
```
OleksiyRudenko, line 105: Bonus learning:

Each object in JS is inherited from the root `Object`.
The root `Object` has a magic method `toString` which is called implicitly if an object is referred in string context. It is the `toString` method that is responsible for outputs like `[Object object]` you might come across at times.

So, if you rename `sayHi` to `toString` you 
(a) will overload the default `.toString` implementation
and
(b) you will benefit from the implicit call of `.toString`, which will make your code a bit shorter and thus cleaner (`print` ensures that an object is referred in string context)


To see the immediate effect do:
(1) add `` console.log(`Object native string representation: ${inhabitant}`) ` in a proper place
(2) run you code and take a note of the output
(3) rename `sayHi`
(4) run your code and see the difference

So, once you rename the method, `print(inhabitant)` turns sufficient.

OleksiyRudenko (changes requested):
@Nemooochka well done!
Let's polish it.

submissions/Nemooochka/tiny-js-upd/index.js
Nemooochka: Cool idea, thank you!

Nemooochka:
@OleksiyRudenko , upd, please re-check :)

OleksiyRudenko (approved):
@Nemooochka well done!

PR #125, opened 2020-11-14
Saiko tiny js world

RomanSaiko:
### Building a Tiny JS World (pre-OOP) - practice

[Demo](https://romansaiko.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/RomanSaiko/a-tiny-JS-world/)

Please, review.


submissions/RomanSaiko/tiny-js-world/index.js
```diff
@@ -0,0 +1,57 @@
...
+    friends: ['Dolf', 'Spike']
+}
+
+const inhabitants = [dog, cat, man, woman, catWoman]
+
+console.log(inhabitants);
+
+for(let i = 0; i < inhabitants.length; i++) {
+    print(`${inhabitants[i].species}; ${inhabitants[i].name}; ${inhabitants[i].gender}; ${inhabitants[i].legs}; ${inhabitants[i].hands}; ${inhabitants[i].saying}; ${inhabitants[i].friends.map(item => item)}.`);
+}
```
OleksiyRudenko, line 57: Please employ an appropriate `Array` method to iterate through list items.

Please encapsulate the output string creation in a function. In that function you might also list object properties and an `Array` method to convert them into values. Of course `freinds` property would still require specific handling.

OleksiyRudenko (changes requested):
@RomanSaiko well done!
Let's improve code conciseness and readability

submissions/RomanSaiko/tiny-js-world/index.js
OleksiyRudenko: Too much code repetition. Keep code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) as much as possible. This is a tiny task and perhaps this effort is a bit tto much for it, but you will find the approach paying on bigger tasks and at work.

What if you are required to add another 10 properties? With current approach you will need to add each twice, in decomposition and in template string literal. With a DRY approach you have a single list of property names.

Hint:
`['propA', 'propB', 'propC'].map(propName => obj[propName])` results in an array of object property values.

Repetetive `; ` can be replaced with a proper `.join()`

Yet `friends` will require special handling, true.

Combination of `.map`'s, `.join`'s, `.forEach`'s, sometimes nested will do the magic.

OleksiyRudenko (changes requested):
@RomanSaiko nice try! Nearly there
Let's make code DRYer.

OleksiyRudenko (approved):
@RomanSaiko great job!
FYI, order of properties (keys) in an object is not guaranteed. So whenever you rely on specific order of keys it is better to list required keys exolicitly.
E.g. `['propA', 'propB', 'propC'].map(propName => obj[propName])`

This way you also process only those props you want instead of all props.

PR #127, opened 2020-11-15
Frogger

AnnaStoyano:
### Frogger

[Demo](https://annastoyano.github.io/Frogger/index) |
[Code base](https://github.com/AnnaStoyano/Frogger)

Please, review.


OleksiyRudenko:
@AnnaStoyano looks nice! :)

## Check-list

Minimal requirements to meet:
 * Requirements re **OOP**:

submissions/AnnaStoyano/frogger/app.js
```diff
@@ -0,0 +1,134 @@
...
+Enemy.prototype = Object.create(Entity.prototype);
+Enemy.prototype.update = function (dt) {
+    this.position.x += this.speed * dt * (this.level * 0.8);
+    this.checkWallCollision(canvas.width, canvas.height, -90, 0);
+    this.checkLose();
+}
+
+Enemy.prototype.checkCollision = function () {
+    let positionX = Math.floor(this.position.x);
+    if (positionX - 50 < player.position.x && player.position.x < positionX + 50) {
```
OleksiyRudenko, line 48: Here `player` is a global variable.

From the checklist:
> classes do not refer to any global variables, like global variable player, which is an instance of Player class
(referring to global constants and globals provided by the gaming platform like Resources is OK);

Double check your entire code base to meet the requirement.

submissions/AnnaStoyano/frogger/app.js
```diff
@@ -0,0 +1,134 @@
...
+            this.position.y += 20;
+            break;
+    }
+}
+
+const player = new Player(200, 404, 7, 'images/char-boy.png');
+
+const enemy1 = new Enemy(30, 120, 60, 'images/enemy-bug.png', player);
+const enemy2 = new Enemy(15, 220, 30, 'images/enemy-bug.png', player);
+const enemy3 = new Enemy(0, 50, 20, 'images/enemy-bug.png', player);
```
OleksiyRudenko, line 120: 200, 404, 7, 220 etc look look magic

From the requirements:
> all numbers like block dimensions, initial locations are defined as constants

OleksiyRudenko (changes requested):
@AnnaStoyano thank you for the update.

The below comments comprise a partial review.

Please check your code against the checklist and:
> Please, confirm all items above are positive. Please, ask a question if anything above requires clarification.

AnnaStoyano:
@OleksiyRudenko  Thank you for your comments. I hope that I corrected my mistakes.

submissions/AnnaStoyano/frogger/app.js
```diff
@@ -0,0 +1,146 @@
...
+const ENEMY_1_SPEED = 20;
+const ENEMY_2_SPEED = 60;
+const ENEMY_3_SPEED = 30;
+
+
+const player = new Player(START_POSITION_PLAYER.X, START_POSITION_PLAYER.Y, PLAYER_SPEED, 'images/char-boy.png');
+const enemy1 = new Enemy(START_POSITION_ENEMY_1.X, START_POSITION_ENEMY_1.Y, ENEMY_1_SPEED, 'images/enemy-bug.png', player);
+const enemy2 = new Enemy(START_POSITION_ENEMY_2.X, START_POSITION_ENEMY_2.Y, ENEMY_2_SPEED, 'images/enemy-bug.png', player);
+const enemy3 = new Enemy(START_POSITION_ENEMY_3.X, START_POSITION_ENEMY_3.Y, ENEMY_3_SPEED, 'images/enemy-bug.png', player);
+const allEnemies = [enemy1, enemy2, enemy3];
```
OleksiyRudenko, line 133: Whenever you feel adding an index in a variable name (like `ENEMY_1_SPEED`, `START_POSITION_ENEMY_3` or `enemy2`) this is clear sign that array would be more appropriate.
Imagine you'd be required to add values for dozens of characters.
There is also massive code repetition here (again, think of dozens of characters).
Can you devise an approach that would employ data structure (array of objects containing individual enemy settings)?
You might then efficiently use e.g. `Array.map`.

Hint:
`allEnemies = enemySettings.map(settings => new Enemy(settings.startPosition.X, settings.startPosition.Y, settings.speed, ENEMY_IMAGE, player))`

OleksiyRudenko (changes requested):
@AnnaStoyano well done. Yet you will benefit more if you master data structures to save time and lines of code.

AnnaStoyano:
@OleksiyRudenko  Thank you for your advice and hint)

OleksiyRudenko (approved):
@AnnaStoyano well done!
There are still numbers with unclear semantic purpose worth being assigned in constant definitions.
At all times keep thinking of how your code **reads**. The closer to plain English the better.

PR #128, opened 2020-11-15
Oop exercise

xT0PERx:
### OOP exercise

[Demo](https://xt0perx.github.io/OOP-exercise/) |
[Code base](https://github.com/xT0PERx/OOP-exercise)

Please, review.


OleksiyRudenko (changes requested):
Well, this is a nice transition from object literals to class syntax.
However, OOP is not just about syntax but also a set of paradigms.
A good hierarchy of classes sllows you to create superclasses that initialize and handle shared properties, while child classes introduce and handle properties unique to them.
Consider revisiting the original repo [guidelines and materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) and then refactor the implementation so it becomes a better model of a real world.

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,93 @@
...
+  printInfo() {
+    print(
+      [
+        this.species,
+        this.name,
+        this.gender,
+        this.hands,
+        this.legs,
+        this.say(),
+      ].join('; ')
```
OleksiyRudenko, line 18: Class Mammals doesn't have properties `hands`, `gender` and some other.
Super class shouldn't know anything about child classes internals.
Make sure the props specific to child classes are nadled by relevant child classes.
You may want to benefit from method overloading.

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,93 @@
...
+    );
+  }
+}
+
+class Animal extends Mammals {
+  constructor({ name, gender, legs = 4, saying }) {
+    super({ name, saying });
+    this.species = 'animal';
+    this.gender = gender;
```
OleksiyRudenko, line 27: All child classes have `species` and `gender` props. This is a clear sign that it should be initialized and handled by a super class.
Double check other classes and other props.

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,93 @@
+class Mammals {
```
OleksiyRudenko, line 1: Plural form in class id normally denotes a collection, while here this class instantiates a single instance of an object.

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,93 @@
...
+  constructor(props) {
+    this.name = props.name;
+    this.saying = props.saying;
+  }
+  say() {
+    return this.saying;
+  }
+  printInfo() {
+    print(
```
OleksiyRudenko, line 10: Way too specific use case. What if we want a different method of output? Wouldn't be great to add yet another method to handle just it. Class shouldn't be aware of interaction interfaces.
Well, this method might receive a callback to apply as an output method. The simpler approach is to have a method that simply returns a string and inquiring party does with that string whatever it likes.

OleksiyRudenko (changes requested):
@xT0PERx well done.
However there is still room for improvements.
Let's make the code more [SOLID](https://en.wikipedia.org/wiki/SOLID) and [OOP principles](https://medium.com/@cancerian0684/what-are-four-basic-principles-of-object-oriented-programming-645af8b43727) compliant.

xT0PERx:
Thanks for the good advice - it helped to understand the topic deeper, I hope I understood all the comments correctly and corrected

submissions/xT0PERx/OOP exercise/index.js
OleksiyRudenko: @xT0PERx cannot identify changes addressing the issue above. Please fix or ask a question

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,84 @@
+class Mammal {
+  constructor(name, gender, saying) {
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+  }
+  toString() {
+    return `${this.name}; ${this.gender}; ${this.saying}`;
```
OleksiyRudenko, line 8: There is a pattern. Can you optimize this fragment employing `Array` methods?
Here and everywhere across all classes.

OleksiyRudenko (changes requested):
@xT0PERx well done.
However some fragments aren't DRY enough.
There is also an issue from the past review that wasn't addressed. Please check it out

submissions/xT0PERx/OOP exercise/index.js
xT0PERx: Unfortunately, the possibility of simplifying the code using array methods does not occur to me. Perhaps you could somehow lead me to this idea? I looked at other works and found this approach.
return [this.species, this.name, this.gender, this.age] .join (',');
But I cannot understand why he is better? Maybe that's not what you mean.

submissions/xT0PERx/OOP exercise/index.js
xT0PERx: There is a question. As I thought when creating the superclass and classes:
1. I have a basic constructor - mammals. He creates a mammal - but he does not yet know whether it will be a human or an animal (I decided to divide mammals this way).
2. Then I create classes of people and animals from my superclass (here I add the species)
3. The following are already specified classes such as cat, dog, etc.
What's wrong with my logic? Why should I add species to the Mammals?
Perhaps during the day I myself will reach what you meant, it also happens. But for now, I would like to hear your thought. Thank you.

submissions/xT0PERx/OOP exercise/index.js
OleksiyRudenko: It is not about "better".
This is quite a simple use case. If you can spot and resolve the pattern in a simpler case then the chances you will spot in more complicated cases are higher. If you don't learn it now you will struggle later.
This may help you to avoid big blocks of template literals in e.g. Friends app and other, and also t oavoid repetitions of bigger blocks of code.
Do you want to master this skill?

xT0PERx: Of course I want, my task is to change my occupation. And while there is an opportunity to learn from me and from you to teach - I would like to use it.

OleksiyRudenko: So what are your own on thoughts on the difference between using a template literal and `Array` methods?
Imagine you need to add a dozen properties more to print or to reorder the properties for output.

xT0PERx: To change the order of properties for display and to add a large number of elements, of course, it is better to use an array and its methods.
#### So I should add an array with elements and then reuse them? The way I see it:

> toString() {
    return [this.name, this.gender, this.saying].join(';');
  }

#### Then the Animal class will already have:

> toString() {
    return super.toString()+';'+[this.species, this.legs].join(';');
  }

That is, in the future, if I need to add any more features, I just add them to the array?
#### Maybe I misunderstood something?

OleksiyRudenko: @xT0PERx 

The 2nd fragment also has a pattern thus optimizable.
```
toString() {
  return super.toString()+';'+[this.species, this.legs].join(';');
}
```

```
toString() {
  return [super.toString(), this.species, this.legs].join(';');
}
```

OleksiyRudenko: Yet another perspective on the problem to reflect on

In
```
toString() {
  return super.toString()+';'+[this.species, this.legs].join(';');
}
```
if replace `super.toString()` speculatevely then the entire fragment look like the following
```
[this.name, this.gender, this.saying].join(';') + ';' + [this.species, this.legs].join(';')
```
Can you see the recurrent nature of the problem?

xT0PERx: With each task, I think that I am no longer a completely newbie, but with each review I am convinced of my inexperience. Thanks for your detailed comments.

OleksiyRudenko: Learning is the endless process. Stretching and challenging is the way to learn.
Mentors challenge students one or two steps at a time to create stretching thus keeping the momentum for learning.

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,88 @@
...
+      this.friend,
+    ].join(";");
+  }
+}
+
+class Cat extends Animal {
+  constructor(name, gender, saying, friend) {
+    super(name, gender, saying);
+    this.species = "cat";
+    this.friend = friend;
```
OleksiyRudenko, line 46: Why `species` and `friend` are handled in a way different from handling `name`, `gender`, and `saying`?
When you answer this question, you will find to optimize your code across entire code base.

OleksiyRudenko (changes requested):
@xT0PERx now the code is easier and faster to read and comprehend.
There are some incosistencies in a way how properties are stored and handled. Let's fix it.

submissions/xT0PERx/OOP exercise/index.js
xT0PERx: Probably they are no different :) And I must understand that the superclass should do the things that all classes have, and the classes, respectively, differ from the superclass in that they do something new, unlike it. And since I have in all  classes property :_species_, _friend_, _name_,_gender_,_saying_ that means I need to implement them through the superclass. Am I thinking in the right direction?

xT0PERx:
I tried to optimize the code, please check.

OleksiyRudenko (changes requested):
@xT0PERx nearly done

Why `legs` property is handled in a way different from handling `name`, `gender`, `friend`, `saying` and `species`?
When you answer this question, you will find to optimize your code across entire code base.
You may also find that `Cat` and `Dog` do not need overloading `toString` at all.

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,82 @@
...
+    super(name, gender, friend, saying, species, legs);
+  }
+  toString() {
+    return super.toString();
+  }
+}
+class Dog extends Mammal {
+  constructor(name, gender, saying, species = "dog", friend, legs = 4) {
+    super(name, gender, friend, saying, species);
+    this.legs = legs;
```
OleksiyRudenko, line 51: Is there a good reason for `Dog` to handle legs in a way different from `Cat`?

submissions/xT0PERx/OOP exercise/index.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+  toString() {
+    return super.toString();
+  }
```
OleksiyRudenko, line 55: Why do we need to overload this method?

OleksiyRudenko (changes requested):
@xT0PERx improvements are seen.
Please do not expect to have every single mistake, similar to that once olready addressed, clearly marked with red ink.

submissions/xT0PERx/OOP exercise/index.js
xT0PERx: Again my inattention :(

submissions/xT0PERx/OOP exercise/index.js
xT0PERx: You are right - there is no need.

OleksiyRudenko: @xT0PERx
Why do we need to call `toString()` in constructors of `Dog` and `Cat`?
Why other classes' constructors do not call `toString()`?
Is this function defined at all?

Please do not consider the questions above rhetoric. I want to understand the decision making process behind.

xT0PERx: I call this method in the _"Dog"_ and _"Cat"_ classes in order to display their properties.
Constructors of other classes call the "toString()" method, but with additional properties that the superclass does not have.
>toString() {
    return [super.toString(), this.hands].join(';');
  }

The function toString() is defined in the superclass Mammals.

xT0PERx: It turns out like. I believe that overriding the "toString()" method from the Object methods in the superclass Mammals - helps me to use
>print (inhabitant));

Because otherwise I will be displayed on the screen __[Object, Object]__.
It just happened that I missed the point that I no longer need to call the method __"toString()"__ in __Dog__ and __Cat__, because he already does everything I need in the Mammals superclass

submissions/xT0PERx/OOP exercise/index.js
OleksiyRudenko: Please take learning on learning process. This type of excuse is OK while you're learning but will be bad excuse in professional work.

submissions/xT0PERx/OOP exercise/index.js
OleksiyRudenko: @xT0PERx please revisit OOP principles. You need to understand inheritance better.

OleksiyRudenko (approved):
@xT0PERx well done!

PR #130, opened 2020-11-16
Frogger game

nadsatt:
### Frogger game

[Demo](https://nadsatt.github.io/kottans-homeworks/frogger-game) |
[Code base](https://github.com/nadsatt/frontend-2021-homeworks/tree/frogger-game/submissions/nadsatt/frogger-game)

Please, review.


OleksiyRudenko:
@nadsatt please remove files that you didn't change as these aren't required to conduct the review and it will speed the review up.
It is expected that you only `app.js` is affected.

nadsatt:
@OleksiyRudenko i removed all extra files except `engine.js`. I left `engine.js` because i changed code in this file. In my case `app.js` describes entities while `engine.js` creates and uses entities.

OleksiyRudenko:
@nadsatt well, not really appropriate approach.
Please keep engine intact.

Original [requirements](https://docs.google.com/document/d/1v01aScPjSWCCWQLIpFqvg3-vXLH2e8_SZQKC8jNO0Dc/pub) from Udacity task say _"The js folder also contains the app engine needed to run the game and a resources.js file. You do not need to edit these files."_

Please keep your entire custom codebase in `app.js`

nadsatt:
@OleksiyRudenko 
I moved all logic in single `app.js` file. Please, review.

OleksiyRudenko (changes requested):
Errm... well...
We do not really need or want original code from `engine.js` here.

Instead we want to keep engine code as it is and any your own code required to fulfill the task in `app.js`. The code in `app.js` should just use engine API.

Why do you want or need to change the engine code? (Please answer this question)

You may also find it helpful to see how other students approached the use of engine in their submissions.

nadsatt:
@OleksiyRudenko, i guess some misunderstanding occurred because of wrong formulation of what was done. 

What **actually** was done:
1) I replaced my `engine.js` with the original one (from udacity assests).
2) I rewrited my `app.js` to make it work with untouched `engine.js`

OleksiyRudenko:
Oh, I see now. I was confused with `Resources.load`. Thank you for the explanation. This type of comments supplements code submissions well and makes a reviewer's life eaiser.

OleksiyRudenko (approved):
@nadsatt great job!

I figure, some values are calculable from even more elementary constants. E.g. there is a tile visible height and width, other things like starting coordinates and step delta are a varity of multiplication by factors representing rows or column of the tile grid, perhaps with an offset at some cases.

If we had that base constant and set of multipliers and offsets that would make switching to a different tile/character size a breeze.

This however is beyond this task.

JFYI, the entity named `.field` in this implementation has a specific name, bounding box.

PR #133, opened 2020-11-16
OOP

AnnaStoyano:
### OOP Exercise

[Demo](https://annastoyano.github.io/a-tiny-JS-world/index) |
[Code base](https://github.com/AnnaStoyano/a-tiny-JS-world/)

Please, review.

### DIFF
--- a/submissions/AnnaStoyano/oop-exercise/index.js
+++ b/submissions/AnnaStoyano/oop-exercise/index.js
@@ -1,0 +1,109 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Mammals{
+   constructor(option){
+      this.species = 'mammals',
+      this.name = option.name,
+      this.gender = option.gender,
+      this.legs = 4,
+      this.hands = 0,
+      this.saying = option.saying,
+      this.friends = option.friends
+   }
+}
+
+class Dog extends Mammals{
+   constructor(option){
+      super(option);
+      this.species = 'dog',
+      this.saying = option.saying || 'guf'
+   }
+};
+
+class Cat extends Mammals{
+   constructor(option){
+      super(option);
+      this.species = 'cat',
+      this.saying = option.saying || 'meow'
+   }
+
+   getSaying(){
+      return this.saying || 'meow';
+   }
+}
+
+class Human extends Mammals{
+   constructor(option){
+      super(option);
+      this.species = 'human'
+      this.hands = 2;
+      this.legs =2;
+      this.saying = option.saying || `Hello! My name is ${this.name}`
+   }
+}
+
+class Woman extends Human{
+   constructor(option){
+      super(option);
+      this.gender = 'female'
+   }
+}
+
+class Man extends Human{
+   constructor(option){
+      super(option);
+      this.gender = 'male'
+   }
+}
+
+class CatWoman extends Woman{
+   constructor(option){
+      super(option);
+      this.saying = Cat.prototype.getSaying();
+   }
+}
+
+const cat = new Cat({name:'Murzic',gender:'male',friends:['Elza','Nicky']})
+const dog = new Dog({name:'Lucy',gender:'female',saying:'guffyy',friends:['Alla','Dima']});
+const man = new Man({name:'Vlad',friends:['Sergay','Anton'],saying:'Hello everyone!'});
+const woman = new Woman({name:'Anna',friends:['Sergay','Artem']});
+const catWoman = new CatWoman({name:'Lily',friends:['Batman']})
+
+
+const inhabitants = [man,woman,cat,dog,catWoman];
+
+const properties = ['name','species','gender','legs','hands','saying','friends'];
+let output = inhabitants.map(inhabitant=>
+   properties.map(property=>`<b>${property}:</b> ${inhabitant[property]}`
+   ).join('\t')
+).join('\n');
+
+print(output);
+// ======== OUTPUT ========
+/* Use print(message) for output.
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
+
--- a/submissions/AnnaStoyano/oop-exercise/index.js
+++ b/submissions/AnnaStoyano/oop-exercise/index.js
@@ -9,30 +9,28 @@
 // Define your objects here
 
 class Mammals{
-   constructor(option){
+   constructor(name,gender,friends){
       this.species = 'mammals',
-      this.name = option.name,
-      this.gender = option.gender,
+      this.name = name,
+      this.gender = gender,
       this.legs = 4,
-      this.hands = 0,
-      this.saying = option.saying,
-      this.friends = option.friends
+      this.friends = friends
    }
 }
 
 class Dog extends Mammals{
-   constructor(option){
-      super(option);
+   constructor(name,gender,friends,saying){
+      super(name,gender,friends);
       this.species = 'dog',
-      this.saying = option.saying || 'guf'
+      this.saying = saying || 'guf'
    }
 };
 
 class Cat extends Mammals{
-   constructor(option){
-      super(option);
+   constructor(name,gender,friends,saying){
+      super(name,gender,saying,friends);
       this.species = 'cat',
-      this.saying = option.saying || 'meow'
+      this.saying = saying || 'guf'
    }
 
    getSaying(){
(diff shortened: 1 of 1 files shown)


submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+      this.legs = 4,
+      this.friends = friends
+   }
+}
+
+class Dog extends Mammals{
+   constructor(name,gender,friends,saying){
+      super(name,gender,friends);
+      this.species = 'dog',
+      this.saying = saying || 'guf'
```
OleksiyRudenko, line 25: Every child class has property `saying`. This is a clear sign that it is a shared property and as such should be initialized in and handled by parent class.

Default value for saying can be defined at function definition. For exmaple `function abc(a, b, c = 5) { ... }`. Class method is basically a function. `constructor` is not an exclusion.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+      this.gender = gender,
+      this.legs = 4,
+      this.friends = friends
+   }
+}
+
+class Dog extends Mammals{
+   constructor(name,gender,friends,saying){
+      super(name,gender,friends);
+      this.species = 'dog',
```
OleksiyRudenko, line 24: Every child class has property saying. This is a clear sign that it is a shared property and as such should be initialized in and handled by parent class.
Even if class `Dog` doesn't receive species via its constructor there are no obastcles to pass a string literal to to its parent via `super()`

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+const dog = new Dog(name='Lucy',gender='female',saying='guffyy',friends=['Alla','Dima']);
+const man = new Man(name='Vlad',friends=['Sergay','Anton'],saying='Hello everyone!');
+const woman = new Woman(name='Anna',friends=['Sergay','Artem']);
+const catWoman = new CatWoman(name='Lily',friends=['Batman'])
+
+const inhabitants = [man,woman,cat,dog,catWoman];
+
+const properties = ['name','species','gender','legs','hands','saying','friends'];
+let output = inhabitants.map(inhabitant=>
+   properties.map(property=>{
```
OleksiyRudenko, line 82: Let's make construction of a self-presentation string a part of classes.
This way you will avoid `.hasOwnProperty` check as every class has very few properties so it handle those properly.

You will also benefit from OOP concepts of inheritance and methods overloading.

Hints:
Each class should handle only properties know to it.
`super.method()` calls a `method` from its super class.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+   getSaying(){
+      return this.saying || 'meow';
+   }
+}
+
+class Human extends Mammals{
+   constructor(name,gender,friends,saying){
+      super(name,gender,friends);
+      this.species = 'human'
```
OleksiyRudenko, line 44: Having no semicolon is generally valid in JS. Yet we recommend to keep code style consitent.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+      this.species = 'dog',
+      this.saying = saying || 'guf'
+   }
+};
+
+class Cat extends Mammals{
+   constructor(name,gender,friends,saying){
+      super(name,gender,saying,friends);
+      this.species = 'cat',
+      this.saying = saying || 'guf'
```
OleksiyRudenko, line 33: Comma-separated statements and having no semicolon are generally valid in JS. Yet we recommend to keep code style consitent.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+      super(name,friends,saying);
+      this.saying = Cat.prototype.getSaying();
+   }
+}
+
+const cat = new Cat(name='Murzic',gender='male',friends=['Elza','Nicky'])
+const dog = new Dog(name='Lucy',gender='female',saying='guffyy',friends=['Alla','Dima']);
+const man = new Man(name='Vlad',friends=['Sergay','Anton'],saying='Hello everyone!');
+const woman = new Woman(name='Anna',friends=['Sergay','Artem']);
+const catWoman = new CatWoman(name='Lily',friends=['Batman'])
```
OleksiyRudenko, line 76: Having no semicolon is generally valid in JS. Yet we recommend to keep code style consitent.

Please take every code review comment and check your entire code against the comment.

Code is written for other people. Please take care of them. Consistent code style is a form of caring.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+// Define your objects here
+
+class Mammals{
+   constructor(name,gender,friends){
+      this.species = 'mammals',
+      this.name = name,
+      this.gender = gender,
+      this.legs = 4,
+      this.friends = friends
+   }
```
OleksiyRudenko, line 18: Presumably, all mammals in our world have legs. But definitely not all have 4. It's unlikely class `Mammal` can have any defaults for this.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,112 @@
...
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Mammals{
```
OleksiyRudenko, line 11: Plural form is normaly used to identify collections. In this case this class creates a single instance of a world inhabitant.

OleksiyRudenko (changes requested):
@AnnaStoyano now it gonna be easier to focus on the class hierarchy and architecture. Thank you.
Let's complete the next iteration of improvements.

AnnaStoyano:
Thank you very much for your advices!

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,106 @@
...
+// Define your objects here
+
+class Mammal{
+   constructor(species,name,gender,legs,friends,saying){
+      this.species = species,
+      this.name = name,
+      this.gender = gender,
+      this.legs = legs,
+      this.friends = friends,
+      this.saying = saying
```
OleksiyRudenko, line 18: Incosistent code style (statements termination). Please fix here and double check everywhere across the code base.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Mammal{
+   constructor(species,name,gender,legs,friends,saying){
```
OleksiyRudenko, line 12: Please install and tune up [prettier](https://prettier.io/) plugin for your IDE. It will help you (even is able to fix minor issues for you) to keep code formatting as per code style conventions (e.g. spaces between arguments in the parameters or arguments list, or before opening figure bracket `{` after class name or method/function params list, or after `if`). Tiny things but they pop out.

Just ask other students in chat how do they make their code well formatted and neat.

This is not critical **now**, but will make the code easier to read and will pay off when you will collaborate with others (at work, for example).

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,106 @@
...
+const woman = new Woman('Anna',['Sergay','Artem']);
+const catWoman = new CatWoman('Lily',['Batman']);
+
+const inhabitants = [man,woman,cat,dog,catWoman];
+
+const properties = ['name','species','gender','legs','hands','saying','friends'];
+let output = inhabitants.map(inhabitant=>
+   properties.map(property=> inhabitant.getProperty(property))
+   .join('\t')
+).join('\n');
```
OleksiyRudenko, line 84: We need to separate concerns.

Please make self-presentation a method of a class so that in this exact place we do not need to know anything about classes' internals (this what OOP is importantly about) but instead we can do something like `dog.tellAbout()` or `inhabitants.map(inhabitant => inhabitant.tellAbout())......`

If done properly you will not have any output on hands for cats or dogs instead of `hands: cat doesn't have any`. Imagine, you have birds in your world, or we want to know something else about humans. How overwhelmed with unnecessary information the output would then be? Proper compliance with OOP principles helps to avoid this.

OleksiyRudenko (changes requested):
@AnnaStoyano well done! The code became shorter, cleaner, and easier to read.
There is yet room for important improvements though.
Let's take yet another iteration.

AnnaStoyano:
@OleksiyRudenko I am grateful for your advice. May you see my corrections again?

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,101 @@
...
+      this.friends = friends;
+      this.saying = saying;
+   }
+
+   tellAbout() {
+      const properties = ['species', 'name', 'gender']; //all mammals have this properties
+      for (let property in this) {
+         if (!properties.includes(property))
+            properties.push(property);
+      }
```
OleksiyRudenko, line 26: A class method should work only with properties defined in the class.
### Why that?
A developer may introduce a child class having properties the developer may not want to be printed.
Current design will force the developer to change code in the base class. Other developers on the project may not be happy.
### Solution
1. Make the method to work with properties known to this class.
2. In every child class overload the method and make it call `super.tellAbout()` and supplement the result with the values from properties know to that child class.

Check [this doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super) out.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,101 @@
...
+   }
+};
+
+class Cat extends Mammal {
+   constructor(name, gender, friends, saying) {
+      super('cat', name, gender, 4, friends, saying || 'meow');
+   }
+
+   getSaying() {
+      return this.saying || 'meow';
```
OleksiyRudenko, line 43: Now we have literal `'meow'` as a default value at two locations in the code.
This is an error prone approach. Someone may change it in one place and forget to change it in any other.
Perhaps we need a constant for this. But is there a chance that `this.saying` won't have definite value at this point?
It may happen that we do not need `|| 'meow'` here. Test it.

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
```
OleksiyRudenko, line 101: The mark below means that there is no newline character at the end of file.
image

You need to find a setting in your IDE that ensures every file ends with a newline character, or ensure having a newline  at every file manually.

[Reason why it is required and hints on how to fix it once forever](https://gist.github.com/OleksiyRudenko/d51388345ea55767b7672307fe35adf3)

OleksiyRudenko (changes requested):
@AnnaStoyano nearly done.
This might be the last iteration of code improvement.

AnnaStoyano:
@OleksiyRudenko  thank you very much for all advices) I hope that everything is good!

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+};
+
+class Cat extends Mammal {
+   constructor(name, gender, friends, saying) {
+      super('cat', name, gender, 4, friends, saying || 'meow');
+   }
+
+   tellAbout() {
+      let properties = ['species', 'name', 'gender', 'legs', 'friends', 'saying']; /* properties for display */
+      return super.tellAbout(properties);
```
OleksiyRudenko, line 45: Below are comments from the past reviews that this implementation doesn't take into account.

> Each class should handle only properties know to it.

> Make the method to work with properties known to this class.
> In every child class overload the method and make it call super.tellAbout() and supplement the result with the values from properties know to that child class.

In current implementation a child class refers to the properties that do not belong to a class.
Current implementation reads "Specify all properties, from current class and all its parent classes and call super class method to manage them" rather than "call super.tellAbout() and supplement the result with the values from properties know to that child class".

submissions/AnnaStoyano/oop-exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+   }
+}
+
+class CatWoman extends Woman {
+   constructor(name, friends) {
+      super(name, friends, 'meow', Cat.prototype.getSaying);
+   }
+   tellAbout() {
+      return super.tellAbout();
+   }
```
OleksiyRudenko, line 91: There is no point to just call an overloaded method from its parent class. All methods from parent classes are inherited by default.

OleksiyRudenko (changes requested):
@AnnaStoyano good job.
However, there is an important principle from OOP that should have been used in the implementation.
Please check the comment below.
You may want to refresh your knowledge of OOP (encapsulation, inheritance) and SOLID (open-closed, Liskov substitution) principles.

AnnaStoyano:
@OleksiyRudenko thank you for comments! I hope that everything is ok in current implementation.

OleksiyRudenko (changes requested):
@AnnaStoyano pretty well done with explicit properties handling.

Now, cats and dogs have a shared property (`tail`). All shared properties and their handling should be encapsulated in a common base class. However, base class for `Cat` and `Dog` is `Mammal`, and not all mammals in this world have manifesting tails. This means we should introduce a class between `Mammal` and `Cat` & `Dog`, e.g. `TailBearer`.
So the inheritance chains would look like `Mammal => TailBearer => Cat` and ``Mammal => TailBearer => Dog`.

**What we gonna do instead**

I'd propose that we just remove `tail` property(properties) and `tellAbout` overloading from Cat and Dog for simplicity.

Just make it carefully and we're done here.

**Learnings taken**

The learning from this is that hierarchy design requires effort and OOP+SOLID principles are worth mastering through practice. Imagine, you have to populate the world with birds, then with insects, fish etc. The better initial design the lesser effort system expansion requires.

I am pretty sure you've lernt something important from this task. OOP and SOLID principles aren't only about class syntax but rathe something about systems design and code organization for the sake of scalability and maintainability.

When done with this task take a moment to look at your initial code, latest version of it, reflelct on difference and put your thoughts down in your student's diary repo.

AnnaStoyano:
@OleksiyRudenko  thank you very much! Your advices and comments helped me understand OOP principles more clear.

OleksiyRudenko (approved):
@AnnaStoyano great job done!

PR #134, opened 2020-11-16
Object-Oriented JavaScript - practice

Iakow:
### Object-Oriented JavaScript - practice 

[Demo](https://iakow.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/Iakow/frontend-2021-homeworks/blob/oop-js-frogger/submissions/iakow/object-oriented-js/app.js)

Please, review.

### DIFF
--- a/submissions/iakow/object-oriented-js/app.js
+++ b/submissions/iakow/object-oriented-js/app.js
@@ -1,0 +1,139 @@
+var enemiesYPositions = {
+    'top': 60,
+    'middle': 145,
+    'bottom': 230
+}
+
+var looseSound = new Audio('sound/eralash-konec.mp3');
+var winSound = new Audio('sound/win.mp3');
+
+var Enemy = function (row, player) {
+    this.sprite = 'images/enemy-bug.png';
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@Iakow nice touch-up with sounds! :)

# Frogger Arcade Game -- JS OO exercise check list

A self-check list. Relates to
[Object-Oriented JavaScript](https://github.com/kottans/frontend/blob/master/tasks/js-oop.md) task.

## Check-list

Minimal requirements to meet:
 * Requirements re **OOP**:

Iakow:
Hello!

Iakow:
I think I brought the code in accordance with the checklist.
Fell down the rabbit hole with all these `this` and prototype chains ))
I've tried a ton of ways to write this code, and I think I've chosen the best option.
I'm not capable of more))

OleksiyRudenko (approved):
@Iakow good job!
There are still mysterious numbers across the code base with not really clear semantic purpose.
At all times keep thinking how the code **reads**. The closer to plain English the better.

PR #139, opened 2020-11-18
Object oriented js task

KikinovK:

### frogger game

[Demo](https://kikinovk.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/KikinovK/frontend-2021-homeworks/tree/object_oriented_js_task/submissions/kikinovk/frogger-game)

Please, review.

### DIFF
--- a/submissions/kikinovk/frogger-game/app.js
+++ b/submissions/kikinovk/frogger-game/app.js
@@ -1,103 +1,167 @@
-// Enemies our player must avoid
-var Enemy = function(x,y) {
-  // Variables applied to each of our instances go here,
-  // we've provided one for you to get started
+const ENEMY_START_POS_X = -100;
+const ENEMY_END_POS_X = 500;
+const ENEMY_BLOCK_HEIGHT = 85;
+const ENEMY_POS_INDENT = 60;
 
-  // The image/sprite for our enemies, this uses
-  // a helper we've provided to easily load images
-  this.sprite = 'images/enemy-bug.png';    
-  this.x = x;
-  this.y = y;
-  this.v = Math.random()*500;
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@KikinovK  nicely done!

## Check-list

Minimal requirements to meet:
 * Requirements re **OOP**:

KikinovK:
Fixed the code according to the checklist

OleksiyRudenko (approved):
@KikinovK well done!

A note:

Current approach with a quite detailed `ENEMY_CONF` in combination with `ENEMY_CONF.map(enemy => new Enemy(enemy.initPosition.x, enemy.initPosition.y, enemy.sprite))` is the right approach when we want to allow a developer customize every enemy's initial position and sprite.

If we wanted to customize only initial Y position this might be as short as
```
const ENEMIES_INITIAL_Y_POSITIONS = [
  ENEMY_POS_INDENT + ENEMY_BLOCK_HEIGHT * 0,
  ENEMY_POS_INDENT + ENEMY_BLOCK_HEIGHT * 0,
  ENEMY_POS_INDENT + ENEMY_BLOCK_HEIGHT * 1,
  ENEMY_POS_INDENT + ENEMY_BLOCK_HEIGHT * 1,
  // etc
];
// ....
const allEnemies = ENEMIES_INITIAL_Y_POSITIONS.map(enemyYposition => new Enemy(ENEMY_START_POS_X, enemyYposition, ENEMY_SPRITE)); // assuming ENEMY_SPRITE is defined
```

Population of `ENEMIES_INITIAL_Y_POSITIONS` could be then further optimized as there is a clear patterns there.

However, I have no doubts you are able to optimize your code further and make DRYer, thus approving.

PR #143, opened 2020-11-19
Memory pair game

xT0PERx:
### Memory pair game

[Demo](https://xt0perx.github.io/Memory-pair-game/) |
[Code base](https://github.com/xT0PERx/Memory-pair-game)

Please, review.


submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,116 @@
+const imageses = [
+  'images/Alfa-romeo.jpeg',
+  'images/Ford.jpeg',
+  'images/Mazda.jpeg',
+  'images/Nissan.jpeg',
+  'images/Peugeot.jpeg',
+];
+//my random array with my img
+const cardDeck = imageses.concat(imageses);
```
madmaxWMFU, line 9: Also, you can to use [spread](http://jsraccoon.ru/es6-spread-rest) 
` const cardDeck = [...imageses, ... imageses];`

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,116 @@
...
+cardDeck.sort(function () {
+  return 0.5 - Math.random();
+});
```
madmaxWMFU, line 12: Wrap this in a function, maybe named here like `getSortList`

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,116 @@
...
+    img.src = cardDeck[i];
+    fragment.children[i].appendChild(img);
+  }
+  document.body.appendChild(fragment);
+};
+
+// this function flips the cards
+
+const overturnsCard = function () {
+  allCardsOnTheDesk.addEventListener('click', function (event) {
```
madmaxWMFU, line 35: In my opinion, it was better like this 
```
const overturnsCard = ({target}) => {
......
}
allCardsOnTheDesk.addEventListener('click', overturnsCard)

```

`{target}` it is [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), you can read more about this

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,116 @@
...
+    const target = event.target;
+
+    if (target.nodeName !== 'DIV') {
+      return;
+    }
+
+    if (target.classList.contains('card__clicked') === true) {
```
madmaxWMFU, line 42: Сan be simplified `if (target.classList.contains('card__clicked'))`

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,116 @@
...
+        checkImg[1].classList.replace('card__clicked', 'card');
+      }, 400);
+    }
+  }
+  document.body.appendChild(allCardsOnTheDesk);
+};
+
+// displays a message about victory
+function checkWin() {
+  winCount++;
```
madmaxWMFU, line 108: Judging by the name of the function, it should only check the value. I think the increment can be inserted into another part of the code.

madmaxWMFU (changes requested):
Good job!
See comments below

submissions/xT0PERx/Memory pair game/transitions.js
xT0PERx: I understood the idea itself, but I could not implement it. I read the article that you offered me, but unfortunately I couldn't fix the code according to your instructions. For some reason, the code didn't work. I committed it and left it in the file. Perhaps you can explain to me where I made a mistake.

xT0PERx:
The button is not pressed - 'Re-request-rewiev'.

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,139 @@
...
+// allCardsOnTheDesk.addEventListener('click', overturnsCard());
+
+const overturnsCard = function () {
+  allCardsOnTheDesk.addEventListener('click', function ({ target }) {
+    if (target.nodeName !== 'DIV') {
+      return;
+    }
+
+    if (target.classList.contains('card__clicked')) {
+      target.classList.remove('card__clicked');
```
madmaxWMFU, line 58: In the description of the task, a link was attached with an example of flipping cards
Try using an [example](https://davidwalsh.name/css-flip)

madmaxWMFU (changes requested):

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+  document.querySelector('main').appendChild(gameBoard);
+};
+
+function CountingMatchingPairsOfCards() {
+  winCount++;
+}
+
+function checkWin() {
+  if (winCount === 5) {
```
madmaxWMFU, line 109: Also a magic number.
As for me is best `const winPoint = imageses.length;`

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,122 @@
...
+    const secondImg = checkImg[1].firstChild;
+
+    if (firstImg.src === secondImg.src) {
+      timer = setTimeout(function () {
+        firstImg.remove();
+        checkImg[0].classList.remove('card__clicked');
+        secondImg.remove();
+        checkImg[1].classList.remove('card__clicked');
+      }, 400);
+      CountingMatchingPairsOfCards();
```
madmaxWMFU, line 91: Easier to do without function, only `winCount++;`

madmaxWMFU (changes requested):

xT0PERx:
I tried to get rid of the fragment and move on to comparison.

madmaxWMFU (approved):
Well done. I think you knew something new for yourself.

madmaxWMFU:
@kottans/fe-2021-mentors, please, can someone else check a work?

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,94 @@
...
+
+const cardDeck = [...images, ...images];
+const duration = 600;
+let winCount = 0;
+let firstOpenCard = null;
+let isBlock = false;
+const pointsWon = images.length;
+const gameBoard = document.querySelector('.cards');
+
+function getSortList() {
```
AMashoshyna, line 17: Usually method's name starting with `get-` implies returning something. This method does not return anything, instead it modifies data in place. I would suggest renaming it to `sortList`. Alternatively you can keep the name and return list from the function.

submissions/xT0PERx/Memory pair game/transitions.js
```diff
@@ -0,0 +1,94 @@
...
+
+const putImgsInCards = function () {
+  cardDeck.forEach(function (value) {
+    template = `
+        <div class="card">
+            <img class='front front-images' src='images/cardClosed.png' alt='front-card'>
+            <img class='back back__image' src='${value}' alt='back-card'>
+        </div>         
+      `;
+    gameBoard.insertAdjacentHTML('beforeEnd', template);
```
AMashoshyna, line 31: This code modifies DOM in a loop, which is a heavy operation. Let's try to avoid it by creating a single element and then appending it to the DOM. Take a look at [Document Fragment API](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment).

AMashoshyna (commented):
@xT0PERx Good job! Left to comments.

PR #149, opened 2020-11-20
OOP

nadsatt:
### post-oop

[Demo](https://nadsatt.github.io/a-tiny-JS-world) |
[Code base](https://github.com/nadsatt/frontend-2021-homeworks/tree/post-oop/submissions/nadsatt/post-oop)

Please, review.

### DIFF
--- a/submissions/nadsatt/post-oop/index.js
+++ b/submissions/nadsatt/post-oop/index.js
@@ -32,7 +32,7 @@ class Inhabitant {
    }
 
    toString(){
-      return `species: ${this.species}, saying: ${this.saying}, gender: ${this.gender}, name: ${this.name}, friends: ${this.friendsManager.getFriends()}`;
+      return Object.entries(this).slice(1).map(([key, value]) => `${key}: ${value}`).join(', ') + `, friends: ${this.friendsManager.getFriends()}`;
    }
 }
 
--- a/submissions/nadsatt/post-oop/index.js
+++ b/submissions/nadsatt/post-oop/index.js
@@ -8,44 +8,31 @@
 
 // ======== OBJECTS DEFINITIONS ========
 
-/* WHAT WAS DONE to make my code SOLID:
-   - FOR S: i created new methods 'addFriends' and 'getFriends' in Inhabitant class and moved it to separate class 'FriendsManager'
-            i also wanted to move 'toString' method to separate class 'InhabitantConverter 
-            but i'm not sure about both ideas because it makes my code strange a bit (especially when i need to create FriendsManager instance for each inhabitant)
-   - FOR O: i changed 'toString' method in 'Inhabitant' to make it dependent on specific properties 
-            to have ability to redefine this method in subclasses
-            to show how this method can be simultaneously closed for modifications in superclass and open for modifications in subclasses
-   - FOR L: as i see, there is no cases where subclass can't be used instead of its superclass
-            for example, in 'addFriends' method of 'FriendsManager' class instance of any subclass can be used
-   - FOR I: as i see, there is no unused/unrelevant superclass' methods/props in subclasses
-   - FOR D: -
-*/ 
+class FriendsManager {
+   constructor(friends = []){
+      this.friends = friends;
+   }
+
+   addFriends(...friends){
+      this.friends.push(...friends);
+   }
+
+   getFriends(){
+      return this.friends.map(friend => friend.name).join(' ') || 'no friends';
+   }
+}
 
 class Inhabitant {
-   constructor(species, saying, gender, name, friends = []){
+   constructor(species, saying, gender, name, friends){
+      this.friendsManager = new FriendsManager(friends);
       this.species = species;
       this.saying = saying;
       this.gender = gender;
       this.name = name;
-      this.friends = friends;
    }
 
    toString(){
-      return `species: ${this.species}, saying: ${this.saying}, gender: ${this.gender}, name: ${this.name}, friends: ${new FriendsManager(this).getFriends()}`;
-   }
-}
-
-class FriendsManager {
-   constructor(inhabitant){
-      this.inhabitant = inhabitant;
-   }
-
-   addFriends(...friends){
-      this.inhabitant.friends.push(...friends);
-   }
-
-   getFriends(){
-      return this.inhabitant.friends.map(friend => friend.name).join(' ') || 'no friends';
+      return `species: ${this.species}, saying: ${this.saying}, gender: ${this.gender}, name: ${this.name}, friends: ${this.friendsManager.getFriends()}`;
    }
 }
 
@@ -117,19 +104,14 @@ class CatWoman extends Cat {
 
 const cat1 = new Cat('female', 'Kitty');
 const dog1 = new Dog('male', 'Bob', cat1);
-const man1 = new Man('Nick');
+const man1 = new Man('Nick'); 
 const woman1 = new Woman('Lucy');
 const woman2 = new Human('i\'m a special woman!', 'female', 'Linda');
 const catWoman1 = new CatWoman('Kate', man1, woman2);
 
-const cat1Friends = new FriendsManager(cat1);
-cat1Friends.addFriends(dog1);
-
-const man1Friends = new FriendsManager(man1);
-man1Friends.addFriends(catWoman1);
-
-const woman2Friends = new FriendsManager(woman2);
-woman2Friends.addFriends(catWoman1);
+cat1.friendsManager.addFriends(dog1);
+man1.friendsManager.addFriends(catWoman1);
+woman2.friendsManager.addFriends(catWoman1);
 
 const inhabitants = [
    man1, 
--- a/submissions/nadsatt/post-oop/index.js
+++ b/submissions/nadsatt/post-oop/index.js
@@ -33,10 +33,10 @@ class Inhabitant {
 
    toString(){
       let props = ['species', 'saying', 'gender', 'name']
-         .map(prop => `${prop}: ${this[prop]}, `)
-         .join('');
-
-      return props + `friends: ${this.friendsManager.getFriends()}, `;
+         .map(prop => `${prop}: ${this[prop]}`)
+         .join(', ');
+         
+      return props + `, friends: ${this.friendsManager.getFriends()}`;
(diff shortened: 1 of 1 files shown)


submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+      this.friends = friends;
+   }
+
+   toString(){
+      return `species: ${this.species}, saying: ${this.saying}, gender: ${this.gender}, name: ${this.name}, friends: ${new FriendsManager(this).getFriends()}`;
+   }
+}
+
+class FriendsManager {
+   constructor(inhabitant){
```
OleksiyRudenko, line 39: **FIX REQUIRED**

Once we have FriendsManager then we could instantiate it in the base class constructor and keep it as a property of the base class. `class Inhabitant{contsructor(...) { this.friendsManager = new FriendsManager(..)...`
This allows you:
- initialize friends via friends manager `constructor(inhabitant, friends = [])`
- you do not need to interact with `this.friends` in base class at all, even in constructor; yet base class will still have the property
- you will refer to `this.friendsManager` in `Inhabitant.toString` instead of instantiation

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+      this.species = species;
+      this.saying = saying;
+      this.gender = gender;
+      this.name = name;
+      this.friends = friends;
+   }
+
+   toString(){
+      return `species: ${this.species}, saying: ${this.saying}, gender: ${this.gender}, name: ${this.name}, friends: ${new FriendsManager(this).getFriends()}`;
+   }
```
OleksiyRudenko, line 35: **FIX REQUIRED**

There is a pattern here `propName: ${this.propName}, `
I wonder if we could make this DRYer, e.g. by employing `Array.map` & `Array.join`?
We'd definitely benefit from it if had e.g. dozens of props to expose here.
`friends` will require special handling we can just attach them after simple props get processed.

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+      super('cat', 'meov-meov!', gender, name, ...friends);
+   }
+}
+
+class CatWoman extends Cat {
+   constructor(name, ...friends){
+      super('female', name, ...friends);
+      this.species = 'human';
+      this.hands = 2;
+      this.legs = 2;
```
OleksiyRudenko, line 110: **DO NOT FIX** anything as per the comment below.

Well, creation of hybrids in traditional class-syntax approach is always a challenge.

Compositional pattern would work much better but that would require you to rewrite the entire codebase into functional style.

You do need to fix anything here. Just FYI to be careful and whenever you see space for hybrids consider switching to compositional pattern instead of classic inheritance pattern.

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+const catWoman1 = new CatWoman('Kate', man1, woman2);
+
+const cat1Friends = new FriendsManager(cat1);
+cat1Friends.addFriends(dog1);
+
+const man1Friends = new FriendsManager(man1);
+man1Friends.addFriends(catWoman1);
+
+const woman2Friends = new FriendsManager(woman2);
+woman2Friends.addFriends(catWoman1);
```
OleksiyRudenko, line 132: **FIX REQUIRED**

If you had FriendsManager instantiated in Inhabitant constructor you would not need to create a new FriendsManager object every time you need to add friends.
Save some memory, take care of JS garbage collector.

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+/* 
+   Use print(message) for output.
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+*/
+
+inhabitants.forEach(inhabitant => print(inhabitant.toString()));
```
OleksiyRudenko, line 152: **FIX REQUIRED**

You do not need to call `.toString` explicitly in string context. It is a built-in magic method defined for the root `Object` in JS, inherited by all object and overridable as you just did.

Try this ``console.log(`${man1}`)``

`print ` function ensures string context for its argument.

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+      return super.toString() + `, hands: ${this.hands}`;
+   }
+}
+
+const cat1 = new Cat('female', 'Kitty');
+const dog1 = new Dog('male', 'Bob', cat1);
+const man1 = new Man('Nick');
+const woman1 = new Woman('Lucy');
+const woman2 = new Human('i\'m a special woman!', 'female', 'Linda');
+const catWoman1 = new CatWoman('Kate', man1, woman2);
```
OleksiyRudenko, line 123: **DO NOT FIX**

**Bonus**
You have sufficient skills and knowledge to comprehend the foolowing fragment.

```
const inhabitants = [[Dog, 1], [Cat, 'a', 'b'], [Human, 7, 8]]
  .map(([className, ...args]) => new className(...args))
```
This however poses a problem: how do we then nicely refer objects to build ties (friends relations) between them?
`inhabitants[1].addFriends(inhabitants[2])` doesn't look really nice.
Well, one approach possible is to employ `Array.find` to find (e.g. by name property) both 'host' friend and a friend to add.

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,152 @@
...
+   }
+}
+
+class FriendsManager {
+   constructor(inhabitant){
+      this.inhabitant = inhabitant;
+   }
+
+   addFriends(...friends){
+      this.inhabitant.friends.push(...friends);
```
OleksiyRudenko, line 44: **DO NOT FIX** anything as per the comment below.

Hmmm, could we make friending mutual? You current objects architecture setup makes it really doable.

OleksiyRudenko (changes requested):
@nadsatt a wow!-class job!
Didn't see that SOLID class-syntax code for this task in this course before.

Although there is some room for improvement, please restrict yourself to the comments marked with **FIX REQUIRED**.

Some comments are for your consideration only.

If you still want to change any code despite the warning **DO NOT FIX** in a comment, please postpone those fixes until you submit your last task in the course.

If you have any questions please let me know.

nadsatt:
@OleksiyRudenko thanks a lot for the review and challenge for further code improvements!

I rewrited my code according to your **FIX REQUIRED** comments. 
Please, re-review.

submissions/nadsatt/post-oop/index.js
```diff
@@ -0,0 +1,122 @@
...
+   constructor(species, saying, gender, name, friends){
+      this.friendsManager = new FriendsManager(friends);
+      this.species = species;
+      this.saying = saying;
+      this.gender = gender;
+      this.name = name;
+   }
+
+   toString(){
+      return Object.entries(this).slice(1).map(([key, value]) => `${key}: ${value}`).join(', ') + `, friends: ${this.friendsManager.getFriends()}`;
```
OleksiyRudenko, line 35: This method now has nice optimization (`.map(....`), although with something that will fire back (`Object.entries(this)`).

The latter part presumes we'd need every property of any child class get printed. What if not? Then we'd need to patch the `Inhabitant` class instead of focusing on the particular child class.

`.slice(1)` also relies upon the fact that `friends` would always be the first property in the object, but this is not guaranteed by design and depends on JS engine in use. Also the purpose of `.slice()` is somewhat cryptic and not really clear when reading the code fragment.

Let's have an explicit list of props here. And let's give the child classes control over their properties back to them.

OleksiyRudenko, line 35: This method now has nice optimization (`.map(....`), although with something that will fire back (`Object.entries(this)`).

The latter part presumes we'd need every property of any child class get printed. What if not? Then we'd need to patch the `Inhabitant` class instead of focusing on the particular child class.

`.slice(1)` also relies upon the fact that `friends` would always be the first property in the object, but this is not guaranteed by design and depends on JS engine in use. Also the purpose of `.slice()` is somewhat cryptic and not really clear when reading the code fragment.

Let's have an explicit list of props here. And let's give the child classes control over their properties back to them.

OleksiyRudenko (changes requested):
Well done with the changes.
A bit overdone though. Let's make child classes great again.

submissions/nadsatt/post-oop/index.js
nadsatt: Should i do smth like this?

**IN BASE CLASS**
```
toString(){
      return ['species', 'saying', 'gender', 'name'].map(prop => `${prop}: ${this[prop]}`).join(', ') + `, friends: ${this.friendsManager.getFriends()}`;
}
```
**IN CHILD CLASS**
```
toString(){
    return super.toString() + ['hands', 'legs'].map(prop => `${prop}: ${this[prop]}`).join(', ');
}
```

OleksiyRudenko: @nadsatt you've caught the idea perfectly fine.
Please pay special attention to code style / code formatting for the sake of code readability.

OleksiyRudenko (changes requested):
A tiny fix yet to shine.

**The problem:**
image

**The solution:**
(1) use `.join(", ")` not just `.join('')` - this way you avoid trailing separators where you do not really need those
(2) in child classes in the overloaded method prepend added data with a separator or have smth like `return [super.method(), extraData1, extraData2].join(separator)` or even nested
```
return 
  [super.method(),
    [extraData1, extraData2].map(...).join(separator)
  ].join(separator)`
```
( ^^ properly fomatted for readability)

The latter looks verbose but scalable (easy expanding class props list, easy copy-paste method to a new class) and less error prone.

OleksiyRudenko (approved):
@nadsatt you did a great job! Approved.
Take a moment to take a look at your initial code, compare with the latest version, reflect on what you've learned and put some notes in your student's diary repo.

PR #151, opened 2020-11-21
Nanodegree arcade game

xT0PERx:
### Nanodegree arcade game

[Demo](https://xt0perx.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/xT0PERx/frontend-nanodegree-arcade-game)

Please, review.


submissions/xT0PERx/Nanodegree-arcade-game/app.js
```diff
@@ -0,0 +1,111 @@
...
+  Y_START_POSITION_ENEMY[1],
+  SPEED_ENEMY[1],
+  player
+);
+const enemie3 = new Enemy(
+  X_START_POSITION_ENEMY,
+  Y_START_POSITION_ENEMY[2],
+  SPEED_ENEMY[2],
+  player
+);
```
OleksiyRudenko, line 98: Can you optimize this fragment employing `Array` methods? There is a pattern in every enemy creation and also some avoidable repetition. Note, that e.g. `.map` callback specification provides you with an element index you could use.

OleksiyRudenko (changes requested):
@xT0PERx a really good job!
The code is readable in plain text thanks to well named constants.

I touch of improvemnt to make it shine.

submissions/xT0PERx/Nanodegree-arcade-game/app.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+Player.prototype.goStartPosition = function () {
+  this.x = X_START_POSITION_PLAYER;
+  this.y = Y_START_POSITION_PLAYER;
+};
+
+const player = new Player(X_START_POSITION_PLAYER, Y_START_POSITION_PLAYER);
+
+const countEmemies = 3;
```
OleksiyRudenko, line 81: A typo obviously

submissions/xT0PERx/Nanodegree-arcade-game/app.js
```diff
@@ -0,0 +1,102 @@
...
+const allEnemies = Array(countEmemies)
+  .fill()
+  .map(function (enemy, index) {
+    return new Enemy(
+      X_START_POSITION_ENEMY,
+      Y_START_POSITION_ENEMY[index],
+      SPEED_ENEMY[index],
+      player
+    );
+  });
```
OleksiyRudenko, line 91: If you had an array of enemies settings like e.g.
```
const ENEMIES_INITIAL_SETTINGS = [{
  start_y_position: 100500,
  speed: 100500,
  }, {
  START_Y_POSITION: 100500,
  SPEED: 100500,
  } // etc
];
```
this fragment might have looked more elegant, e.g.
```
const enemies = 
  ENEMIES_INITIAL_SETTINGS.map(({start_y_position, speed}) => 
    new Enemy(ENEMY_START_X_POSITIOn, start_y_position, speed, .....)
  );
```

It's a bit unobvious to use `Array(x).fill()...` to replace iteration over the list of same-rank values.

OleksiyRudenko (changes requested):
@xT0PERx let's polish it

submissions/xT0PERx/Nanodegree-arcade-game/app.js
xT0PERx: Why typos? I create a player variable with given positions where it starts on the x-axis and the y-axis. 79 line.
And the second variable denotes the number of enemies 81 lines.

submissions/xT0PERx/Nanodegree-arcade-game/app.js
xT0PERx: I agree, my optimization is somewhat more difficult to understand than yours. But I'm learning - at that moment I managed to come up with what I could. Should I remake this piece of code after your example?

submissions/xT0PERx/Nanodegree-arcade-game/app.js
OleksiyRudenko: countE**m**emies

xT0PERx: You are right - I was doubly not attentive. My fault

submissions/xT0PERx/Nanodegree-arcade-game/app.js
```diff
@@ -0,0 +1,102 @@
...
+  this.x = X_START_POSITION_PLAYER;
+  this.y = Y_START_POSITION_PLAYER;
+};
+
+const player = new Player(X_START_POSITION_PLAYER, Y_START_POSITION_PLAYER);
+
+const countEnemies = 3;
+const allEnemies = Array(countEnemies)
+  .fill()
+  .map(function (enemy, index) {
```
OleksiyRudenko, line 84: Your IDE/linter should have warned you that `enemy` is an unused variable.
Call back signature promises we may expect that the array by that moment contains elements, each of which represents an enemy. In fact it is not. The array contains elements that are `undefined` each. It is misleading. Imagine this `.map` doesn't follow `.fill()` immediately but instead an array of undefineds is passed over to some function defined in a completely different part of code base and gets mapped there. A peer developer will spend some time trying to debug and understand why all elements are `undefined` while they are expected to be something that identifies an enemy.

In a nutshell, this `.map` here doesn't transform incoming array elements into something new but rather replaces them with entities that do not depend on incoming array elements at all.

`Array#map` only makes sense where outcoming array elements depend on incoming array elements.

The only useful thing here is index or rather an iteration through a sequence of integers starting with `0` limited with a certain count.

This is a clear use case to employ the `for` operator. `Array#push` is still a legal method to use in cases like this.

OleksiyRudenko (approved):
@xT0PERx approved with reservations.

Check the comment below to take a learning.

submissions/xT0PERx/Nanodegree-arcade-game/app.js
OleksiyRudenko: There is yet another approach to avoid `Array#push` similar to what you had initially, yet not really elegant as well but more obvious:
`const someEntities = Array.from({length: numberOfEntitiesToCreate}, (_, i) => new Entity(parameters, i, evenMoreParameters))`
In fact it is equivalent to `Array().fill().map()` yet reads better since it is all-in-one.
Although this approach is questionnable.

xT0PERx: Thanks a lot, good full explanation of "why it is better not to do this". I will try to continue to write code first of all for people and for ease of maintenance of the code in the future, and not just for it to work.

PR #160, opened 2020-11-22
DOM API is ready for review

m-ruslan:
DOM API

[Demo](https://m-ruslan.github.io/DOM_Practice/#) |
[Code base](https://github.com/m-ruslan/DOM_Practice)

Please, review.


submissions/m-ruslan/js-dom/app.js
```diff
@@ -0,0 +1,86 @@
+document.addEventListener("DOMContentLoaded", () => {
+  console.log("DOM is loaded");
+  loadPageData("info");
+
+  for (i in pageSections) {
+    document
+      .querySelector("#" + pageSections[i].id)
+      .addEventListener("click", (event) => {
+        console.log(event.path[0].id);
```
AMashoshyna, line 9: Here is what I see when I open the site in Firefox and click on menu items: 

image

The problem is that `MouseEvent.path` is not a standard property and is not supported in many browsers. Don you need this particular property or maybe, for example, `Event.currentTarget` would be ok?

submissions/m-ruslan/js-dom/app.js
```diff
@@ -0,0 +1,86 @@
+document.addEventListener("DOMContentLoaded", () => {
+  console.log("DOM is loaded");
+  loadPageData("info");
```
AMashoshyna, line 3: I would put it in a constant, something like "DEFAULT_PAGE", to minimize risk of typos or inconsistent changes throughout the code.

submissions/m-ruslan/js-dom/app.js
```diff
@@ -0,0 +1,86 @@
...
+
+  for (i in pageSections) {
+    document
+      .querySelector("#" + pageSections[i].id)
+      .addEventListener("click", (event) => {
```
AMashoshyna, line 8: I suggest to user [event delegation](https://learn.javascript.ru/event-delegation) instead of adding a listener on every menu item here. Let me know if you need more detailed explanation

AMashoshyna (commented):
Great job! I left some comments, please take a look.

As a possible next step in this task you can render the menu from js instead of hard-coding it in index.html file. Your solution is ok, so please change it only if you are interested. Here is why I suggest it: now you have 2 separate sources of truth as for menu items list, ordering etc - html code and `pageSections` variable in js. What if you need to add a page? You would have to change both html & js (and it may happen that you forget one of them and introduce a bug). Keeping one source of truth will help you avoid errors. Also js-based rendering will be first step to SPA & React for later parts of the course.

m-ruslan:
@AMashoshyna thank you for the review!)

> As a possible next step in this task you can render the menu from js instead of hard-coding it in index.html file.

I've changed the code, now menu is rendered by js. Could you please check?

submissions/m-ruslan/html-css-popup/app.css
```diff
@@ -114,6 +114,8 @@ a {
 .apps-cont {
   width: 100%;
   height: 100%;
+  overflow-y: auto;
+  overflow-x: hidden;
```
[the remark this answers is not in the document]
m-ruslan: Thank you Anastasiya.
I've removed changes in previous task from this PR.
I will create new PR for changes in previous task.

submissions/m-ruslan/js-dom/app.js
m-ruslan: Thank you
I've changed to `Event.target` . I've thried `Event.currentTarget` at first, but when i added the event deligation my code hasn't worked with `Event.currentTarget` but it works with `Event.target` (currentTarget was `.nav-container` and target was `a`)

submissions/m-ruslan/js-dom/app.js
m-ruslan: Thank you
I've added constant for this

submissions/m-ruslan/js-dom/app.js
m-ruslan: Thank you.
I've changed code for adding event delegation. Could you please check?

submissions/m-ruslan/js-dom/app.js
```diff
@@ -0,0 +1,142 @@
...
+    .querySelector(".nav-container")
+    .addEventListener("click", (event) => {
+      let aEl = event.target.closest("a");
+      if (!aEl) return;
+      let section = aEl.id;
+      loadPageData(section);
+    });
+});
+
+const DEFAULT_PAGE = () => loadPageData("info");
```
AMashoshyna, line 16: No, I meant to put `"info"` in a constant:
```javascript
const DEFAULT_PAGE = "info";

loadPageData(DEFAULT_PAGE)
```
This way a) risk of typos is reduced b) if you decide the other page to be default it will be easier to make this change

m-ruslan: Thank you Anastasiya)
I've changed constant, could you please check?

AMashoshyna:
Well done!

m-ruslan:
@AMashoshyna thank you)

PR #161, opened 2020-11-22
Tiny js world

ses3332002:
### A Tiny JS World

[Demo](https://ses3332002.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/ses3332002/a-tiny-JS-world)

Please, review.


submissions/ses3332002/tiny-js-world/index.js
```diff
@@ -0,0 +1,105 @@
...
+    "Puss"
+  ]
+};
+
+const inhabitants = [dog, cat, man, woman, catWoman];
+
+inhabitants.forEach((item) => {
+  let description = "";
+  for (let key in item) {
+    if (typeof(item[key]) == "object") {
```
OleksiyRudenko, line 79: Perhaps `Array.isArray` would work better here.

submissions/ses3332002/tiny-js-world/index.js
```diff
@@ -0,0 +1,105 @@
...
+  friends: [
+    "Puss"
+  ]
+};
+
+const inhabitants = [dog, cat, man, woman, catWoman];
+
+inhabitants.forEach((item) => {
+  let description = "";
+  for (let key in item) {
```
OleksiyRudenko, line 78: Just replace unobvious iteration across properties with an excplicit list of properties, map them into values from current object (with a special handling of nested arrays just like you did in code fragment below), and join the result with a delimeter.

Code will become cleaner and more obvious.

OleksiyRudenko (changes requested):
@ses3332002 neat objects setup.
THe challenge here is process them nicely and neately.

submissions/ses3332002/tiny-js-world/index.js
ses3332002: :) Well, may be it works better, but the result is exactly the same.
I will remake it with Array.isArray.

submissions/ses3332002/tiny-js-world/index.js
ses3332002: Do you mind some like this:
inhabitants.forEach((item) => {
  print(`${item.species}; ${item.name}; ${item.gender}`);
});
But I wanted to make an universal solution that could print all the properties from all the objects.
Ok, I can remake it.

submissions/ses3332002/tiny-js-world/index.js
OleksiyRudenko: Think of future extension. If someone adds a property like `this.wings = { span: 30, canfly: true };` this code will try to apply `.join` and throw an exception.

Whereas there is an appropriate and more precise tool, just use it to have less headache maintaining codebase. Take care of yourself future if not of your peer developers.

submissions/ses3332002/tiny-js-world/index.js
OleksiyRudenko: This is better. It will print exactly the props we want. This makes us safe adding props that we may not want to print.
It refers properties explicitly which is helpful when you revert to the code back after centuries and you can **quickly** understand what happens here **exactly**.

Also there is a pattern here: `` `${item.species}; ${item.name}; ${item.gender}` ``
Can you make this part DRYer? E.g. employing `Array` methods.

submissions/ses3332002/tiny-js-world/index.js
OleksiyRudenko: Both `console.log(0); console.log(1); console.log(2); console.log(3); ...` and `for (let i = 0; i < 3; i++) console.log(i);` also give exactly the same result ;-)

submissions/ses3332002/tiny-js-world/index.js
ses3332002: I think I can make it more DRY if I convert objects to arrays and use some array method for all props (like .map()). But in this case I will not be able to select which props to print and which not to print, and in fact this will become my old code but with "Array.map()" method instead of loop "for (let key in object)".
Sorry, but I don't understand which way you want me to go - list all props "by hands" or use automation, but some specific automation, not the one I used before.

submissions/ses3332002/tiny-js-world/index.js
ses3332002: I remade array checking with Array.isArray.

submissions/ses3332002/tiny-js-world/index.js
ses3332002: I remade it with ability of choosing props to print and without repeating any pattern parts.

ses3332002: Thank for hints in AMA session :)
I hope I understood you right.

submissions/ses3332002/tiny-js-world/index.js
```diff
@@ -0,0 +1,106 @@
...
+inhabitants.forEach((item) => {
+  let description = "";
+  propsToPrint.forEach((prop) => {
+    if (Array.isArray(item[prop])) {
+      description = description + item[prop].join(', ');
+    } else {
+      description = description + item[prop] + "; ";
+    };
+  });
+  print(description);
```
OleksiyRudenko, line 86: `.forEach` assumes side effects (e.g. reference to variables defined beyond the forEach callback). Side effects should be avoided whenever possible.

String concatenation is an error prone approach (like here, freinds list will never be terminated with `';'` if the property appears in the middle of the list).

Consider the following simplified fragments to build a cleaner code.

`someString = props.map(propName => item[propName] /* and other magic for array props*/).join("some delimiter");`

Consider mastering `Array` methods. They are powerful and help creating cleaner code.

OleksiyRudenko (changes requested):
@ses3332002 looks much better. Let's make code cleaner and less error prone.

submissions/ses3332002/tiny-js-world/index.js
ses3332002: I haven't seen your review yet, but I was in a good mood :) and improved the code. 
Now beautiful output does not depend on where the array is located and whether it exists at all.

submissions/ses3332002/tiny-js-world/index.js
```diff
@@ -0,0 +1,98 @@
...
+  friends: [
+    "Puss"
+  ]
+};
+
+const inhabitants = [dog, cat, man, woman, catWoman];
+const propsToPrint = ["species", "name", "gender", "legs", "hands", "friends", "saying"];
+
+inhabitants.forEach(item => {
+  print(Object.entries(item).filter(entry => propsToPrint.includes(entry[0])).map(entry => Array.isArray(entry[1]) ? entry[1].join(', ') : entry[1]).join('; '));
```
OleksiyRudenko, line 78: You are not required to fix it here. Just consider emplying formatting to improve readability.
E.g.
```
array
  .filter(..)
  .map(...)
  .join(...)
```

Consider using [prettier](https://prettier.io/) to do this job for you. Some of the IDEs offer on-the-fly formatting.

Here `entry[0]` and `entry[1]` contain values for entities that are different by nature. Use destructuring to have named variables instead of cryptic referencing by index, thus improving readability. E.g. `Object.entries(...).filter(([key, value]) => ... ...)`

You may also skip an unused part of an entry. E.g.
`Object.entries(...).map(([, value]) => ... .isArray(value) ...)`

OleksiyRudenko (changes requested):
@ses3332002 well done with code optimization. Let's polish the details.

submissions/ses3332002/tiny-js-world/index.js
ses3332002: I remade it.

OleksiyRudenko (approved):
@ses3332002 great job!

PR #171, opened 2020-11-25
Friends App - #done

evgenii-del:

### Friends App

[Demo](http://worried-band.surge.sh/) |
[Code base](https://github.com/evgenii-del/kottans-friends-app)

Please, review.

### DIFF
--- a/submissions/evgenii-del/friends-app/js/main.js
+++ b/submissions/evgenii-del/friends-app/js/main.js
@@ -32,7 +32,7 @@ const createBlock = (friend) => {
     return block;
 }
 
-const renderItem = (friends) => {
+const renderItems = (friends) => {
     main.innerHTML = "";
     let fragment = document.createDocumentFragment();
     friends.forEach((friend) => {
@@ -42,14 +42,12 @@ const renderItem = (friends) => {
     main.append(fragment);
 }
 
-const getUsers = () => {
+const fetchFriends = () => {
     const headers = {
         dataType: 'json'
     }
-    fetch('https://randomuser.me/api/?inc=gender,name,phone,dob,picture&results=12', headers)
+    return fetch('https://randomuser.me/api/?inc=gender,name,phone,dob,picture&results=12', headers)
         .then(response => response.json())
-        .then(response => friendsArr = response.results)
-        .finally(() => renderItem(friendsArr));
 }
 
 const sortByName = (arr, type) => {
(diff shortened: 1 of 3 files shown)


submissions/evgenii-del/friends-app/js/main.js
```diff
@@ -0,0 +1,146 @@
...
+    main.append(fragment);
+}
+
+const getUsers = () => {
+    const headers = {
+        dataType: 'json'
+    }
+    fetch('https://randomuser.me/api/?inc=gender,name,phone,dob,picture&results=12', headers)
+        .then(response => response.json())
+        .then(response => friendsArr = response.results)
```
AMashoshyna, line 51: I see why you use this approach, but let's try to handle asynchronous code in a more traditional manner (this approach works for this small app, but would fail in a bigger project). 
What you are trying to do is get data asynchronously, but then use it outside asynchronous context. More common approach would be to make all you code asynchronous.
I suggest two possible solutions:
- wrap all of you app in `async` function, use `await` inside it and write code that looks like synchronous
- wrap logic of your app in a function that receives data as an input and pass this function to `then` of fetch's promise:
```javascript
function runApp(data) {
   // render friends list from data
   // attach event litseners for filters and rerender when filter is applied
 
}

fetchFriends().then(runApp)
```

submissions/evgenii-del/friends-app/js/main.js
```diff
@@ -0,0 +1,146 @@
...
+        .finally(() => renderItem(friendsArr));
```
AMashoshyna, line 52: why do you prefer `finally` to `then` in this case?

submissions/evgenii-del/friends-app/js/main.js
```diff
@@ -0,0 +1,146 @@
...
+                    <div class="card__bottom">
+                        <p>Age: ${friend.dob.age}</p>
+                        <p>Gender: ${friend.gender}</p>
+                    </div>
+                </div>
+        `;
+    return block;
+}
+
+const renderItem = (friends) => {
```
AMashoshyna, line 35: `renderItem` usually suggests that you render one item out of a list, but in this case you render complete list

AMashoshyna (commented):
Awesome! Please see my suggestions below.

submissions/evgenii-del/friends-app/js/main.js
evgenii-del: This is mistake. I have not fully understood fetch

MaxKalashnyk:
@AMashoshyna , are we good with this PR?

AMashoshyna:
Yes, sorry for delyayed reply. Approved

evgenii-del:
Thanks!

PR #174, opened 2020-11-27
DOM JS

zalevskyi:
### DOM

[Demo](https://zalevskyi.github.io/kottans-frontend/practice/dom/index.html) |
[Code base](https://github.com/zalevskyi/frontend-2021-homeworks/tree/dom/submissions/zalevskyi/dom)

Please, review.

### DIFF
--- a/submissions/zalevskyi/dom/index.html
+++ b/submissions/zalevskyi/dom/index.html
@@ -6,18 +6,7 @@
         <link rel='stylesheet' href='./css/app.css'>
     </head>
     <body class='page'>
-        <nav id='nav' class='nav hidden_small'>
-            <ul id='menu' class='menu_list'>
-                <li><a href='#' class='menu_item menu_item_selected menu_item_first'>governors</a></li>
-                <li><a href='#' class='menu_item'>amani</a></li>
-                <li><a href='#' class='menu_item'>liang</a></li>
-                <li><a href='#' class='menu_item'>magnus</a></li>
-                <li><a href='#' class='menu_item'>moksha</a></li>
-                <li><a href='#' class='menu_item'>pingala</a></li>
-                <li><a href='#' class='menu_item'>reyna</a></li>
-                <li><a href='#' class='menu_item menu_item_last'>victor</a></li>
-            </ul>
-        </nav>
+        <nav id='nav' class='nav hidden_small_screen'></nav>
         <header class='header'>
             <div class='disclaimer'>Document Object Model Practice task | <a href="../../index.html" class='active_element disclaimer_link'>Back to all tasks</a></div>
             <div class='title'>Civilization VI - Civilopedia</div>
--- a/submissions/zalevskyi/dom/index.html
+++ b/submissions/zalevskyi/dom/index.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html lang='en' class='full_screen_height'>
+    <head>
+        <title>Civilization VI - Governors</title>
+        <meta name='viewport' content='width=device-width, initial-scale=1'>
+        <link rel='stylesheet' href='./css/app.css'>
+    </head>
+    <body class='page'>
+        <nav class='nav hidden_small'>
+            <ul id='menu' class='menu_list'>
+                <li><a href='#' class='menu_item menu_item_selected menu_item_first'>governors</a></li>
+                <li><a href='#' class='menu_item'>amani</a></li>
+                <li><a href='#' class='menu_item'>liang</a></li>
+                <li><a href='#' class='menu_item'>magnus</a></li>
+                <li><a href='#' class='menu_item'>moksha</a></li>
+                <li><a href='#' class='menu_item'>pingala</a></li>
+                <li><a href='#' class='menu_item'>reyna</a></li>
+                <li><a href='#' class='menu_item menu_item_last'>victor</a></li>
+            </ul>
+        </nav>
+        <header class='header'>
+            <div class='disclaimer'>Document Object Model Practice task | <a href="../../index.html" class='active_element disclaimer_link'>Back to all tasks</a></div>
+            <div class='title'>Civilization VI - Civilopedia</div>
+        </header>
+        <main class='main'>
+            <div id='menu_button' class='menu_button'><a href='#'>Menu</a></div>
+            <div id='civilopedia_path' class='civilopedia_path'>Civilization VI | Governors</div>
+            <div id='article' class='article'>
+                <h1 id='title' class='text_heading_1'>Governors - Overview</h1>
+                <p id='description' class='text'>
+                    Governors are a feature added to Civilization VI in the Rise and Fall expansion pack.
+                    They are agents of your government which may be appointed and assigned to cities in order to boost their Loyalty and enhance gameplay elements.
+                    As such they directly channel the power of the central authority in their areas of specialty.
+                </p>
+            </div>
+        </main>
+        <script src='./js/app.js'></script>
+    </body>
+</html>
--- a/submissions/zalevskyi/dom/index.html
+++ b/submissions/zalevskyi/dom/index.html
@@ -6,7 +6,7 @@
         <link rel='stylesheet' href='./css/app.css'>
     </head>
     <body class='page'>
-        <nav class='nav hidden_small'>
+        <nav id='nav' class='nav hidden_small'>
             <ul id='menu' class='menu_list'>
                 <li><a href='#' class='menu_item menu_item_selected menu_item_first'>governors</a></li>
                 <li><a href='#' class='menu_item'>amani</a></li>
@@ -22,18 +22,7 @@
             <div class='disclaimer'>Document Object Model Practice task | <a href="../../index.html" class='active_element disclaimer_link'>Back to all tasks</a></div>
             <div class='title'>Civilization VI - Civilopedia</div>
         </header>
-        <main class='main'>
-            <div id='menu_button' class='menu_button'><a href='#'>Menu</a></div>
-            <div id='civilopedia_path' class='civilopedia_path'>Civilization VI | Governors</div>
-            <div id='article' class='article'>
-                <h1 id='title' class='text_heading_1'>Governors - Overview</h1>
-                <p id='description' class='text'>
-                    Governors are a feature added to Civilization VI in the Rise and Fall expansion pack.
-                    They are agents of your government which may be appointed and assigned to cities in order to boost their Loyalty and enhance gameplay elements.
-                    As such they directly channel the power of the central authority in their areas of specialty.
-                </p>
-            </div>
-        </main>
+        <main id='main' class='main'></main>
         <script src='./js/app.js'></script>
     </body>
 </html>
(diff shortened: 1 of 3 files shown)


submissions/zalevskyi/dom/index.html
```diff
@@ -0,0 +1,39 @@
...
+                <li><a href='#' class='menu_item'>reyna</a></li>
+                <li><a href='#' class='menu_item menu_item_last'>victor</a></li>
+            </ul>
+        </nav>
+        <header class='header'>
+            <div class='disclaimer'>Document Object Model Practice task | <a href="../../index.html" class='active_element disclaimer_link'>Back to all tasks</a></div>
+            <div class='title'>Civilization VI - Civilopedia</div>
+        </header>
+        <main class='main'>
+            <div id='menu_button' class='menu_button'><a href='#'>Menu</a></div>
```
MaxKalashnyk, line 26: Make this element as a button. It will be more correct semantically.

submissions/zalevskyi/dom/index.html
```diff
@@ -0,0 +1,39 @@
...
+            <div id='civilopedia_path' class='civilopedia_path'>Civilization VI | Governors</div>
+            <div id='article' class='article'>
```
MaxKalashnyk, line 28: Generate the content from this section from your script file, please

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,110 @@
...
+                 <img id='portrait' alt='' src='' class='portrait'>
+                 <h2 class='text_heading_2 description_heading'>Description</h2>
+                 <p id='description' class='text description'></p>
+                 <h2 class='text_heading_2 context_heading'>Historical Context</h2>
+                 <p id='context' class='text context'></p>`
+}
+Object.freeze(WIKI)
+Object.freeze(ARTICLE_INNER_HTML)
+function updateCommon(name) {
+    document.getElementById('title').textContent = `${name} - ${WIKI[name].title}`
```
MaxKalashnyk, line 78: Make all of these elements constants. It would decrease the number of DOM manipulations on each click.

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,110 @@
...
+    document.getElementById('description').textContent = WIKI[name].description
+}
+function updateGovernor(name) {
+    document.getElementById('context').textContent = WIKI[name].context
+    document.getElementById('portrait').src = WIKI[name].img_src
+    document.getElementById('portrait').alt = WIKI[name].img_alt
+}
+document.getElementById('menu').addEventListener('click', ({target}) => {
+    const target_before = document.querySelector('.menu_item_selected')
```
MaxKalashnyk, line 87: Why do you check for the previous clicked element? 
To my mind, it's a redundant operation. 
You can get rid of all checking of the previous elements and just switch content.

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,110 @@
...
+    if (target!==target_before) {
+        const article = document.getElementById('article')
+        const name = target.textContent
+        const name_before = target_before.textContent
+        target_before.classList.toggle('menu_item_selected')
+        target.classList.toggle('menu_item_selected')
+        article.classList.add('hidden')
```
MaxKalashnyk, line 94: Why do you hide this element and then display it?

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+    document.querySelector('.main').classList.toggle('hidden_small')
+    document.querySelector('.nav').classList.toggle('hidden_small')
+})
+document.getElementById('menu').addEventListener('click', e => e.preventDefault())
+document.getElementById('menu_button').addEventListener('click', () => {
+    document.querySelector('.main').classList.toggle('hidden_small')
+    document.querySelector('.nav').classList.toggle('hidden_small')
+})
+document.getElementById('menu_button').addEventListener('click', e => e.preventDefault())
```
MaxKalashnyk, line 110: You can remove this line when you change this element to the button

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,110 @@
...
+            article.innerHTML=ARTICLE_INNER_HTML[WIKI[name]['type']]
+        }
+        updateCommon(name)
+        if (WIKI[name]['type']==='governor') updateGovernor(name)
+        article.classList.remove('hidden')
+    }
+    document.querySelector('.main').classList.toggle('hidden_small')
+    document.querySelector('.nav').classList.toggle('hidden_small')
+})
+document.getElementById('menu').addEventListener('click', e => e.preventDefault())
```
MaxKalashnyk, line 105: no need to add so much event listeners. It can descrease the performance of your app.
Let's move this operation to the previous event listener

MaxKalashnyk (changes requested):
Nice work! See my comments below

submissions/zalevskyi/dom/js/app.js
zalevskyi: Hiding and then showing again it was from
Udacity: JavaScript and the DOM: Lesson 4 Performance: 3 Reflow & Repaint [link](https://classroom.udacity.com/courses/ud117/lessons/364ca3da-1eb9-4213-beb4-7791ac4e1ea4/concepts/d1bee920-e33f-41db-aac3-2c3c5a08db3b)
> we could hide #comments, delete the spam, and show it again -- that's surprisingly fast, to the cost of one reflow and two repaints (and little else). It's fast because hiding doesn't change the layout, it just erases that section of the screen (1 repaint). When you make the changed section visible again, that's a reflow and a repaint.

Any way I have removed it and simply added image size into CSS to avoid reflow after image loading. For this case it works just fine.
Thank you for paying attention to it.

submissions/zalevskyi/dom/js/app.js
zalevskyi: Agree.
That was overthinking about performance improvement in extremely rare situation of selecting already selected item.

zalevskyi:
Thank you for the comments!
They were very helpful on performance (constants for elements, number of event listeners) and semantic (button instead of div)

Please review the code once again.

submissions/zalevskyi/dom/js/app.js
MaxKalashnyk: ok, got it. It's up to you. let's leave it. no problem

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,124 @@
...
+                         </div>`
+const main = document.getElementById('main')
+const nav = document.getElementById('nav')
+main.innerHTML=MAIN_INNER_HTML
+const title = document.getElementById('title')
+const description = document.getElementById('description')
+const portrait = document.getElementById('portrait')
+const context = document.getElementById('context')
+const governorHeadings = document.querySelectorAll('.text_heading_2')
+hideGovernorDetails()
```
MaxKalashnyk, line 84: you can divide your constants functions declarations and it's calls by adding empty lines. It can increase readability of your code for others developers.

submissions/zalevskyi/dom/index.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang='en' class='full_screen_height'>
+    <head>
+        <title>Civilization VI - Governors</title>
+        <meta name='viewport' content='width=device-width, initial-scale=1'>
+        <link rel='stylesheet' href='./css/app.css'>
+    </head>
+    <body class='page'>
+        <nav id='nav' class='nav hidden_small'>
+            <ul id='menu' class='menu_list'>
```
MaxKalashnyk, line 10: You can leave it as it is. But next time you can generate this menu from your `WIKI` object

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,124 @@
...
+                         <div id='article' class='article'><h1 id='title' class='text_heading_1'></h1>
+                            <img id='portrait' alt='' src='' class='portrait'>
+                            <h2 class='text_heading_2 description_heading'>Description</h2>
+                            <p id='description' class='text description'></p>
+                            <h2 class='text_heading_2 context_heading'>Historical Context</h2>
+                            <p id='context' class='text context'></p>
+                         </div>`
+const main = document.getElementById('main')
+const nav = document.getElementById('nav')
+main.innerHTML=MAIN_INNER_HTML
```
MaxKalashnyk, line 78: I'd recommend you to use code structure something like this:
- defining constants;
- defining functions;
- calling functions.
It's not a pattern. Just my recommendation. I got used to using this structure in my projects and it can really save time to follow this code and make some changes in future.

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,124 @@
...
+    context.classList.remove('hidden')
+}
+function toggleHiddenSmallNavMain() {
+    main.classList.toggle('hidden_small')
+    nav.classList.toggle('hidden_small')
+}
+document.getElementById('menu').addEventListener('click', evt => {
+    const name = evt.target.textContent
+    const type = WIKI[name]['type']
+    document.querySelector('.menu_item_selected').classList.toggle('menu_item_selected')
```
MaxKalashnyk, line 116: please use correct formatting. it helps to navigate your code easily. 
Also, I'd recommend starting to query this element not from the `document`, but just from the `evt.target`. And as you used `evt.target` multiple times here, you can create a constant here for storing the target value there.

MaxKalashnyk (changes requested):
it looks much better! Please, pay attention to my recommendations and we're ready to merge.

submissions/zalevskyi/dom/js/app.js
```diff
@@ -0,0 +1,142 @@
...
+    navHTML += Object.keys(WIKI).map(element => `<li><a href='#' class='menu_item'>${element}</a></li>`).join('')
+    navHTML +=`</ul>`
+    NAV.innerHTML = navHTML
+    const MENU_ITEMS = document.querySelectorAll('.menu_item')
+    MENU_ITEMS[0].classList.add('menu_item_first')
+    MENU_ITEMS[MENU_ITEMS.length-1].classList.add('menu_item_last')
+    return MENU_ITEMS[0]
+}
+function selectMenuItem(item) {
+    try {
```
MaxKalashnyk, line 97: I'd recommend you to get rid of this try-catch construction here.
No need to store this variable as global because it can be not so easy to check where it was changed etc.
You can pass the `target` to the `updateMain` function and work with it there.
Maybe my suggestion would help you to make this function more clean:
```suggestion
    function selectMenuItem(target) {
       target.classList.remove('menu_item_selected');
       target.classList.add('menu_item_selected');
       updateMain(target);
    }
```

MaxKalashnyk (changes requested):
Nice work! Look at my comments, please

MaxKalashnyk (approved):
LGTM!

PR #176, opened 2020-11-27
Add OOP

VBystrov:
### Object-Oriented JavaScript

[Demo](https://vbystrov.github.io/frontend-nanodegree-arcade-game/index.html) |
[Code base](https://github.com/VBystrov/frontend-nanodegree-arcade-game)

Please, review.


OleksiyRudenko:
@VBystrov please keep only `app.js`  in this PR for it is the only file required for code review.

OleksiyRudenko:
While we are seeking for review capacity feel free checking your code against the following checklist.

Minimal requirements to meet:
         (e.g. `const STEP = 101;`)
 * Requirements re **OOP**:

VBystrov:
> 
> 
> @VBystrov please keep only `app.js` in this PR for it is the only file required for code review.

In my solution I apply changes to engine.js too. So I should add app.js and engine.js into commit or is it prohibited to change something else except app.js?

OleksiyRudenko:
If all required changes were scoped to `app.js` only this would allow completely ignore original engine code thus saving reviewers' time and effort. Consider game engine as something you have no license to change.
Take some time to decide if you can do this easily. Let me know.  Feel free asking questions in these regards if any.

VBystrov:
I can move all changes to app.js but it will take some time. I confirm when it will be ready. 
Also thanks for requirements list, I already see what can be improved.

VBystrov:
@OleksiyRudenko task ready to review.

submissions/VBystrov/task_js_oop/app.js
```diff
@@ -0,0 +1,169 @@
...
+  }
+  this.update(dx, dy);
+};
+
+Player.prototype.checkCollisions = function (enemies) {
+  // coordinates collision
+  const minX = 18;
+  const maxX = 84;
+  const minY = 85;
+  const maxY = 137;
```
OleksiyRudenko, line 100: Worth grouping with other constants. Think of it as of game settings. Most likely user would want have them all together.

submissions/VBystrov/task_js_oop/app.js
```diff
@@ -0,0 +1,169 @@
...
+      enemy.startPosition();
+    });
+    player.startPosition();
+  }
+}
+
+const allEnemies = [];
+for (let i = 0; i < numEnemies; i++) {
+  allEnemies.push(new Enemy());
+  allEnemies[i].startPosition();
```
OleksiyRudenko, line 143: Consider calling `.startPosition` from inside constructor since this is smth called anyway every time when an enemy gets created.
`startPosition` reads ambiguously. If `start` here is a verb then it reads "start to position (itself?)". If it is a noun then we're lacking a verb in method name.

OleksiyRudenko (changes requested):
@VBystrov good job!
A few fixes and we're done.

submissions/VBystrov/task_js_oop/app.js
VBystrov: Done.

submissions/VBystrov/task_js_oop/app.js
VBystrov: Done.

OleksiyRudenko (approved):
@VBystrov well done!

P.S. There is still no [verb](https://en.wikipedia.org/wiki/Verb) in the name of the method responsible for putting an enemy into start position. :) Please keep the rule to start a function/method name with a verb.

PR #178, opened 2020-11-28
Task js oop

webdevtoday:
### task_js_oop

[Demo](https://webdevtoday.github.io/frogger-game/index.html) |
[Code base](https://github.com/webdevtoday/frogger-game)

Please, review.


OleksiyRudenko:
Wasn't able to play the game until I looked into the code. Normally, we do not expect users to look into the code to find out why arrow keys used by intuition do not work.

While we're seeking review capacity please check the code against the following checklist.

Minimal requirements to meet:
 * Requirements re **OOP**:

webdevtoday:
Hi, I fixed problems. And I really forgot to change keys, I fixed it.

OleksiyRudenko (changes requested):
@webdevtoday pretty well done.
However there are still many mysterious numbers.
Define more constants. Think how your code **reads**. The closer to plain English the better.
Also think of constants as of game or app settings: user wants have them all put together and also well grouped.

webdevtoday:
@OleksiyRudenko Hi! I replaced numbers with constants.

submissions/webdevtoday/task_js_oop/app.js
```diff
@@ -0,0 +1,124 @@
...
+    this.sprite = 'images/char-boy.png';
+};
+Player.prototype.update = function(dt) {};
+Player.prototype.render = function() {
+    this.checkCollision();
+    ctx.drawImage( Resources.get(this.sprite), this.x, this.y );
+};
+
+Player.prototype.checkCollision = function() {
+    allEnemies.forEach(enemy => {
```
OleksiyRudenko, line 80: > - [ ] classes do not refer to any global variables, like global variable `player`, which is an instance of `Player` class (referring to global constants and globals provided by the gaming platform like `Resources` is OK); Hint: pass `Player` instance as an argument to every enemy

Always consider inducing specific statements into more general. The purpose of the review is not to specify every single mistake or inconsistency.

submissions/webdevtoday/task_js_oop/app.js
```diff
@@ -0,0 +1,124 @@
...
+        const offset = 20;
```
OleksiyRudenko, line 81: 20 is mysterious number.

Always consider inducing specific statements into more general. The purpose of the review is not to specify every single mistake or inconsistency.

submissions/webdevtoday/task_js_oop/app.js
```diff
@@ -0,0 +1,124 @@
...
+        case 'down':
+            if (this.y <= FIELD_HEIGHT - LIMITING_BOTTOM_PADDING) this.y += this.upDownStepSize;
+            break;
+    }
+};
+
+// Now instantiate your objects.
+// Place all enemy objects in an array called allEnemies
+// Place the player object in a variable called player
+var allEnemies = [new Enemy(-50, 50), new Enemy(-50, 135), new Enemy(-50, 220)];
```
OleksiyRudenko, line 109: Mysterious numbers.

OleksiyRudenko (changes requested):
@webdevtoday please double-check yourself against the checklist.

submissions/webdevtoday/task_js_oop/app.js
webdevtoday: Do I really need to set dynamic coordinate values using constants?

submissions/webdevtoday/task_js_oop/app.js
webdevtoday: Sorry, I feel very stupid when it comes to OOP((

submissions/webdevtoday/task_js_oop/app.js
webdevtoday: Sorry, I came close to a topic that was very difficult for me and now I cannot quickly get used to it, as with the previous topics.

submissions/webdevtoday/task_js_oop/app.js
OleksiyRudenko: I suppose `-50` is some enemy's initial X position and `50`, `135` and `220` are calculable from some base for Y position + row height multiplied by row number.
So you need a couple of base constants and apply calculations.
Something like `for (let row = 0; row < 3; row++)` can be a source of rows (`for` is still a good operator).

submissions/webdevtoday/task_js_oop/app.js
OleksiyRudenko: No worries. You are learning and you are all good while you do.
The next task offers more materials on OOP and SOLID. These are to be followed to keep code maintainable and scalable.
However, this point is not so much about OOP itself (it is indeed) but more about inducing and deducing new information from that already available to you.

submissions/webdevtoday/task_js_oop/app.js
webdevtoday: Ahh, that's what you mean, about the dynamic generation of the enemy. Yes, no problem, I'll change it.

OleksiyRudenko: Imaging, we extend the game so it has different number of rows and dozens of enemies. How would you change your code so that it is scalable and maintainable.

submissions/webdevtoday/task_js_oop/app.js
webdevtoday: I understand you, thank you! I'll try to really understand this interesting topic! Thanks again for your support!

submissions/webdevtoday/task_js_oop/app.js
OleksiyRudenko: No worries, that's why the mentors are here and you are not on your own completely.

webdevtoday: That's true!

submissions/webdevtoday/task_js_oop/app.js
webdevtoday: I'll think about it

OleksiyRudenko: For now focus only on this fragment with regards to scalability.

webdevtoday: Ok, I'll do it

webdevtoday:
@OleksiyRudenko I reread all your tips and edits. I added a new super class and extended it with subclasses, etc. But so far I have no idea at all how to add collision by transferring the player's object to the enemy's objects. But I am still thinking about how to achieve this.

webdevtoday:
@OleksiyRudenko I've done it!

OleksiyRudenko (approved):
@webdevtoday excellent job!

webdevtoday:
@OleksiyRudenko OMG!!! Thank you, thank you very much!!! It was very useful experience!!!! Thank you for it!

PR #188, opened 2020-11-29
HTML&CSS Popup task

kt-std:
### HTML & CSS Popup

[Demo](https://kt-std.github.io/html-css-popup/) |
[Code base](https://github.com/kt-std/html-css-popup)

Please, review.

### DIFF
--- a/submissions/kt-std/html-css-popup/css/style.css
+++ b/submissions/kt-std/html-css-popup/css/style.css
@@ -1,0 +1,370 @@
+*{
+	margin: 0;
+	padding: 0;
+	list-style: none;
+	text-decoration: none;
+	outline: none;
+	font-size: 16px;
+	line-height: 1.2;
+}
+
+body,
+.nav__list,
+.popup__item,
+.checkbox__input:checked ~ .popup,
+.checkbox__input:checked ~ .popup__more{
+	display: flex;
+}
+
+.flex-center{
+	display: flex;
+	justify-content: center;
+	align-items: center;
+}
+
+.full{
+	width: 100%;
+}
+
+.column{
+	flex-direction: column;
+}
+ 
+.box{
+	box-sizing: border-box;
+}
+
+body{
+	font-family: Lato, sans-serif;
+	color: #303030;
+	overflow: hidden;
+	width: 100vw;
+	height: 100vh;
+}
+
+.nav {
+	float: right;
+}
+
+.nav__list,
+.checkbox__image:hover{	
+	padding: 10px;
+}
+
+.nav__item {	
+	min-width: 80px;
+	height: 60px;
+	margin-right: 10px;	
+}
+
+a:visited,
+a,
+a:active{
+	color: #303030;
+}
+
+a:hover{
+	color: rgba(77, 182, 172, 1);
+}
+
+a:hover,
+.more:hover,
+.bell:hover, 
+.search-field__input:hover,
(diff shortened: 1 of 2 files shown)


submissions/kt-std/html-css-popup/index.html
```diff
@@ -0,0 +1,135 @@
...
+			<ul class="nav__list">
+				<li class="nav__item box flex-center"><a href="#mail" class="item__link flex-center column box">Mail</a></li>
+				<li class="nav__item box flex-center"><a href="#pictures" class="item__link flex-center column box">Pictures</a></li>
+				<li class="nav__item box flex-center popup-btn">
+					<input type="checkbox" class="checkbox__input" id="check">
+					<label class="checkbox flex-center full" for="check">
+						<img src="assets/list.svg" alt="List of other applications" class="checkbox__image">
+					</label>					
+					<ul class="popup">
+						<li class="popup__item item">
```
miroslav-kolomiets, line 23: You have unused class `item` here.

miroslav-kolomiets (changes requested):
Looks nice! But something can be improved.
Use formatter with your css code.
Items in menu have different hover styles.
Снимок экрана 2020-11-29 в 10 15 20 AM

submissions/kt-std/html-css-popup/css/style.css
```diff
@@ -0,0 +1,370 @@
...
+}
+
+.button{
+	position: absolute;
+	width: 60px;
+	z-index: 1;
+	right: 0;
+}
+
+button::-moz-focus-inner {
```
miroslav-kolomiets, line 266: And please don't use html tags in css.
Read this article for more understanding: https://frontstuff.io/you-need-to-stop-targeting-tags-in-css

submissions/kt-std/html-css-popup/index.html
kt-std: Thank you reviewing! 
I've deleted unused classes, changed the color of hovering items so that they match each other, and formatted the code.

submissions/kt-std/html-css-popup/css/style.css
kt-std: I was thinking of this case while writing the code. I tore between the usage of tag and the class. I've already updated the code with class, but could you please explain what solution would be better in that case? The case actually is: I decided to use the tag instead of the class because this rule fixes the default outline of a button in Firefox (shown on the screenshot below) so that the style should be applied to all button tags. 
fix
I guess, it matches the idea from the article you sent:

> Now following that logic, there’s one case when styling on tags can actually make sense: when absolutely all tags of a certain type need to be styled the same way (or when there’s only one of them in the DOM, like HTML or body).  

Therefore, _to tag_ or _not to_ _tag_?

lenadgit: Imagine that not only you will support or upgrade this application. And this app will have 10000 lines of code (in additional files too). And someone knows nothing about the styling of tags and will try to add pictures, links, (whatever) with their own styles and will not understand why some of them do not apply or apply not as was desirable
It is a just good practice

kt-std: Thank you for clarifying! :)

miroslav-kolomiets (approved):

miroslav-kolomiets:
Good job!

PR #189, opened 2020-11-29
Frogger game

Lu-sure:
### Frogger game

[Demo](https://lu-sure.github.io/Frogger/) | [Code base](https://github.com/Lu-sure/Lu-sure.github.io/tree/main/Frogger)

Please, review.


OleksiyRudenko:
@Lu-sure looks nice.
I might be wrong, but it sime like the hero gets killed while in the safe area on the top when an enemy is too close. It also mysteriously gets frozen (irresponsive to keypress) when trying to avoid a really close enemy.
There are 5 collision checks across the code. Seems a bit excessive.

OleksiyRudenko:
While we're looking for capacity to review please check yourself against the checlist below and confirm all items are positive. Please, ask a question if anything requires clarification.

## Check-list

Minimal requirements to meet:
 * Requirements re **OOP**:
 * Most common mistakes

Lu-sure:
Made changes according above comments, should I ask about review in FE Questionarium chat once more?

OleksiyRudenko:
> Made changes according above comments, should I ask about review in FE Questionarium chat once more?

No need

submissions/Lu-sure/frogger/app.js
```diff
@@ -0,0 +1,133 @@
...
+};
+
+Enemy.prototype = Object.create(Hero.prototype);
+Enemy.prototype.constructor = Enemy;
+
+Enemy.prototype.update = function (dt) {
+    this.x += this.speed * dt;
+    if (this.x > MOVE_AREA.width.end)
+        this.x = ENEMY_CONFIG.startPointX;
+    checkCollision();
```
OleksiyRudenko, line 55: Not a really good approach, because at every enemy's every move `checkCollision` compares all enemies against the player. So number of comparisons is `numberOfEnemies ** 2`. This will pose a performance issue if number of enemies grows.

A better approach is when every enemy checks itself only against player for collision.

OleksiyRudenko (changes requested):
@Lu-sure well done!
However there is a potential performance issue that can be easily avoided.

submissions/Lu-sure/frogger/app.js
Lu-sure: Fixed, review, please)

OleksiyRudenko (approved):
@Lu-sure great job!

PR #190, opened 2020-11-29
JS Post OOP

ustymchyk:
### JS Post OOP

[Demo](https://ustymchyk.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/ustymchyk/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/ustymchyk/js-post-oop/index.js
+++ b/submissions/ustymchyk/js-post-oop/index.js
@@ -1,6 +1,5 @@
 /* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
    Complete the below for code reviewers' convenience:
-
    Code repository: _put repo URL here_
    Web app: _put project's github pages URL here_
    */
@@ -8,14 +7,19 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 class Creature {
-    constructor(name, gender, age) {
+    constructor(name, gender, age, species) {
         if (new.target === Creature) {
             throw new TypeError("Cannot construct Abstract instances directly");
         }
 
+        this.species = species;
         this.name = name;
         this.gender = gender;
         this.age = age;
+    }
+
+    toString() {
+        return `${this.species}, ${this.name}, ${this.gender}, ${this.age}`;
     }
 
     bornCreature(partner = {}, femaleName, maleName) {
@@ -35,12 +39,12 @@ class Pet extends Creature {
     #angrySound;
     #energy;
 
-    constructor(name, gender, age, happySound, angrySound) {
-        if (new.target === Pet) {
+    constructor(name, gender, age, species, happySound, angrySound) {
+        if (new.target === Creature) {
             throw new TypeError("Cannot construct Abstract instances directly");
         }
 
-        super(name, gender, age);
+        super(name, gender, age, species);
 
         this.#happySound = happySound;
         this.#angrySound = angrySound;
@@ -66,8 +70,8 @@ class Pet extends Creature {
         this._talk(this.#happySound);
     }
 
-    introduce() {
-        return `${this.#happySound} ${this.name} ${this.#happySound} ${this.#happySound} ${this.gender} ${this.#happySound} ${this.age}`;
+    toString() {
+        return `${this.#happySound} ${super.toString()} ${this.#happySound}`;
     }
 
     _talk(sound) {
@@ -82,15 +86,13 @@ class Pet extends Creature {
 // I use _ instead of # in methods for prototype inheritance.
 class Cat extends Pet {
     constructor(name, gender, age) {
-        super(name, gender, age, 'meow', 'shhhh');
-        this.species = 'cat';
+        super(name, gender, age, 'cat', 'meow', 'shhhh');
     }
 }
 
 class Dog extends Pet {
     constructor(name, gender, age) {
-        super(name, gender, age, 'wof', 'grrrr');
-        this.species = 'dog';
+        super(name, gender, age, 'dog', 'wof', 'grrrr');
     }
 }
 
@@ -100,14 +102,14 @@ class Human extends Creature {
     #contacts = [];
 
     constructor(name, gender, age) {
-        super(name, gender, age);
-        this.species = 'human';
+        super(name, gender, age, 'human');
         this.spouse = null;
+        this.children = [];
     }
 
     talkWith(otherHuman) {
         if (otherHuman instanceof Human) {
-            if (!this.#contacts.includes(otherHuman.name)) {
+            if (!this.#contacts.includes(otherHuman)) {
                 this._createContact(otherHuman);
                 otherHuman.talkWith(this);
             }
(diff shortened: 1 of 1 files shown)


ustymchyk:
@OleksiyRudenko  what means 'reference-code' tag?

OleksiyRudenko:
> @OleksiyRudenko  what means 'reference-code' label? 

It denotes code worth referencing to in educational purposes. In other words, code outstanding when compared with the majority of submissions

ustymchyk:
Nice. I've thought that it means that i'm cheating.

submissions/ustymchyk/js-post-oop/index.js
```diff
@@ -0,0 +1,212 @@
...
+            friends = `There is my friends: ${this.getFriendsNames()}. `;
+        }
+
+        if (this.spouse) {
+            spouse = `And i have a spouse ${this.spouse}; `;
+        }
+
+        return intro + friends + spouse;
+
+    }
```
OleksiyRudenko, line 152: It is not illegal that child classes use properties introduced by a base class.
However for the sake of method overloading demonstration let's make all inhabitants intorducing themselves in a similar manner. In terms of hierarchy and OOP principles let's implement it so that every class manages only the part with properties introduced by such base or child class.

OleksiyRudenko (changes requested):
@ustymchyk Excellent job!
Let's make it brilliant and shiny.

You are demonstarting good knowledge of many OOP principles, es6 classes syntax and even private props syntax.

There is however a tiny flaw in hierarchy architecture we need to fix.
It also would be great to see method overloading yet at the cost of creativity with non-human inhabitants intro. However the latter will pay back with some bonus knowledge. (code-word `toString`)

submissions/ustymchyk/js-post-oop/index.js
```diff
@@ -35,12 +39,12 @@ class Pet extends Creature {
     #angrySound;
     #energy;
 
-    constructor(name, gender, age, happySound, angrySound) {
-        if (new.target === Pet) {
+    constructor(name, gender, age, species, happySound, angrySound) {
+        if (new.target === Creature) {
```
OleksiyRudenko: I can remember you fixed this to have a comparison against `Pet`. It didn't work as expected?

submissions/ustymchyk/js-post-oop/index.js
```diff
@@ -1,23 +1,27 @@
...
         }
 
+        this.species = species;
         this.name = name;
         this.gender = gender;
         this.age = age;
     }
 
+    toString() {
+        return `${this.species}, ${this.name}, ${this.gender}, ${this.age}`;
```
OleksiyRudenko, line 22: Imagine, you have dozens of props to form a string from.
How about emploing `Array.join`?

submissions/ustymchyk/js-post-oop/index.js
```diff
@@ -118,64 +120,101 @@ class Human extends Creature {
...
     }
 
-    introduce() {
+    toString() {
         const intro = `Hello! My name is ${this.name}, I'm ${this.age}. `;
-        let friends = '';
-        let spouse = '';
+        let friends = this.#friends.length ? `There is my friends: ${this.getFriendsNames()}. ` : '';
+        let spouse = this.spouse ? `And i have a spouse ${this.spouse.name}; ` : '';
+        let children = this.children.length ? `My ` + this.children.length > 1 ? `children: ${this._getChildren()}` : `child: ${this._getChildren()}` : '';
```
OleksiyRudenko, line 143: Nested ternaries are easier to read when formatting reflects nesting levels.
Consider employing [prettier.io](https://prettier.io/) plugin to your IDE for it to take care of formatting and readability.
**This is not required to fix now.**

submissions/ustymchyk/js-post-oop/index.js
```diff
@@ -203,10 +241,9 @@ const puppy = ruby.bornCreature(asher, 'Sushi', 'Sid');
...
    print('<strong>ABC</strong>');
    print('<strong>ABC</strong>', 'div');
-
    print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
    */
 
-[...men, ...women, luna, leo, kitty, asher, ruby, puppy].forEach(creature => print(creature.introduce()));
+[...men, ...women, luna, leo, kitty, asher, ruby, puppy].forEach(creature => print(creature.toString()));
```
OleksiyRudenko, line 249: Here is the bonus:
`.toString` is a magic method defined for every object in JS. It called implicitly in string context. It is `Object.toString` that is responsible for `[Object object]` output in console under node or via `alert` in browser.
`print` function ensures string context for its arguments, so `print(creature)` will work as expected.

submissions/ustymchyk/js-post-oop/index.js
```diff
@@ -118,64 +120,101 @@ class Human extends Creature {
...
+    }
+}
+
 const getRandomAge = () => Math.floor(Math.random() * 15 + 20);
+const names = new Names();
 
-const men = ['Liam', 'Noah', 'Oliver', 'William', 'Elijah', 'James', 'Benjamin'].map(name => new Human(name, 'm', getRandomAge()));
-const women = ['Olivia', 'Emma', 'Ava', 'Sophia', 'Isabella', 'Charlotte', 'Amelia'].map(name => new Human(name, 'f', getRandomAge()));
+const men = [0, 1, 2, 3, 4].map(() => new Human(names.getMaleName(), 'm', getRandomAge()));
+const women = [0, 1, 2, 3, 4].map(() => new Human(names.getFemaleName(), 'f', getRandomAge()));
```
OleksiyRudenko, line 211: Not a really valid use case for `.map`. We do not transform source array into new array, but rather use it as a quesi-iterator. What if we create dozens if not thousands of men and women? This approach won't scale up.
**You are not required to fix this now.**

OleksiyRudenko (changes requested):
@ustymchyk well done!
A few more tiny things and we are done.

submissions/ustymchyk/js-post-oop/index.js
ustymchyk: Totally agree. Just I`m enjoying how it looks in one line.  😂

submissions/ustymchyk/js-post-oop/index.js
ustymchyk: Do i understand correctly? Just Array.join([this.species, this.name, this.gender, this.age]).join(', ');

submissions/ustymchyk/js-post-oop/index.js
ustymchyk: No, I`ve fixed it in the 'submission' repo and forgot in my repo. Then, when made improvements in my repo, forgot about this point. So, it just a mistake during working with two repo.

submissions/ustymchyk/js-post-oop/index.js
ustymchyk: Or its about `Object.values(this).join(', ')`? 
But then i`ll have methods code.

OleksiyRudenko: `[this.species, this.name, this.gender, this.age].join(',  ');` as per changes is totally fine.

OleksiyRudenko (approved):
Brilliant work!

ustymchyk:
Thanks a lot)

PR #192, opened 2020-11-29
Oop exercise

webdevtoday:
### OOP Exercise

[Demo](https://webdevtoday.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/webdevtoday/a-tiny-JS-world/tree/gh-pages)

Please, review.


OleksiyRudenko (changes requested):
@webdevtoday nice job!
However we need to change some basic things before we can work productively.
Please
(a) switch over to ES6 classes syntax
(b) do not pass objects as arguments to constructor as this shadows fundamental flaws in the OOP code
(c) make sure OOP, SOLID and DRY principles are followed

As the least many mammals do not have hands at all, which is much different from having zero hands. Make your world more reality-like.

webdevtoday:
@OleksiyRudenko Thank you very much! I'll change those things.

webdevtoday:
Hi! I changed it with class syntax.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+   constructor(name, gender, friends) {
+      this.name = name;
+      this.gender = gender;
+      this.friends = friends;
+   }
+}
+
+class Human extends Mammal {
+   legs = 2;
```
OleksiyRudenko, line 22: `legs` prop already defined by a superclass. It's an antipattern to redefine an inherited property. It should be handled by property owning class.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,76 @@
...
+      super(name, gender, friends);
+      this.saying = saying;
+   }
+}
+
+class Animal extends Mammal {
+   constructor(name, gender, saying, species, friends) {
+      super(name, gender, friends);
+      this.saying = saying;
+      this.species = species;
```
OleksiyRudenko, line 36: Looks like `species` is a shared property. Thus should be handled by a super class.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
+
+const printList = [dog, cat, man, woman, catwoman];
+
+function tmplStr({species, name, gender, legs, hands, saying, friends}) {
```
OleksiyRudenko, line 68: Introduction of classes makes it reasonable that classes handle and use their properties on their own.
Make this function a method of class(es).
Benefit from inheritance, method overloading and `super.methodName` calls.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,76 @@
...
+   return `${species};${name};${gender};${legs};${hands};${saying};${friends.join(', ')}`;
```
OleksiyRudenko, line 69: There is a patern here worth employing `join`.

OleksiyRudenko (changes requested):
@webdevtoday nicely done.
You moved away from objects as literals but didn't arrive at OOP, SOLID and DRY yet.
Let's use not only es6 classes syntax but also OOP and SOLID principles.
[These materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) are to your rescue.

submissions/webdevtoday/oop_exercise/index.js
webdevtoday: I don't get you(( Do I need to create getter and setter for that property?

submissions/webdevtoday/oop_exercise/index.js
webdevtoday: super(name, gender, friends, species);  ?

webdevtoday:
@OleksiyRudenko I've changed my code. I reread articles about OOP and I understood where were my mistakes. I didn't use encapsulation and used class's properties directly in my code without public methods. I made two methods getProperties() and [Symbol.toString]. [Symbol.toString] takes result from getProperties() and then makes string.

webdevtoday:
@OleksiyRudenko But I don't understand some things. I have Animal class absolute empty. Do I need to delete it and make animals  use Mammal class?

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+      this.legs = 2;
+      this.hands = 2;
+   }
+
+   getProperties() {
+      return [this.species, this.name, this.gender, this.legs, this.hands, this.friends.join(', ')];
+   }
+}
+
+class Animal extends Mammal {}
```
OleksiyRudenko, line 42: We do not really need this class. It does nothing indeed.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+   constructor(name, gender, saying, species, friends) {
+      this.name = name;
+      this.gender = gender;
+      this.species = species;
+      this.saying = saying;
+      this.friends = friends;
+      this.legs = 4;
+   }
+   getProperties() {
+      return [this.species, this.name, this.gender, this.legs, this.friends.join(', ')];
```
OleksiyRudenko, line 21: `saying` is a property of this class but never returned by `getProperties`

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+   }
+   [Symbol.toPrimitive](hint) {
+      if (hint == 'string') {  
+         return this.getProperties().join(';');
+      }
+   }
```
OleksiyRudenko, line 27: This method code can be as short as
```
toString() {
  return this.getProperties().join('; ');
}
```

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+class Human extends Mammal {
+   constructor(name, gender, saying, friends) {
+      super(name, gender, saying, 'human', friends);
+      this.legs = 2;
+      this.hands = 2;
```
OleksiyRudenko, line 34: Adding `hands` as an own property of a `Human` is absolutely required as it is a property unique to humans.

`legs` is a shared property. If didn't have any assumptions as to the default value for it in `Mammal` you would manage this in a better way just like `name`, `species` etc.

Ask yourself, why 4 is a default value for `legs`? Are there any good reasons to have a default value for it in `Mammal`? Other properties do not deserve to have default values in `Mammal` so what's different with `legs`?

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+   }
+
+   getProperties() {
+      return [this.species, this.name, this.gender, this.legs, this.hands, this.friends.join(', ')];
+   }
```
OleksiyRudenko, line 39: Benefit from a method overload technique.

A child class `methodA` can call an inherited (predecessor's) method via `super.methodA` and supplement it.

This way both parent and child manage only those properties they own directly.

So, get props from the parent by calling `super.getProps`, supplement the list with props unique to a human and return combined array.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+class Animal extends Mammal {}
+
+const dog = new Animal('Jhonny', 'male', 'woof-woof!', 'dog', ['Andrew', 'Anastasia']);
+const cat = new Animal('Murka', 'female', 'meow!', 'cat', ['Anastasia']);
+const woman = new Human('Anastasia', 'female', 'I\'m a woman!', ['Andrew', 'Murka', 'Jhonny']);
+const man = new Human('Andrew', 'male', 'I\'m a man!', ['Anastasia', 'Jhonny', 'Murka']);
+const catwoman = new Human('CatWoman', 'female', cat.saying, []);
```
OleksiyRudenko, line 48: `Animal` (or `Mammal` when you get rid of an "empty" `Animal`) is a generic class that requires a species being explicitly specified, while `Human` is more specific and implies a particular species. A reader is able to catch a human's species right from its name.

How about having `Cat` and `Dog` classes? This way an inhabitant will be created from the class where name is a species name. This way code would look more consistent.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
+
+const printList = [dog, cat, man, woman, catwoman];
```
OleksiyRudenko, line 70: I'd rather say that here we define this would's inhabitants rather than a list to print.
Current naming implies that we want to have them "printed" only.
A variable name should reflect the nature of what is stored in that variable.

You could combine names, e.g. this are inhabitants we want to print. In either way they are inhabitants in the first hand.

OleksiyRudenko (changes requested):
@webdevtoday great job!
Some of your questions will be addressed in the comments below.
A few improvements and we're done.

submissions/webdevtoday/oop_exercise/index.js
webdevtoday: > Benefit from a method overload technique.
> 
> A child class `methodA` can call an inherited (predecessor's) method via `super.methodA` and supplement it.
> 
> This way both parent and child manage only those properties they own directly.
> 
> So, get props from the parent by calling `super.getProps`, supplement the list with props unique to a human and return combined array.

Yes, I think about it, but I found info about it) I knew only super()

webdevtoday:
@OleksiyRudenko Thank you very much! Your comments are amazing info. But I have one more problem. I changed CatWoman class, but it has relations with Human and Cat and I don't know how to do that. I decided that class extends Human class and overrides species prop and gets cat.saying as argument. But I violate encapsulation principle of OOP by doing so. I'm confused about this.

webdevtoday:
@OleksiyRudenko I decided to add a new method to Inhabitant class to don't use properties directly and save encapsulation.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+class Human extends Inhabitant {
+   constructor(name, gender, saying, friends) {
+      super(name, gender, saying, 'human', 2, friends);
+      this.hands = 2;
+   }
+
+   getProperties() {
+      const props = super.getProperties();
+      props.splice(3, 0, this.hands);
```
OleksiyRudenko, line 39: This approach relies on a particular order of props returned by the `super.getProperties`.
In reality a peer developer may change super class implementation legally taking no care about child classes (this is one of the points of OOP), so that this fragment may result in unexpected outcome.
Beware of such risks and design your code so that each class doesn't depend on particular implementations in neither parent nor child classes.

submissions/webdevtoday/oop_exercise/index.js
```diff
@@ -0,0 +1,99 @@
...
+      super(name, 'female', saying, friends);
+      this.species = 'CatWoman';
+   }
+}
+
+const dog = new Dog('Jhonny', 'male', ['Andrew', 'Anastasia']);
+const cat = new Cat('Murka', 'female', ['Anastasia']);
+const woman = new Human('Anastasia', 'female', 'I\'m a woman!', ['Andrew', 'Murka', 'Jhonny']);
+const man = new Human('Andrew', 'male', 'I\'m a man!', ['Anastasia', 'Jhonny', 'Murka']);
+const catwoman = new CatWoman('CatWoman', cat.getSaying(), []);
```
OleksiyRudenko, line 67: As we use a saying from a particular cat, our CatWoman is a breed with a particular cat rather than a species on her own.
Alternatively, `Cat` class could have a static property `defaultSaying` (or static method `getDefaultSaying`) that both `Cat` and `CatWoman` classes could use at construction time by referring to `Cat.defaultSaying` or calling `Cat.getDefaultSaying()`.

In fact, composition in es6 classes syntax is a real challenge and cannot be implemented nicely. This is the major learning on composition in this task.

OleksiyRudenko (approved):
@webdevtoday great job!
Check the comments below for your information.
Feel free asking questions on those.

submissions/webdevtoday/oop_exercise/index.js
webdevtoday: > This approach relies on a particular order of props returned by the `super.getProperties`.
> In reality a peer developer may change super class implementation legally taking no care about child classes (this is one of the points of OOP), so that this fragment may result in unexpected outcome.
> Beware of such risks and design your code so that each class doesn't depend on particular implementations in neither parent nor child classes.

That's true, I'll think about it

submissions/webdevtoday/oop_exercise/index.js
webdevtoday: > As we use a saying from a particular cat, our CatWoman is a breed with a particular cat rather than a species on her own.
> Alternatively, `Cat` class could have a static property `defaultSaying` (or static method `getDefaultSaying`) that both `Cat` and `CatWoman` classes could use at construction time by referring to `Cat.defaultSaying` or calling `Cat.getDefaultSaying()`.
> 
> In fact, composition in es6 classes syntax is a real challenge and cannot be implemented nicely. This is the major learning on composition in this task.

I wanted to use static method, but I didn't sure

webdevtoday:
@OleksiyRudenko Thank you very much for your help! I know I haven't understand OOP yet but I'm nearly to it!

submissions/webdevtoday/oop_exercise/index.js
OleksiyRudenko: No worries. This requirement was intentionally introduced to challenge the students. You took the challenge and solved it.

Your solution, whether it is a borrowing of a particular cat property or Cat class static property, is good enough, given the restrictions that classic OOP approach poses against us.

In fact, if we would employ composition design pattern to build a cat-woman this would require us to have the entire world, all of its inhabitants built by the means of composition.

You most likely will deal with classes+inheritance on your job. Now you know OOP better and will be able to handle pure object-oriented code.

But whenever you have an opportunity to design the entities from the scratch, consider the composition pattern first.

PR #201, opened 2020-12-02
OOP

natashafir:
### Frogger game

[Demo](https://natashafir.github.io/frogger-game/) |
[Code base](https://github.com/natashafir/frogger-game/blob/main/js/app.js)

Please, review.


OleksiyRudenko (changes requested):
@natashafir nice job!

It seems like bugs at their first appearance storm through the scene like jets! Not sure if it was intentional.

There are few flaws in the code to fix.

A checklist to your rescue.

 * Requirements re **OOP**:
 * Most common mistakes

natashafir:
@OleksiyRudenko thank you a lot for the review, I will make an effort to improve my code.

submissions/natashafir/oop/app.js
```diff
@@ -0,0 +1,113 @@
...
+            player.x = INITIAL_POSITION_X;
+            player.y = INITIAL_POSITION_Y;
+        }, 200)
+    }
+};
+
+let bugLocation = [60, 145, 230];
+
+bugLocation.forEach(function (bugY) {
+    let enemy = new Enemy(BUGS_START, bugY, 200);
```
OleksiyRudenko, line 94: There must be a semantic meaning behind mysterious number `200`. Tell a reader what it is without browsing the codebase.

submissions/natashafir/oop/app.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+    if (this.y < 0) {
+        setTimeout(function () {
+            player.x = INITIAL_POSITION_X;
+            player.y = INITIAL_POSITION_Y;
+        }, 200)
+    }
+};
+
+let bugLocation = [60, 145, 230];
```
OleksiyRudenko, line 91: Please demistify these numbers.
Can these be calculated from some basic values? Are these some pixel positions calculable from the game scene row number plus or minus certain offset?

OleksiyRudenko (changes requested):
@natashafir a really good job!
A few fixes and we're done

submissions/natashafir/oop/app.js
```diff
@@ -0,0 +1,113 @@
...
+// You should multiply any movement by the dt parameter
+// which will ensure the game runs at the same speed for
+// all computers.
+Enemy.prototype.update = function (dt) {
+    this.x += this.speed * dt;
+    if (this.x > BUGS_END) {
+        this.x = BUGS_START;
+        this.speed = (Math.floor(Math.random() * MIN_SPEED) + MAX_SPEED) * dt;
+    }
+    if (this.Collision()) player.resetPosition();
```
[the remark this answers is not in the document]
natashafir: should be ```this.player.resetPosition()```?

OleksiyRudenko (approved):
@natashafir great job!

PR #212, opened 2020-12-04
Frogger Game is ready for review

m-ruslan:
### Frogger Game

[Demo](https://m-ruslan.github.io/jsoop-practice/) |
[Code base](https://github.com/m-ruslan/jsoop-practice)

Please, review.


submissions/m-ruslan/js-oop-practice/app.js
```diff
@@ -0,0 +1,146 @@
...
+var pDefualtYPos = rowStep * 5 - 30;
+
+var player = new Player(pDefualtXPos, pDefualtYPos);
+
+var enemy1 = new Enemy(-colStep, rowStep * 3 - 30, 100);
+var enemy2 = new Enemy(-colStep, rowStep * 2 - 30, 200);
+var enemy3 = new Enemy(-colStep, rowStep * 1 - 30, 300);
+var enemy4 = new Enemy(-1000 - colStep, rowStep * 3 - 30, 200);
+var enemy5 = new Enemy(-1000 - colStep, rowStep * 2 - 30, 250);
+var enemy6 = new Enemy(-1000 - colStep, rowStep * 1 - 30, 300);
```
OleksiyRudenko, line 118: Quite many repetitions. Can you devise an approach to benefit from calculable values, dependence on enemy index and eliminating repetitons?
Imagine we need to create dozens of enemies.

OleksiyRudenko (changes requested):
@m-ruslan well done!

Below is the checklist focused on the room for improvements in the code. Please also check comments below.

 * Requirements re **OOP**:

submissions/m-ruslan/js-oop-practice/app.js
m-ruslan: @OleksiyRudenko thank you
I'll refactor my code and will come back to you for one more review)

m-ruslan:
@OleksiyRudenko  it seems to me that I've fixed all issues you wrote about. 
Could you please review?

OleksiyRudenko (approved):
@m-ruslan great job!

PR #216, opened 2020-12-05
JS OOP Exercise

UA-girl:

### OOP exercise

[Demo](https://ua-girl.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/UA-girl/a-tiny-JS-world)

Please, review.


submissions/UA-girl/task_js_oop_exersise/index.js
```diff
@@ -0,0 +1,60 @@
...
+        message += `I have ${this.legs} legs and ${this.hands} hands. My friends are ${this.friends.join(', ')}.`;
+        return message
+    }
+}
+
+class Animal extends Creature {
+    constructor(species, name, age, gender, say, friends = []) {
+        super(species, name, age, gender, say);
+        this.legs = 4;
+        this.friends = friends.length > 0 ? friends : [];
```
OleksiyRudenko, line 42: Why shared properties like species, gender, name, age and say are generalized into a base class, while legs and friends are not?

submissions/UA-girl/task_js_oop_exersise/index.js
```diff
@@ -0,0 +1,60 @@
...
+class Creature {
+    constructor(species, name, age, gender, say) {
+        this.species = species;
+        this.name = name;
+        this.age = age;
+        this.gender = gender;
+        this.say = say;
+    }
+
+    getDescription() {
```
OleksiyRudenko, line 19: Check the magic [`toString` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString)

It is defined for every object in JS, it is overridable and called implictly in string context.
`print` ensures string context for its first argument.

OleksiyRudenko (changes requested):
@UA-girl great job!
Let's polish it.

submissions/UA-girl/task_js_oop_exersise/index.js
UA-girl: Fixed. Waiting for more advices )))

submissions/UA-girl/task_js_oop_exersise/index.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+class Person extends Creature {
+    constructor(species, name, age, gender, legs, say, friends) {
+        super(species, name, age, gender, legs, say, friends);
+        this.hands = 2;
+    }
+
+    toString() {
+        return super.toString() + ` I have ${this.legs} legs and ${this.hands} hands. My friends are ${this.friends.join(', ')}.`
```
OleksiyRudenko, line 33: Why shared properties like species, gender, name, age and say are handled in the base class `toString`, while legs and friends are not?

Note how similar this question to the one I had at the previous iteration of the review.

Please induce and deduce new information from explicit statements like this.

OleksiyRudenko (changes requested):
@UA-girl my hope was we'd be done ) But alas, need to complete the generalization.
There must be good reasons to handle properties outside the class that owns them.

submissions/UA-girl/task_js_oop_exersise/index.js
```diff
@@ -0,0 +1,47 @@
...
+        this.name = name;
+        this.age = age;
+        this.gender = gender;
+        this.say = say;
+        this.legs = legs
+        this.friends = friends.length > 0 ? friends : [];
+    }
+
+    toString() {
+        return `${this.say}! My name is ${this.name} and I am ${this.age} years old ${this.gender} ${this.species}.  I have ${this.legs} legs and ${this.hasOwnProperty('hands') ? this.hands : 'no'} hands. My friends are ${this.friends.join(', ')}.`
```
OleksiyRudenko, line 22: > There must be good reasons to handle properties outside the class that owns them

Class `Creature` doesn't own property `hands`. A base class should never be aware of child classes props. Let child classes handle their own props.

Imagine we enrich our world with other species, e.g. birds. Following current approach will make the base class unnecessary overwhelmed with "foreign" properties handling.

Employ method overriding where necessary.

A child class method may call base class method by means of `super.methodName()`

This will make your code cleaner.

submissions/UA-girl/task_js_oop_exersise/index.js
UA-girl: I've made this by design for my simplicity to form a message (Animals have no hand), but you right, I've found better way for this. In fact, method toString can be stored in the parent class only.

submissions/UA-girl/task_js_oop_exersise/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+class Creature {
+    constructor(species, name, age, gender, legs, say, friends = []) {
+        this.species = species;
+        this.name = name;
+        this.age = age;
+        this.gender = gender;
+        this.say = say;
+        this.legs = legs
+        this.friends = friends.length > 0 ? friends : [];
```
OleksiyRudenko, line 18: This situation is handled by default value in constructor arguments.
Perhaps we'd rather need to handle a situation when `friends` argument is not an array at all.

submissions/UA-girl/task_js_oop_exersise/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+const woman = new Person('human', 'Alice', 25, 'female', 2, 'Hello, dear!', ['Adam', 'Stephany', 'Iren']);
+const man = new Person('human', 'Adam', 35, 'male', 2, 'Hey!', ['Alice', 'Stephany', 'Jeorge', 'Jack']);
+const dog = new Animal('dog', 'Margo', 7, 'female', 4, 'Buf-buf!', ['Jack', 'Pashtet', 'Cesar']);
+const cat = new Animal('cat', 'Murzik', 4, 'male', 4, 'Meow-meow!', ['Margo', 'Murka']);
+
+// ======== OUTPUT ========
+
+const inhabitats = [man, woman, dog, cat];
+inhabitats.forEach(item => print(item.toString(), 'div'));
```
OleksiyRudenko, line 47: `print` ensures string context for its first argument. `print(item, ....... )` would do.

OleksiyRudenko (changes requested):
@UA-girl almost there

UA-girl:
Added another fixes. Please, look now. I hope now code is better.

OleksiyRudenko (approved):
@UA-girl well done with method overloading.

`super.toString().slice(0, -1);` relies on what method from the super class does. It's OK here but in general this approach is error prone. Imagine someone changes the super class behaviour slightly so that it doesn't end with a full stop character. The child class may then "eat" a significant letter or number.

Your code might look neater if the code style was followed more consistently. I mean semicolons usage.

Just install and use https://prettier.io/docs/en/editors.html and https://eslint.org/docs/user-guide/integrations to do the dirty formatting job for you.

UA-girl:
Thank you a lot for your attention! I'll try to do better.

PR #217, opened 2020-12-05
Add solution for task tiny-js-world

Ant-C-tech:
### tiny-js-world

[Demo](https://ant-c-tech.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Ant-C-tech/a-tiny-JS-world)

Please, review.


submissions/Ant-C-tech/tiny-js-world/index.js
```diff
@@ -0,0 +1,135 @@
...
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+for (const item of inhabitants) {
+   print(createGreetingContent(item), 'div')
+}
```
OleksiyRudenko, line 135: Employ `inhabitants.forEach`

submissions/Ant-C-tech/tiny-js-world/index.js
```diff
@@ -0,0 +1,135 @@
...
+            if (obj[key].length > 1) {
+               content += `I have friends: ${obj[key]}.`
+            } else {
+               content += `I have friend: ${obj[key]}.`
+            }
+
+            break;
+         default:
+            break;
+      }
```
OleksiyRudenko, line 108: Using conditions in `switch` `case` statemnets is antipattern and should be avoided by all means.
Consider the following:
```
const propPrefixes = {
  hands: 'I have % hands',
  legs: 'I have % legs',
};

const content = ['legs', 'hands']
  .map(propName => prefixes[propname].replace('%', obj[propName]))
  .join('. ');
```
You'd need to devise an approach to handle friends, but let be your challenge to take.

```

File: submissions/Ant-C-tech/tiny-js-world/index.js [hunk 3]
```diff
@@ -0,0 +1,135 @@
...
+         break;
+      case obj.species == 'cat':
+         interlocutor = 'a dog'
+         break;
+      case obj.species == 'human' && obj.gender == 'male':
+         interlocutor = 'a biker'
+         break;
+      case obj.species == 'human' && obj.gender == 'female':
+         interlocutor = 'my husband'
+         break;
```
OleksiyRudenko, line 66: If we have many cats, or dogs, men and women, they all have similar interlocutors.
Not really realistic and even if so there are no reasons to assign interlocutor just like legs or hands, or species at object definition stage.

`switch (true) { case condition: ... }` is an antipattern that should be avoided.

OleksiyRudenko (changes requested):
@Ant-C-tech nicely done but requires quite an improvement.

Ant-C-tech:
Thank you for your review! It is very important for me. I have tried to make my code better. Please check my solution. Regards.

submissions/Ant-C-tech/tiny-js-world/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+   Object.keys(obj).forEach((key) => {
+      if (Array.isArray(obj[key])) {
+         for (const item of obj[key]) {
+            content += propPrefixes[key].replace('%', item)
+         }
+      } else if (obj[key] && propPrefixes[key]) {
+         content += propPrefixes[key].replace('%', obj[key])
+      }
+   })
```
OleksiyRudenko, line 71: Order of keys in an object is not guaranteed. Whenever a particualr order is required avoid `Object.values|keys|entries` methods as well as `for ... in` or `for ... of`.
List propertis to publish explicitly. This way you will also have an opportunity to skip any properties.

When processing arrays string concatenation is an anti-pattern. Employ `Array.map` and `Array.join`. This way you will get rid of unnecessary `content` variable and `.trim()`.

OleksiyRudenko (changes requested):
@Ant-C-tech almost done.
Please check the comments below.
The code style is inconsistent. Some of the statements end with a semi-colot, others do not.
Consider employing prettier and es-lint to improve code formatting. These tools offer relevant warnings and automatic code style fixing.
Please take care of your code readers.

submissions/Ant-C-tech/tiny-js-world/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+  const greeting = Object.keys(propPrefixes).map((key) => {
+      if (Array.isArray(obj[key])) {
+         for (const item of obj[key]) {
+            return propPrefixes[key].replace('%', item);
+         };
+      } else if (obj[key] && propPrefixes[key]) {
+         return propPrefixes[key].replace('%', obj[key]);
+      };
+   });
```
OleksiyRudenko, line 69: > Order of keys in an object is not guaranteed. Whenever a particualr order is required avoid Object.values|keys|entries methods as well as for ... in or for ... of.
List propertis to publish explicitly.

Please feel free asking a question if the above is not really clear.

Now that you have templates did you try treating `friends` just like any other property.

Ant-C-tech:
Thank you for your review and explanations! I hope I understood you correct and changed my code properly. Excuse me for my mistakes and necessity to check twice my work.

submissions/Ant-C-tech/tiny-js-world/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+   const propPrefixes = {
+      species: 'I am a %. ',
+      name: 'My name is %. ',
+      gender: 'My gender is %. ',
+      hands: 'I have % hands. ',
+      legs: 'I have % legs. ',
+      saying: 'I want to say you "%". ',
+      friends: 'I have friend %. ',
+   };
```
OleksiyRudenko, line 59: Please avoid copy-pasting code examples. They might be not good enough and often intentinally are not good enough to give a student freedom to elaborate on this rather than just copy-paste.

In particular here variable name is bad. These are not prefixes but rather... erm... something else.

submissions/Ant-C-tech/tiny-js-world/index.js
```diff
@@ -0,0 +1,97 @@
...
+      if (Array.isArray(obj[key])) {
+         for (const item of obj[key]) {
+            return propPrefixes[key].replace('%', item);
+         };
+      } else if (obj[key] && propPrefixes[key]) {
+         return propPrefixes[key].replace('%', obj[key]);
+      };
+   });
+
+   return greeting.join('');
```
OleksiyRudenko, line 71: There is a pattern in your props templates that can be generalized as a more appropriate parameter for `join`.

OleksiyRudenko (changes requested):
@Ant-C-tech almost there.
Some of the review comments seem to be ignored.
Please feel free asking questions if anything is not clear.

submissions/Ant-C-tech/tiny-js-world/index.js
Ant-C-tech: Thank you for your review and for the opportunity to ask questions! I think you mean something like this:
const greeting = Object.keys(propPrefixes).map((key) => {
      if (obj[key] && propPrefixes[key]) {
         return propPrefixes[key].replace('%', obj[key]);
      };
   });
But now, I still need one condition to show correctly how many friends the character has: "1 friend" or  "many friendS". Please, help me  resolve this problem properly.

submissions/Ant-C-tech/tiny-js-world/index.js
Ant-C-tech: Thank you for your review and for the opportunity to ask questions! What do you think, can I name this variable "template"?

submissions/Ant-C-tech/tiny-js-world/index.js
Ant-C-tech: Thank you for your review! I tried to do this way, but now when I treating "friends" just like any other property I have a problem with plural of word "friend". I think I need one more condition, but I am not sure you would think it as a good idea.

Ant-C-tech:
Thank you for your review! I noticed that my array has "undefined" elements, so I added method filter() to resolve this problem.

submissions/Ant-C-tech/tiny-js-world/index.js
Ant-C-tech: Thank you for your review! It's my mistake. What do you think, can I name the variable "template"?

submissions/Ant-C-tech/tiny-js-world/index.js
Ant-C-tech: Thank you for your review! I have tried to resolve this issue.

OleksiyRudenko (approved):
@Ant-C-tech great job!
`createGreetingContent` looks a bit overcomplicated, but it is OK for now.
You will be back to your Tiny JS World after a while and will have an opportunity to overcome this type of complexity thanks to OOP and SOLID principles.
Approved

Ant-C-tech:
Thank you for your review and approving.

PR #221, opened 2020-12-05
Add JS DOM task 

kt-std:
### JS DOM

[Demo](https://kt-std.github.io/js-dom/) |
[Code base](https://github.com/kt-std/js-dom)

Please, review.


submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+        heading: 'Mars',
+        paragraph: 'The fourth planet from the Sun and the second-smallest planet in the Solar System, being larger than only Mercury. Mars carries the name of the Roman god of war and is often referred to as the "Red Planet". The days and seasons are comparable to those of Earth, because the rotational period as well as the tilt of the rotational axis relative to the ecliptic plane are similar.',
+        period: '687 Earth days',
+        position: '04'
+    }
+];
+
+
+document.querySelector("#menu-btn").addEventListener('click', function display(e) {
+    document.querySelector("#menu").classList.toggle("visible");
```
MaxKalashnyk, line 32: You can move `document.querySelector("#menu")` to the global variables. It helps you not to query this element on each click of the "#menu-btn" element

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+});
+
+document.querySelector("#menu").addEventListener('click', function display(e) {
+    let sectionName = e.target.innerHTML.toLowerCase(),
+        target = e.target;
+    index = CONTENT.findIndex(el => el.heading.toLowerCase() == sectionName);
```
MaxKalashnyk, line 38: It's better to use `===` the strict comparison

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+    setData(index);
+});
+
+let addAnimation = (element, ...animationOptions) => element.animate(...animationOptions);
```
MaxKalashnyk, line 42: you can use const for all of your functions

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+        opacity: 1
+    }], {
+        duration: 500
+    });
+}
+
+let setInnerData = (element, data) => element.innerHTML = data;
+
+let setData = (index) => {
+    let image = document.getElementById("tabImage"),
```
MaxKalashnyk, line 58: you can use const here

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+    addAnimation(container, [{
+        opacity: 0
+    }, {
+        opacity: 1
+    }], {
+        duration: (600)
+    });
+
+    setInnerData(period, CONTENT[index].period);
+    if (window.innerWidth > 800) {
```
MaxKalashnyk, line 79: It's better to move `800` to constants

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+        updateBackground(headerMenu, ASSETS_PATH, `aside_${(CONTENT[index].heading).toLowerCase()}`)
+    };
+}
+
+window.addEventListener('resize', function(e) {
+    let image = document.querySelector(".header");
```
MaxKalashnyk, line 85: use const here.

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+    if (window.innerWidth <= 800) {
+        image.style.backgroundImage = 'none';
+    } else {
+        updateBackground(image, ASSETS_PATH, `aside_${(document.querySelector(".tab__header").textContent).toLowerCase()}`);
+    }
+});
+
+let renderContent = (data) => {
+    let menu = document.getElementById('menu'),
```
MaxKalashnyk, line 94: use const here.

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+        listItemClasses = ['menu__item', 'flex', 'center', 'full'];
+    fragment = new DocumentFragment();
+
+    data.forEach(element => {
+        let listItem = document.createElement('li');
+        listItem.classList.add(...listItemClasses);
+        listItem.textContent = `${element.heading}`;
+        listItem.setAttribute('data-content', `— ${element.position} —`);
+        listItem.setAttribute('tabindex', `1`);
```
MaxKalashnyk, line 103: Do you really need the `tabindex` here?

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+});
+
+let renderContent = (data) => {
+    let menu = document.getElementById('menu'),
+        listItemClasses = ['menu__item', 'flex', 'center', 'full'];
+    fragment = new DocumentFragment();
+
+    data.forEach(element => {
+        let listItem = document.createElement('li');
```
MaxKalashnyk, line 99: You can use const here because you don't reassign this value. You just change DOM node object properties.
More info [here](https://www.geeksforgeeks.org/javascript-const/) (example 6)

MaxKalashnyk (changes requested):
Nice work! look at my comments below

submissions/kt-std/js-dom/js/script.js
kt-std: Thank you for reviewing! 
I really hope, that I rightly fixed this issue. I've also implemented all your comments about other code issues (created constants for some variables and changed `let` to `const`).

submissions/kt-std/js-dom/js/script.js
kt-std: Replaced this one with the strict comparison

submissions/kt-std/js-dom/js/script.js
kt-std: Created `const` in the top section of code for that value

submissions/kt-std/js-dom/js/script.js
kt-std: It was used only for highlighting the menu tabs while navigating by pressing the Tab key. I've already removed it as it is not a mandatory stuff.

submissions/kt-std/js-dom/js/script.js
kt-std: Thank you for providing useful info:)

submissions/kt-std/js-dom/index.html
```diff
@@ -0,0 +1,41 @@
...
+	<link rel="preconnect" href="https://fonts.gstatic.com">
+	<link rel="stylesheet" media="screen and (min-width:300px) and (max-width: 800px)" href="css/mobile.css">
+	<link href="https://fonts.googleapis.com/css2?family=Playfair+Display&family=Roboto:wght@300;600&display=swap" rel="stylesheet"> 
+	<title>JS DOM</title>
+</head>
+<body class="flex">
+	<header class="header flex center">
+		<nav class="nav flex column full center">
+			<div class="menu__icon-container flex full">
+				<img class="menu__icon" id="menu-btn" src="assets/menu.svg" alt="Menu button">
```
MaxKalashnyk, line 16: Don't forget about the semantic elements in your layout or accessibility attributes.
I'd recommend using the `button` tag here or the `role` attribute on this element

submissions/kt-std/js-dom/index.html
```diff
@@ -0,0 +1,41 @@
...
+				<h1 class="tab__header"></h1>
+				<p class="tab__paragraph"></p>
+			</div>
+			<div class="tab__statistics flex">
+				<h2 class="statistics__header">Orbital period</h2>
+				<p class="statistics__text"></p>
+			</div>
+		</section>
+	</main>	
+	<script type="text/javascript" src="js/script.js"></script>
```
MaxKalashnyk, line 39: You can not use `type="text/javascript"`. Now it works fine without this attribute as well

submissions/kt-std/js-dom/js/script.js
```diff
@@ -0,0 +1,113 @@
+const ASSETS_PATH = "assets/";
```
MaxKalashnyk, line 1: Do you really still need this file?

submissions/kt-std/js-dom/index.html
MaxKalashnyk: Also, don't forget to change the exact script file path

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+    },
+];
+
+document
+    .querySelector('#menu-btn')
+    .addEventListener('click', function display(e) {
+        menu.classList.toggle('visible');
+    });
+
+document
```
MaxKalashnyk, line 41: You can use your new constant `menu` for this operation

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+        paragraph:
+            'The fourth planet from the Sun and the second-smallest planet in the Solar System, being larger than only Mercury. Mars carries the name of the Roman god of war and is often referred to as the \'Red Planet\'. The days and seasons are comparable to those of Earth, because the rotational period as well as the tilt of the rotational axis relative to the ecliptic plane are similar.',
+        period: '687 Earth days',
+        position: '04',
+    },
+];
+
+document
+    .querySelector('#menu-btn')
+    .addEventListener('click', function display(e) {
```
MaxKalashnyk, line 37: It's up to you of course, but you can use anonymous arrow function without arguments as well
```suggestion
    .addEventListener('click', () => menu.classList.toggle('visible'));
```

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+        menu.classList.toggle('visible');
+    });
+
+document
+    .querySelector('#menu')
+    .addEventListener('click', function display(e) {
+    const sectionName = e.target.innerHTML.toLowerCase();
+    index = CONTENT.findIndex((el) => el.heading.toLowerCase() === sectionName);
```
MaxKalashnyk, line 45: Don't forget about declaring variables. And it's better to rename it to `sectionIndex`, for example.
Because you can be confused when you have a lot of different indexes

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+                opacity: 1,
+            },
+        ],
+        {
+            duration: 600,
+        }
+    );
+
+    setInnerData(period, CONTENT[index].period);
+    if (window.innerWidth > 800) {
```
MaxKalashnyk, line 104: Please use the `SCREEN_WIDTH` variable here

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+        updateBackground(
+            headerMenu,
+            ASSETS_PATH,
+            `aside_${CONTENT[index].heading.toLowerCase()}`
+        );
+    }
+};
+
+window.addEventListener('resize', function(e) {
+    const image = document.querySelector('.header');
```
MaxKalashnyk, line 114: you can make this variable global in order not to query it on each resize of the window. It can help to increase the performance of your app

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+                .querySelector('.tab__header')
+                .textContent.toLowerCase()}`
+        );
+    }
+});
+
+const renderContent = (data) => {
+    const menu = document.getElementById('menu'),
+        listItemClasses = ['menu__item', 'flex', 'center', 'full'];
+    fragment = new DocumentFragment();
```
MaxKalashnyk, line 131: use const for this variable

submissions/kt-std/js-dom/script.js
```diff
@@ -0,0 +1,145 @@
...
+    {
+        heading: 'Mars',
+        paragraph:
+            'The fourth planet from the Sun and the second-smallest planet in the Solar System, being larger than only Mercury. Mars carries the name of the Roman god of war and is often referred to as the \'Red Planet\'. The days and seasons are comparable to those of Earth, because the rotational period as well as the tilt of the rotational axis relative to the ecliptic plane are similar.',
+        period: '687 Earth days',
+        position: '04',
+    },
+];
+
+document
```
MaxKalashnyk, line 35: you can move adding  this event listener to the bottom of your script when you declared `resize` event listener. it would be better to read your code for other develoeprs

MaxKalashnyk (changes requested):
Good job! Look at my comments below

submissions/kt-std/js-dom/index.html
kt-std: Thanks a lot!
I've added button tag to wrap the image

submissions/kt-std/js-dom/index.html
kt-std: I've deleted the `type` attribute. The file used in the project is stored in `js/script.js`. In that case, is there any need to change the path?

submissions/kt-std/js-dom/js/script.js
kt-std: The ASSETS_PATH variable is needed to set background images depending on the screen width. It sets the part of the path to where the images are stored. I guess there is no need to remove it because, in that case, it won't work / the path will have to be set explicitly, which is not a good idea, am I right?

submissions/kt-std/js-dom/script.js
kt-std: I've updated it with the constant

submissions/kt-std/js-dom/script.js
kt-std: I've changed it to an anonymous arrow function

submissions/kt-std/js-dom/script.js
kt-std: Sorry, missed that one previous time

submissions/kt-std/js-dom/script.js
kt-std: I've updated the variable name so that it became more clear what it actually stands for

submissions/kt-std/js-dom/script.js
kt-std: I've moved the event listener to the bottom of the script

submissions/kt-std/js-dom/script.js
kt-std: Constant is now used, therefore this variable is not global anymore

submissions/kt-std/js-dom/script.js
kt-std: I've changed it due to your advice

MaxKalashnyk (approved):
Good job! Approved!

PR #222, opened 2020-12-05
Saiko oop

RomanSaiko:
### oop exercise

[Demo](https://romansaiko.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/RomanSaiko/a-tiny-JS-world)

Please, review.


submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+        this.legs = legs
+        this.hands = hands
+        this.saying = saying
+        this.friends = friends
+    }
+}
+
+class Animal extends Inhabitant {
+    constructor ({ species, name, gender, saying, friends }) {
+        super({ species, name, gender, legs: 4, hands: 0, saying, friends })
```
OleksiyRudenko, line 15: Dogs and cats derive from animals in hierarchy we've got here.
Problem is that pets do not have hands at all by design.
The point of OOP and SOLID is that objects have only the properties they really need.
Imagine you will extend your world with birds, insects... Will every inhabitant have e.g. wings props?

submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const dog = new Dog({ name: 'Spike', gender: 'male', friends: ['Dolf', 'Roman', 'Mariia'] })
+const cat = new Cat({ name: 'Dolf', gender: 'male'})
+const man = new Man({ name: 'Roman', saying: 'Hello!', friends: ['Dolf', 'Roman', 'Mariia'] })
+const woman = new Woman({ name: 'Mariia', saying: 'Hi!', friends: ['Dolf', 'Roman', 'Mariia'] })
+
+const inhabitants = [dog, cat, man, woman]
+
+const introduceInhabitants = (inhabitants) => {
```
OleksiyRudenko, line 56: Make this function a method of class(es).
Benefit from inheritance, methods overloading and `super.methodName` calls.

submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+    const inhabitantValues = Object.values(inhabitants)
```
OleksiyRudenko, line 57: `Object.values` here implies that the order of props in an object is guaranteed. It is not.
Use explicit properties listing.
And remeber: every class may handle only props owned by itself.

OleksiyRudenko (changes requested):
@RomanSaiko well done, but not SOLID enough. Yet some of the OOP principles aren't followed and employed.
[These materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) are to your rescue.

submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,72 @@
...
+class Inhabitant {
+    constructor ({ species, name, gender, saying, friends = "No friends! One man army.", paws, hands, legs }) {
+        this.species = species
+        this.name = name
+        this.gender = gender
+        this.saying = saying
+        this.friends = friends
+        this.paws = paws
+        this.hands = hands
+        this.legs = legs
```
OleksiyRudenko, line 11: This makes all inhabitants to have paws AND hands AND legs.
Some of those propwerties are just ignore in some cases, but still all inhabitants have those.
Of the points of the OOP si that any predecessor has only properties that are commong for its ancestors, and every child (ancestor) class introduces properties unique to it (and shared by its ancestors if any).

submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    toString() {
+        return `${this.species}; ${this.name}; ${this.gender}; ${this.saying}; ${this.friends}`;
```
OleksiyRudenko, line 15: Nicely done.
However, there is pattern here.
Imagine you have dozens of props. It is easy to miss a space and/or a semicolon under current approach.
Also `this.whatever` looks a bit wordy.
How about listing property names, then mapping them against object values and finally joining them with a delimiter?

OleksiyRudenko (changes requested):
@RomanSaiko well improved code.
However we still need to resolve an issue with shared props vs unique props.
Please also make your code style consistent in terms of semicolons use.

submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+    constructor ({ species, name, gender, saying, friends = "No friends! One man army" }) {
+        this.species = species
+        this.name = name
+        this.gender = gender
+        this.saying = saying
+        this.friends = friends
+    }
+
+    toString() {
+        const props = Object.entries(this).map(([key, value]) => `${key}: ${value}` )
```
OleksiyRudenko, line 11: `Object` methods like `keys`, `values`, `entries` as well as `for...in` shouldn't be used as these do not guarantee any particular order of keys/values.
List properties explicitly.
Also remember that that parent class shouldn't know anything about its children props. Child classes must have good reasons to assign or refer parents' class props. There are no such reasons in this task.
Method overloading is to the rescue when building an inhabitant's presentation string.

submissions/RomanSaiko/oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+const dog = new Dog({ name: 'Spike', gender: 'male', friends: ['Dolf', 'Roman', 'Mariia'] })
+const cat = new Cat({ name: 'Dolf', gender: 'male'})
+const man = new Man({ name: 'Roman', saying: 'Hello!', friends: ['Spike', 'Dolf', 'Mariia'] })
+const woman = new Woman({ name: 'Mariia', saying: 'Hi!', friends: ['Spile', 'Dolf', 'Roman'] })
+
+const inhabitants = [dog, cat, man, woman]
+
+inhabitants.forEach((item) => {
+    print(item.toString())
```
OleksiyRudenko, line 63: `.toString` is called implictly in string context. `print` ensures string context, so you may skip calling `.toString` explicitly.

OleksiyRudenko (changes requested):
@RomanSaiko nearly there!
A couple of improvements and we are done.

submissions/RomanSaiko/oop/index.js
RomanSaiko: @OleksiyRudenko, but you said "this.whatever looks a bit wordy". 
Is it correct that I must implement toString() at all ending classes (dog, cat, man, woman)?

OleksiyRudenko: > @OleksiyRudenko, but you said "this.whatever looks a bit wordy".

Hint: `['propAname', 'propBname'].map(propName => this[propName])` -- `this` is referred only once.

> Is it correct that I must implement toString() at all ending classes (dog, cat, man, woman)?

Make benefit from method overloading:
- Each class has its own version of `toString` method to build a string from properties it **owns**
- Every child class inherits it
- If a child class has a unique property it can implement its own version of `toString` that will call parent's version of `toString` (by calling it as `super.toString()`) and supplement the result with values of child class' properties

OleksiyRudenko (approved):
@RomanSaiko great job!
Consider improving code style (use of semicolons, chained calls indentation). This adds to reading experience. Prettier and eslint are to do the "dirty job" for you.

PR #224, opened 2020-12-06
OOP-exercise (Tiny-es6-world)

Iakow:
### OOP-exercise

[Demo](https://iakow.github.io/my-tiny-js-world/) |
[Code base](https://github.com/Iakow/frontend-2021-homeworks/blob/OOP-exercise/submissions/iakow/OOP-exercise/index.js)

Please, review.

### DIFF
--- a/submissions/iakow/OOP-exercise/index.js
+++ b/submissions/iakow/OOP-exercise/index.js
@@ -7,8 +7,7 @@ class Inhabitant {
   }
 
   makeFriends(...applicants) {
-     applicants.filter(applicant => applicant instanceof Inhabitant)
-        .filter(applicant => applicant != this)
+     applicants.filter(applicant => applicant instanceof Inhabitant && applicant != this)
         .forEach(applicant => this.friends.add(applicant));
   }
 
--- a/submissions/iakow/OOP-exercise/index.js
+++ b/submissions/iakow/OOP-exercise/index.js
@@ -25,7 +25,6 @@ class Animal extends Inhabitant {
    constructor({ species, name, gender, say, paws = 4 }) {
       super({ species, name, say, gender });
       this.paws = paws;
-      this.gender = gender;
    }
 
    toString() {
@@ -56,7 +55,6 @@ class Human extends Inhabitant {
       super({ species, name, say, gender });
       this.legs = legs;
       this.hands = hands;
-      this.gender = gender;
    }
 
    getFriendsNamesList() {
--- a/submissions/iakow/OOP-exercise/index.js
+++ b/submissions/iakow/OOP-exercise/index.js
@@ -15,9 +15,9 @@ class Inhabitant {
    toString() {
       const { name, species, friends, gender, say } = this;
 
-      const friendNames = [];
-      friends.forEach((friend) => friendNames.push(friend.name));
-      const friendsNamesString = friendNames.length ? `${friendNames.join(', ')};` : '';
+      const friendsNames = (friends.size)
+         ? [...friends].map(friend => friend.name).join(', ')
+         : '';
 
       return `${species}; ${name}; ${gender}; ${say}; ${friendsNamesString}`;
    }
--- a/submissions/iakow/OOP-exercise/index.js
+++ b/submissions/iakow/OOP-exercise/index.js
@@ -16,7 +16,7 @@ class Inhabitant {
       const { name, species, friends, gender, say } = this;
 
       const friendsNames = (friends.size)
-         ? [...friends].map(friend => friend.name).join(', ')
+         ? [...friends].map(friend => friend.name).join(', ') + ';'
          : '';
 
       return [species, name, gender, say, friendsNames].join('; ')
--- a/submissions/iakow/OOP-exercise/index.js
+++ b/submissions/iakow/OOP-exercise/index.js
@@ -13,11 +13,13 @@ class Inhabitant {
    }
 
    toString() {
-      const { name, species, friends, gender } = this;
-      return (
-         `Data: ${species}, ${gender}, ${name}, ${friends.size || 'no'} friend` +
-         `${friends.size == 1 ? '' : 's'}`
-      )
+      const { name, species, friends, gender, say } = this;
+
+      const friendNames = [];
+      friends.forEach((friend) => friendNames.push(friend.name));
+      const friendsNamesString = friendNames.length ? `${friendNames.join(', ')};` : '';
+
+      return `${species}; ${name}; ${gender}; ${say}; ${friendsNamesString}`;
    }
 }
 
@@ -28,7 +30,7 @@ class Animal extends Inhabitant {
    }
 
    toString() {
-      return `<b>${this.say}</b> \n${super.toString()}, ${this.paws} paws.`
+      return `${super.toString()} ${this.paws};`
    }
 }
 
@@ -57,22 +59,8 @@ class Human extends Inhabitant {
       this.hands = hands;
    }
 
-   getFriendsNamesList() {
-      if (this.friends.size) {
-         return (
-            `I\'m friends with ${[...this.friends].map(friend => friend.name).join(', ')}`
-         );
-      } else {
-         return 'I don\'t need friends';
-      }
-   }
-
    toString() {
-      return (
-         `<b>${this.say}</b>\n` +
-         `${super.toString()}, ${this.gender}, ${this.hands} hands and ${this.legs} legs.\n` +
-         `<b>${this.getFriendsNamesList()}.</b >`
-      )
+      return `${super.toString()} ${this.hands}, ${this.legs};`;
    }
 }
 
--- a/submissions/iakow/OOP-exercise/index.js
+++ b/submissions/iakow/OOP-exercise/index.js
@@ -1,80 +1,81 @@
 class Inhabitant {
-  constructor({ species, name, say }) {
-     this.species = species;
-     this.name = name;
-     this.say = say;
-     this.friends = new Set();
-  }
+   constructor({ species, name, say, gender }) {
+      this.species = species;
+      this.name = name;
+      this.say = say;
+      this.gender = gender;
+      this.friends = new Set();
+   }
 
-  makeFriends(...applicants) {
-     applicants.filter(applicant => applicant instanceof Inhabitant && applicant != this)
-        .forEach(applicant => this.friends.add(applicant));
-  }
+   makeFriends(...applicants) {
+      applicants.filter(applicant => applicant instanceof Inhabitant && applicant != this)
+         .forEach(applicant => this.friends.add(applicant));
+   }
 
-  toString() {
-     const { name, species, friends } = this;
-     return (
-        `Data: ${species}, ${name}, ${friends.size || 'no'} friend` +
-        `${friends.size == 1 ? '' : 's'}`
-     )
-  }
+   toString() {
+      const { name, species, friends, gender } = this;
+      return (
+         `Data: ${species}, ${gender}, ${name}, ${friends.size || 'no'} friend` +
+         `${friends.size == 1 ? '' : 's'}`
+      )
+   }
 }
 
 class Animal extends Inhabitant {
-  constructor({ species, name, gender, say, paws = 4 }) {
-     super({ species, name, say });
-     this.paws = paws;
-     this.gender = gender;
-  }
+   constructor({ species, name, gender, say, paws = 4 }) {
+      super({ species, name, say, gender });
+      this.paws = paws;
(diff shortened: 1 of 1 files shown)


submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+  makeFriends(...args) {
+    const { friends } = this;
+
+    args.forEach((item) => {
+      if (item instanceof Inhabitant && item != this && !friends.includes(item.name)) {
+        friends.push(item.name);
+      }
+    });
+  }
```
OleksiyRudenko, line 20: Not a KO to fix. Fixing is at your discretion.

`args` is too generic name for the arguments expected. `friends` already in use, but we might distinguish new friends passed as arguments by more specific naming. Sematically speaking naming is important.

Test against `instanceof Inhabitant` can be implemented using `Array.filter`.
Same is true with regards to equality with `this`.

A one-liner to ensure only unique values/objects are in array is `list = [...new Set(list)]`. Check [docs on `Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), you will find it helpful.

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,87 @@
+class Inhabitant {
+  constructor({ species, name, gender, legs, hands, say }) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
```
OleksiyRudenko, line 7: Some of our inhabitants do not have hands by design. The beauty of OOP/SOLID is that every base class handles props shared by its child classes, and every child class introduces props unique to that child class.
This way we make scalable and maintainable code.

It is important that every class handles its properties. I.e. a base class should know nothing about props unique to its child classes and therefore may not refer to them.

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  toString() {
+    const { name, species, gender, legs, hands, say, friends } = this;
+
+    const aboutFriends = (friends.length) ? `I am friends with ${friends.join(', ')}.` : '';
+
+    return `${say} My name is ${name} end I'm ${species}, ${gender}. Obviously, I have ${legs} legs and ${hands} hands.
+${aboutFriends}`;
+  }
```
OleksiyRudenko, line 62: Complete overloading of a method is legal approach. Let's demonstrate the ability to expand (rather than replace) functionality of overlodaded methods. A child class should handle only props unique to it.

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+class CatWoman extends Human {
+  constructor({ name, say }) {
+    super({ name, gender: 'female', say });
+    this.species = 'humacat';
```
OleksiyRudenko, line 68: It is not a really good pattern to change props owned by a base class in a child class.
Hierarchies can be complex and base class may change the way it handles the property, e.g. it switches to employ `this.constructor.name` to refer to a species, ignoring `species` argument. Humans and pets will still work but this line in `CatWoman` may not reach the desired effect.

Possible solutions:
1. Employ composition pattern
2. Use a method to assign a species inherited from the top base class (which owns this property).

OleksiyRudenko (changes requested):
@Iakow a really good job.
However we need to employ OOP and SOLID principles while the code base is small. It will pay if we decide to populate our world with more inhabitants and introduce more species.

Iakow:
Hello!
Thanks for all the comments. I have benefited from all of them.
Hope I didn't create any new problems)

submissions/iakow/OOP-exercise/index.js
Iakow: I'm not sure if my new solution is correct?

Iakow: And I have a question about the composition pattern. Do it mean creating a mixin, and only that? I have no idea how it might look specifically in my code ... Where can I find examples?

AMashoshyna (commented):
Very nice improvement! I left one small comment to fix.

Concerning composition: here is my favorite video on composition: https://youtu.be/wfMtDGfHWpA
The basic idea is that you assemble an class from building blocks as opposed to extending existing classes gradually (as inheritance-based approach would suggest).
Composition is a very powerful tool, but if you don't feel like diving this deep, you can simply pass 'humacat' as an argument to `super`.

Iakow:
Thank you, Anastasiya!

Iakow:
And Oleksiy! ))

OleksiyRudenko (changes requested):
@Iakow well done!

A couple of fixes and we are done.

1. Gender
All inhabitants have gender, thus it should be generalized in their common predecessor class just like `species`, `name` and a few others.

2. Props usage
There must be good reason to refer inherited properties directly, like `this.say` in any of the child classes' `toString`. You may want to call `super.toString` from a child class' `toString` method.

Why that? Because parent class may change the way it stores its properties and this would require changes across all child classes (maybe dozens of them).

Item 1 is to be fixed.

Item 2 is either be fixed or explanations provided to ground the current approach.

Iakow:
@OleksiyRudenko
> 1. Gender
>    All inhabitants have gender, thus it should be generalized in their common predecessor class just like `species`, `name` and a few others.

I just assumed that not all inhabitants can have gender. Like a talking tree, for example. But I guess I read the task badly. I'll fix it now.

> 2. Props usage
>    There must be good reason to refer inherited properties directly, like `this.say` in any of the child classes' `toString`. You may want to call `super.toString` from a child class' `toString` method.
> 
> Why that? Because parent class may change the way it stores its properties and this would require changes across all child classes (maybe dozens of them).

I'm not sure, either I don't understand or see no problem in my given toy code. Since I have to pass the `say` to every constructor call, I control it in every instance. But I understand that if I fixed the `paws = 4` in `Animals`, and this would not be controlled in `Dog`, then I would not have to refer to `paws` in `Dog`.

OleksiyRudenko:
> I just assumed that not all inhabitants can have gender. Like a talking tree, for example. But I guess I read the task badly. I'll fix it now.

We do not have trees yet. If we wanted to introduce those then most likely we would introduce an extra parent class that would contain only props shared by current inhabitants and trees.

In that case we'd had to change anything in the current top-most class (generalizing shared properties to a new super-class) and remaining animals classes would remin intact.

Of course we'd add a `Tree` class.

The point here is any pre-mature optimization jst adds unnecessary complexity to the code.

OleksiyRudenko:
Alternatively, you might introduce an interim superclass, e.g. `GenderedInhabitants`. So, trees would inherit from `Inhabitant` and all others would inherit from `GenderedInhabitants`.

In either case a shared property should be a property of some class that it extended by child classes that have this shared property by design.

Iakow:
@OleksiyRudenko  thanks for the detailed answer!

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,97 @@
...
+         `${friends.size == 1 ? '' : 's'}`
+      )
+   }
+}
+
+class Animal extends Inhabitant {
+   constructor({ species, name, gender, say, paws = 4 }) {
+      super({ species, name, say, gender });
+      this.paws = paws;
+      this.gender = gender;
```
OleksiyRudenko, line 28: Why do we need re-initialize gender after it is being passed to the parent class constructor?
Check also other classes and fix.

OleksiyRudenko (changes requested):
@Iakow please double check your code, republish your app and make sure the output matches expectations and requirements.
Please also make all inhabitants that have friends have their friends represented in the output.

Iakow:
@OleksiyRudenko, done? )

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,83 @@
...
+   makeFriends(...applicants) {
+      applicants.filter(applicant => applicant instanceof Inhabitant && applicant != this)
+         .forEach(applicant => this.friends.add(applicant));
+   }
+
+   toString() {
+      const { name, species, friends, gender, say } = this;
+
+      const friendNames = [];
+      friends.forEach((friend) => friendNames.push(friend.name));
```
OleksiyRudenko, line 19: Here is an equivalent fragment:
```
const friendNames = friends.map(friend => friend.name);
```

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,83 @@
...
+      const friendsNamesString = friendNames.length ? `${friendNames.join(', ')};` : '';
```
OleksiyRudenko, line 20: Or even shorter alltogether
```
friendNamesString = friends.length
  ? friends.map(friend => friend.name).join(', ')
  : '';
```

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+      return `${species}; ${name}; ${gender}; ${say}; ${friendsNamesString}`;
```
OleksiyRudenko, line 22: There is a pattern here.
Consider the following example: `return [propA, propB, varC].join("; ");`
The latter is more scalable: you do not need to remember about `; `, no need to deal with extra-long string if number of properties grows.
Eliminate code repetition whenever possible to make this a habit.

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+class Human extends Inhabitant {
+   constructor({ species = "human", name, say, gender, hands = 2, legs = 2 }) {
+      super({ species, name, say, gender });
+      this.legs = legs;
+      this.hands = hands;
+   }
+
+   toString() {
+      return `${super.toString()} ${this.hands}, ${this.legs};`;
```
OleksiyRudenko, line 63: You may want here to join an array of properties for code style consistency and for the sake of scalability.

Not required to fix now for you'd need to resolve inconsistency with friends list that is always terminated with a semicolon.

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+class Animal extends Inhabitant {
+   constructor({ species, name, gender, say, paws = 4 }) {
+      super({ species, name, say, gender });
+      this.paws = paws;
+   }
+
+   toString() {
+      return `${super.toString()} ${this.paws};`
```
OleksiyRudenko, line 33: You may want here to have an array of values (e.g. `[super.toString(), this.paws]`) also joined for code style consistency and for the sake of scalability.

Not required to fix now for you'd need to resolve inconsistency with friends list that is always terminated with a semicolon.

OleksiyRudenko (changes requested):
@Iakow properties are now nicely resolved.
Please do not expect that every line of code affected with required changes will be marked by a reviewer, and deduce related changes on your own.

Let's polish the code and we're done.

submissions/iakow/OOP-exercise/index.js
Iakow: friends is Set, no Array

OleksiyRudenko: Makes sense. Then you can leave this as it is, or employ `Array.from`.

In either case let me know if you're gonna fix anything.

submissions/iakow/OOP-exercise/index.js
Iakow: I left it as it is, but the pattern took it into use, thanks.

submissions/iakow/OOP-exercise/index.js
Iakow: I left it as it is, but the pattern took it into use, thanks.

submissions/iakow/OOP-exercise/index.js
```diff
@@ -0,0 +1,83 @@
...
+   }
+
+   toString() {
+      const { name, species, friends, gender, say } = this;
+
+      const friendsNames = (friends.size)
+         ? [...friends].map(friend => friend.name).join(', ')
+         : '';
+
+      return [species, name, gender, say, friendsNames].join('; ')
```
Iakow, line 22: >Please do not expect that every line of code affected with required changes will be marked by a reviewer, and deduce related changes on your own.

This way?

Iakow: @OleksiyRudenko
I think done

OleksiyRudenko (approved):
@Iakow great job!

PR #226, opened 2020-12-06
Task a-tiny-js-world (oop es6 classes) added

BlueLamantine:
### Task a-tiny-js-world (post-oop es6 classes)

[Demo](https://bluelamantine.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/BlueLamantine/a-tiny-JS-world)

Please, review.


OleksiyRudenko (changes requested):
Please use named arguments in class constructors.
Current approach shadows issues with hierarchy and OOP principles.
Besides, code like
`constructor(props) {  this.name = props.name;   this.friends = ... }` makes class API unobvious and IDE intellisence helper (hints on functions signatures) useless.

Hints from intellisence help developers not to keep multiple functions API in mind.

BlueLamantine:
@OleksiyRudenko thanks for reviewe! I have made some changes to my code: named arguments in class constructors and ect. Also I used prettier to format my code.  But set up auto insert final new line at the end of file didn't work this time, I even added this line by hadns     ¯\__( ͡° ͜ʖ ͡°)__/¯

submissions/BlueLamantine/oop/index.js
```diff
@@ -0,0 +1,125 @@
...
+   constructor({ name, gender, friends }) {
+     super({ name, gender, friends, species: 'cat' });
+     this.sound = 'Meow Purrrr Purrr';
+   }
+ }
+ 
+ class Dog extends Pet {
+   constructor({ name, gender, friends }) {
+     super({ name, gender, friends, species: 'dog' });
+     this.sound = 'Woof woof';
```
OleksiyRudenko, line 88: Both cats and dogs have a shared property `sound`. 
Whenever two sibling classes have a similar property it should be generalized to a base class. To the top base class if needed (like `name`).

The point of having similar things generalized is that any handling of those happens in a single place.

Please check your entire codebase to follow this principle.

submissions/BlueLamantine/oop/index.js
```diff
@@ -0,0 +1,125 @@
...
+       `This ${this.species} is a ${this.gender}, ` +
+       `${this.getGender()[this.gender][1]} name is ${this.name}. ` +
+       `${this.getGender()[this.gender][0]} has ${this.legs} legs and ${
+         this.hands
+       } hands. ` +
+       `${this.getGender()[this.gender][0]} says \'${
+         this.voise
+       }\'. Friend with : ${this.friends}`
+     );
+   }
```
OleksiyRudenko, line 42: A class should handle the properties it owns immediately. There must be a good reason to handle a class' property in any of its children.

Use method overloading and `super.methodName` calls where required.

OleksiyRudenko (changes requested):
@BlueLamantine good work!
Check the [extra materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) to master OOP and SOLID principles.
Imagine you are required to rename `.sound`. How many fragments of code you would need to  fix? And if you had more classes for other species of pets?
The whole point of OOP and SOLID is that one can write a code that scales up easily and remains maintainable.

OleksiyRudenko:
> @OleksiyRudenko thanks for reviewe! I have made some changes to my code: named arguments in class constructors and ect. Also I used prettier to format my code. But set up auto insert final new line at the end of file didn't work this time, I even added this line by hadns ¯__( ͡° ͜ʖ ͡°)__/¯

Can it be that you work on a Mac and git settings or your IDE enforces a newline character to be a CR rather than LF (Linux) or CRLF (Windows)?

Consider setting your git and/or IDE configuration so that LF newline character (Linux style) is used at all times.

submissions/BlueLamantine/oop/index.js
BlueLamantine: Now it is clear for me, removed that and some other duplications !

submissions/BlueLamantine/oop/index.js
BlueLamantine: Ok

BlueLamantine:
> > @OleksiyRudenko thanks for reviewe! I have made some changes to my code: named arguments in class constructors and ect. Also I used prettier to format my code. But set up auto insert final new line at the end of file didn't work this time, I even added this line by hadns ¯__( ͡° ͜ʖ ͡°)__/¯
> 
> Can it be that you work on a Mac and git settings or your IDE enforces a newline character to be a CR rather than LF (Linux) or CRLF (Windows)?
> 
> Consider setting your git and/or IDE configuration so that LF newline character (Linux style) is used at all times.

I work on Windows.  That's weird, I push two identical files index.js to source repo and to this. Only in this repo with homeworks problem with new line at the end of file.

submissions/BlueLamantine/oop/index.js
```diff
@@ -0,0 +1,133 @@
...
+    super({ name, gender, friends, species: 'human' });
+    this.voise = voise;
+    this.legs = 2;
+    this.hands = 2;
+  }
+
+  getInfo() {
+    return (
+      super.getInfo() +
+      ` ${super.getGender()[this.gender][0]} has ${this.legs} legs and ${
```
OleksiyRudenko, line 42: By default class inherits parent's properties and methods.
We can refer these by `this.propName` or `this.methodName()` as if they were this class' own props and methods. We just need good reasoning for this, but in general it is legit.

So we can call `.getGender` as `this.getGender()`.

So, why do we call `.getInfo` as `super.getInfo()`? Because it is overloaded and we want call parent's method explicitly.

Take a look at the following fragment:
```
class A {
  constructor() {}
  getNestingLevel() { return 1; }
}

class B {
  constructor() { 
    super();
    this.getNestingLevel(); // 2
    super.getNestingLevel(); // 1
  }
  getNestingLevel() { return 2; }
  someMethod() {}
}

class C extends B {
  contsructor() {
    super();
    this.getNestingLevel(); // 2
    super.getNestingLevel(); // 2
  }
  someMethod() {
    super.someMethod(); // all good
    this.someMethod(); // current method call; leads to an infinite loop
  }
}
```

submissions/BlueLamantine/oop/index.js
```diff
@@ -0,0 +1,133 @@
...
+    this.gender = gender;
+    this.species = species;
+    this.friends = Array.isArray(friends) ? friends : 'no friends';
+  }
+  getGender() {
+    return {
+      male: ['He', 'his'],
+      female: ['She', 'her'],
+    };
+  }
```
OleksiyRudenko, line 19: Effectively this method returns a set of constants. We do not even need a method for this, having instead a static class property as these contants are all the same for all instances.

Alternatively we could make this method static. Then it would be called as `General.getGender()` denoting its general nature as it returns the same value for any instance of the class or its children.

**But let's make it useful as a regular non-static class method. And here is how:**

By the moment this method is called the gender is already assigned. Can you make this method returning only set of pronouns which would depend on current object's gender?
You may even retain the literal object here as a dictionary and just pick a proper pair of pronouns based on `this.gender`.

By the way, rename this method please. It doesn't return any genders at all. What is it that it actully returns?

If you can improve your code as per above, then here is the next challenge:
`super.getGender()[this.gender][0]` looks pretty cryptic.

What if we could tell the method what exactly pronoun case we need?
Just look how the following fragments read:
```
getPronoun('nominative') // returns 'He' or 'She' depending on gender
getPronoun('posessive') // returns 'His' or 'Her' depending on gender
```

OleksiyRudenko (changes requested):
@BlueLamantine well done!
There is still room for improvement. Let's make the code friendlier to read.

submissions/BlueLamantine/oop/index.js
BlueLamantine: Step by step I feel more confident with oop but still not strong in it

submissions/BlueLamantine/oop/index.js
BlueLamantine: Challenge accepted and accomplished!

OleksiyRudenko (approved):
@BlueLamantine excellent job!

BlueLamantine:
@OleksiyRudenko  thank you so much for review. You really help me as a personal coach with my 
weakness in oop  !

PR #228, opened 2020-12-06
Tiny JS World task

kt-std:
### Tiny JS World

[Demo](https://kt-std.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/kt-std/kottans-frontend/tree/main/task_tiny_js_world)

Please, review.


submissions/kt-std/tiny-js-world/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+   ((array) => {
+	   array.forEach( obj => {
+	   	let string = '';
+	   	for(prop in obj){
+	   		Array.isArray(obj[prop])? string += `${obj[prop].join(", ")}; ` : string += `${obj[prop]}; `;
+	   	}
+	   	print(string);
+	   });
+   })([dog, cat, woman, man, catWoman]);
```
OleksiyRudenko, line 35: Why do you need an IIFE here?

submissions/kt-std/tiny-js-world/index.js
```diff
@@ -0,0 +1,57 @@
...
+   const cat = new Inhabitant('cat', 'Murka', 'female', 0, 4, 'Meow!', ['Murchik', 'Anna']);
+   const woman = new Inhabitant('woman', 'Anna', 'female', 2, 2, 'Hi, folks!', ['John', 'Tom']);
+   const man = new Inhabitant('man', 'John', 'male', 2, 2, 'Ahoy!');
+   const catWoman = new Inhabitant('superhero', 'Selina', 'female?', 2, 2, cat.saySomething, ['Alice', 'Arizona']);
+
+   ((array) => {
+	   array.forEach( obj => {
+	   	let string = '';
+	   	for(prop in obj){
+	   		Array.isArray(obj[prop])? string += `${obj[prop].join(", ")}; ` : string += `${obj[prop]}; `;
```
OleksiyRudenko, line 31: Order of keys in object not guaranteed by design. Please use an explicit list of properties.
Consider also employing `Array` methods `map` and `join`

submissions/kt-std/tiny-js-world/index.js
```diff
@@ -0,0 +1,57 @@
...
+   
+   function Inhabitant (species, name, gender, hands, legs, sayings, friends){
+   		this.species = species;
+   		this.name = name;
+   		this.gender = gender;
+   		this.hands = hands === 0 ? `${this.species}s don't have hands! They have paws`: hands;
+   		this.legs = legs;
+   		this.friends = Array.isArray(friends) && friends.length ? friends : 'Forever alone...';
+   		this.saySomething = sayings;
+   }
```
OleksiyRudenko, line 19: It is good to see a constructor function as an OOP-oriented approach.
However, OOP is not only a syntax but rather a set of principles to follow.
Current approach qualifies at this level, but it won't be sufficient when you are back to your Tiny JS World in OOP Exercise task.

OleksiyRudenko (changes requested):
@kt-std pretty good job.
Please check comments below and [this message](https://t.me/c/1382428271/27235)  in students' chat.

submissions/kt-std/tiny-js-world/index.js
kt-std: Thank you, @OleksiyRudenko, for your time and for reviewing this!
It is not needed here. It was used just for shortening the code a bit and for the avoidance of global namespace pollution as the function returns nothing (I felt like a C programmer for a while  :sweat_smile:). Waiting for your reply to understand whether my reasoning is correct or it is not. I changed it to the function declaration in the updated version of the code, as well as fixed improper code formatting. I reformatted it via JSPrettier and ESLint due to the Google-style preset. I've also added some js-docs as ESLint was failing with errors. I hope it looks just how it has to. 
Finally, I calmed down with message: `✓ 0 errors and warnings`.
BTW, lots of thanks for providing info about such useful plugins!

submissions/kt-std/tiny-js-world/index.js
kt-std: I updated the the code with explicitly set order of object properties and applied `map` method in conjunction with `join`.

submissions/kt-std/tiny-js-world/index.js
kt-std: So should I change the code now to implement OOP principles, or it is better to do it later?

OleksiyRudenko: @kt-std you did well. You will have an opportunity to further improve your world code after a while.

OleksiyRudenko (approved):
@kt-std excellent job!
JS Doc annotations, extensive and relevant use of `Array` methods are among a few things worth special mentioning.
Carry on!
After a while you're gonna be back to your Tiny JS World. You will be expected to benefit from OOP learnings (just use es6 claases syntax then).

kt-std:
> @kt-std excellent job!
> JS Doc annotations, extensive and relevant use of `Array` methods are among a few things worth special mentioning.
> Carry on!
> After a while you're gonna be back to your Tiny JS World. You will be expected to benefit from OOP learnings (just use es6 claases syntax then).

Can't wait ! :)

PR #234, opened 2020-12-07
Oop exercise

Lu-sure:
### OOP exercise

[Demo](https://lu-sure.github.io/a-tiny-JS-world/) | [Code base](https://github.com/Lu-sure/a-tiny-JS-world/tree/repopulate-world)

Please, review.


submissions/Lu-sure/post-oop/index.js
```diff
@@ -0,0 +1,140 @@
...
+      this.species;
+    }
+
+    introduce() {
+      return `I\'ve existed for ${this.age} years`
+    }
+  }
+
+  class Human extends Inhabitant {
+    constructor(options) {
```
OleksiyRudenko, line 22: Please replace implicit options with named arguments.

submissions/Lu-sure/post-oop/index.js
```diff
@@ -0,0 +1,140 @@
...
+
+  class Woman extends Human {
+    constructor(options) {
+      super(options);
+      this.gender = 'female';
+      this.lovedBy;
+    }
+
+    beLoved() {
+      return `\"I\'m loved by ${this.lovedBy == this.name ? 'myself' : this.lovedBy}.\"`;
```
OleksiyRudenko, line 42: Template string literals do not requre `"` or `'` escaped.

OleksiyRudenko (changes requested):
@Lu-sure well done!
Please check comments below.
Please also make sure that
- shared properties are generalized in a super class (or super classes)
- props unique to a certain class are introduced in that class
- `introduce` method in every class refers directly only to props that are directly owned by that class
- child classes do not refer directly to props owned by parent class(es)

submissions/Lu-sure/post-oop/index.js
Lu-sure: Replaced. Check if correct, please.

submissions/Lu-sure/post-oop/index.js
Lu-sure: Fixed.

Lu-sure:
I've looked through your comments and have made according changes, hope code is correct now. Review, please.

submissions/Lu-sure/post-oop/index.js
```diff
@@ -0,0 +1,162 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+  // ======== OBJECTS DEFINITIONS ========
+  class Inhabitant {
+    constructor(name, age, spesies) {
```
OleksiyRudenko, line 10: A typo in `spesies`

submissions/Lu-sure/post-oop/index.js
```diff
@@ -0,0 +1,162 @@
...
+  class Human extends Inhabitant {
+    constructor(name, lastName, age, gender, saying) {
+      super(name, age, 'human');
+      this.gender = gender;
+      this.lastName = lastName;
+      this.saying = saying;
+    }
+
+    getName() {
+      return `${this.name} ${this.lastName}`;
```
OleksiyRudenko, line 34: Why not to follow `introduce` re-use/overloading pattern?

submissions/Lu-sure/post-oop/index.js
```diff
@@ -0,0 +1,162 @@
...
+    defendMyLord() {
+      return `Go away, strangers! woof-woof..`
+    }
+
+    introduce() {
+      return super.introduce() + this.defendMyLord();
+    }
+  }
+
+  function add_prop_value(o, S, prop) { o[prop] += ` & ${S[prop]}` };
```
OleksiyRudenko, line 118: What are `o` and `S` in fact? Quite mysterious naming

OleksiyRudenko (changes requested):
@Lu-sure pretty nice job
You exposed a plenty of creativity.
However there are some minimal requirements to meet.
E.g. the task requires that every inhabitant is represented not only with their name, but also with the number of limbs they have.
Please check the minimal requirements and adjust the inhabitants' design.

We welcome creativity as long as menimal requirements are met.

submissions/Lu-sure/post-oop/index.js
Lu-sure: Fixed.

submissions/Lu-sure/post-oop/index.js
Lu-sure: I don't know the reason, so I made changes))
Check if correct, please.
I need a space between first and last name, didn't found better way than using template literals. If there are a better way to add that space, please, tell me

submissions/Lu-sure/post-oop/index.js
Lu-sure: Got my mistake, made changes.

Lu-sure:
Fixed trivial bugs, still in progress about adjusting the inhabitants' design.
Have a few question about it.
Does it matter how exactly the output printed, or the main purpose is just print the needed info, no matter the order and decorative text?
Is it required that shared property like `.saying` is a property of super class  (`Inhabitant` in my case) or it is ok, that some saying text is actualy printed, but using different functionality (`.saying` property for humans, `.sayMeow()` method for cats for example)?
If there are `.hands` not defined for `cat` obj or `.gender` not defined for `alien`, should I print `undefined`, etс. or can just drop out this info from the printed message? 
Sorry for so many questions, I just don't really understand, fulfillment of which requirements are sufficient for pr to be approved.

OleksiyRudenko:
> Does it matter how exactly the output printed, or the main purpose is just print the needed info, no matter the order and decorative text?

Info is of higher priority if compared with decoration

> Is it required that shared property like `.saying` is a property of super class (`Inhabitant` in my case) or it is ok, that some saying text is actualy printed, but using different functionality (`.saying` property for humans, `.sayMeow()` method for cats for example)?

Saying is basically what every inhabitant has. Whether it is implemented by means of a shared property or shared method doesn't really matter. If a method then certain level of abstraction is approrpiate (e.g. `getSaying` where implementation is different for humans and pets)

> If there are `.hands` not defined for `cat` obj or `.gender` not defined for `alien`, should I print `undefined`, etс. or can just drop out this info from the printed message?

If alien has no gender at all, then this means you need a base class that has no `gender` property at all and interim base class that introduces `gender` property. This will add to implementation complexity but that was your choice at the end of the day.

> Sorry for so many questions, I just don't really understand, fulfillment of which requirements are sufficient for pr to be approved.

Those that are described in the original project. Adding support for aliens adds the level of complexity, it is up to you to keep them or not. Minimal requirements should be met anyway.

Lu-sure:
Fixed, as I understood. Please, review. 
And thank you so much for patience))

submissions/Lu-sure/post-oop/index.js
```diff
@@ -0,0 +1,154 @@
...
+      super('human', name, age, gender, [2, 2], saying);
+      this.lastName = lastName;
+    }
+
+    getName() {
+      return `${super.getName()} ${this.lastName}`;
+    }
+
+    getLimbs() {
+      return `I have ${this.limbs[0]} hands and ${this.limbs[1]} legs`
```
OleksiyRudenko, line 43: Not a really obvious interpretation.
Imagine we introduce birds. What index will represent wings? Should a developer keep it just in mind or browse the code to find it out?
Explicitness is good for maintainability.

OleksiyRudenko (approved):
@Lu-sure approved with a reservation
There are reasons we have named properties, not just an array of properties

PR #236, opened 2020-12-07
JS DOM task

vakulinina:
### JS DOM

[Demo](https://vakulinina.github.io/kottans-js-dom/) |
[Code base](https://github.com/vakulinina/kottans-js-dom)

Please, review.

### DIFF
--- a/submissions/vakulinina/js-dom/index.html
+++ b/submissions/vakulinina/js-dom/index.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <link rel="preconnect" href="https://fonts.gstatic.com">
+  <link href="https://fonts.googleapis.com/css2?family=Quicksand:wght@400;700&display=swap" rel="stylesheet">
+  <link rel="stylesheet" href="style.css">
+  <title>DOM</title>
+</head>
+
+<body>
+  <header>
+    <button class="menu-button"><img src="img/menu-icon.svg" alt="menu"></button>
+    <h1>Coffee drinks</h1>
+  </header>
+  <main>
+    <nav class="nav-bar">
+      <ul class="nav-buttons">
+        <li><button class="nav-button" id="americano">Americano</button></li>
+        <li><button class="nav-button" id="black-eye">Black-eye</button></li>
+        <li><button class="nav-button" id="caffe-latte">Caffe-latte</button></li>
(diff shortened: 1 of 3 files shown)


submissions/vakulinina/js-dom/script.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+const getProp = (id, prop) => coffeeDrinks.filter(obj => obj.name === id)[0][prop];
+
+menuButton.addEventListener('click', function () {
+  navBar.classList.toggle('visible');
+})
+
+navButtons.addEventListener('click', function (evt) {
+  if (evt.target.tagName === 'BUTTON') {
```
MaxKalashnyk, line 95: I'd recommend using checking this element by class name because you can have multiple elements with button tag.

submissions/vakulinina/js-dom/script.js
```diff
@@ -0,0 +1,103 @@
...
+    const id = evt.target.id;
+    hideMenu();
+    itemTitle.textContent = getProp(id, 'name').toUpperCase();
```
MaxKalashnyk, line 98: Look how many times you called the `getProp` in this function. It seems 5 times. 
So 5 times you filter the `coffeeDrinks` list and take the necessary property from the element.
I'd recommend using `Array.prototype.find` for finding necessary object by the `id` from the `coffeeDrinks` list and then just get necessary properties from it.
It can help you to increase the performance of your app.

submissions/vakulinina/js-dom/index.html
```diff
@@ -0,0 +1,48 @@
...
+
+<body>
+  <header>
+    <button class="menu-button"><img src="img/menu-icon.svg" alt="menu"></button>
+    <h1>Coffee drinks</h1>
+  </header>
+  <main>
+    <nav class="nav-bar">
+      <ul class="nav-buttons">
+        <li><button class="nav-button" id="americano">Americano</button></li>
```
MaxKalashnyk, line 21: You can generate this list from your script file

MaxKalashnyk (changes requested):
Good job! Look at my comments below, please

submissions/vakulinina/js-dom/index.html
vakulinina: May I ask what for? Will it have an impact on performance or something else?

MaxKalashnyk: It will help you depend only on your `coffeeDrinks` list. If you'll change some element in your hardcoded navigation items , your app can be broken.

vakulinina:
Thanks, all fixed.

MaxKalashnyk (approved):
Well done! Nice work!

PR #241, opened 2020-12-07
Memory game task

UA-girl:
### Memory – Pair Game

[Demo](https://ua-girl.github.io/MemoryGame/) |
[Code base](https://github.com/UA-girl/MemoryGame)

Please, review.


submissions/UA-girl/task_memory_game/js/index.js
```diff
@@ -0,0 +1,127 @@
...
+    div_back.classList.add('back');
+    cardItem.appendChild(div_front);
+    cardItem.appendChild(div_back);
+    return cardItem;
+}
+
+
+function updateGame() {
+    shuffleArray(cards);
+    cards.forEach(card => container.appendChild(card));
```
AMashoshyna, line 61: DOM operations are very expensive and should be batched if possible. Here instead of appending 12 card one by one you can create a wrapper (take a look at `document.createDocumentFragment` API), append cards to it and then attach this wrapper to a DOM node. this will result in 1 DOM access instead of 12.

submissions/UA-girl/task_memory_game/js/index.js
```diff
@@ -0,0 +1,127 @@
...
+    },
+    {
+        name: 'lingonberry',
+        src: 'img/lingonberry.svg'
+    }
+];
+const TIME_DELAY = 400;
+const container = document.querySelector('.container-game');
+let cards = [];
+let starTime, endTime;
```
AMashoshyna, line 30: Typo? starTime => startTime

AMashoshyna (commented):
Nice job! Just some small fixes to do

submissions/UA-girl/task_memory_game/js/index.js
UA-girl: Thank you! Fixed )

submissions/UA-girl/task_memory_game/js/index.js
UA-girl: Typo )

AMashoshyna:
Very good!

One more thing to fix: the user may click the space between cards and then this happens:

image

Please handle this case in `toggleCard` function

UA-girl:
Ok, fixed toggleCard )

AMashoshyna:
Nice! Approved.

UA-girl:
Thank you for attention and advices!

PR #244, opened 2020-12-08
Add OOP exercise

VBystrov:
### OOP exercise

[Demo](https://vbystrov.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/VBystrov/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/VBystrov/tiny-world-oop/index.js
+++ b/submissions/VBystrov/tiny-world-oop/index.js
@@ -9,9 +9,12 @@
 // Define your objects here
 
 class Creature {
-  constructor(name, gender = 'unknown') {
+  constructor(name, gender, species, legs, saying) {
     this.name = name;
     this.gender = gender;
+    this.species = species;
+    this.legs = legs;
+    this.saying = saying;
     this.friends = [];
   }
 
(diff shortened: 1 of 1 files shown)


submissions/VBystrov/tiny-world-oop/index.js
```diff
@@ -0,0 +1,110 @@
...
+    return this.friends;
+  }
+}
+
+class Dog extends Creature {
+  constructor(name, gender) {
+    super(name, gender);
+    this.species = 'dog';
+    this.legs = 4;
+    this.saying = 'woof!';
```
OleksiyRudenko, line 29: All of the creatures are of  certain species, have legs and a saying.
All shared props should be generalized into a base class.
Base class constructor can have different specification (number and set of arguments) than any of its child classes.
Let a child class constructor take arguments minimally required to set up an instance and call `super` with parameters required to get initialized via parent's constructor.

submissions/VBystrov/tiny-world-oop/index.js
```diff
@@ -0,0 +1,110 @@
...
+    if (key === 'friends' && !creature[key].length) {
+      continue;
+    }
+    if (creature[key]) {
+      template += `${key}: ${creature[key]}; `;
+    }
+  }
+
+  return template;
+};
```
OleksiyRudenko, line 96: Each class can represent itself.
Make this function a method of the base class and child classes.
Remember every class should refer to the properties it owns directly.
In a child class you can overload an inherited method and also benefit from calling an inherited method by means of `super.methodName()`. This way child's methods gets whatever is needed from a parent class and supplements it with its own data.

If method is named `toString` then this method is called implicitly in string context. I.e. `print(object.toString())` is equivalent of `print(object)` as `print` function ensures string context for its arguments.

OleksiyRudenko (changes requested):
@VBystrov well done!
Let's make it more OOP and SOLID principles compliant.

submissions/VBystrov/tiny-world-oop/index.js
VBystrov: Done.

submissions/VBystrov/tiny-world-oop/index.js
VBystrov: Done.

OleksiyRudenko (approved):
@VBystrov excellent job!

PR #259, opened 2020-12-10
add dom-api-task

NJul:
### dom-api-task

[Demo](https://njul.github.io/dom-api-task/)
[Code base](https://github.com/NJul/dom-api-task)

Please, review.


submissions/NJul/dom-api-task/app.js
```diff
@@ -0,0 +1,98 @@
...
+          panel.classList.add('active');
+        } else {
+          panel.classList.remove('active');
+        }
+      });
+    }
+    navBar.classList.toggle('open');
+  });
+
+  let links = ul.getElementsByClassName('nav-item');
```
AMashoshyna, line 88: Why do you prefer `getElementsByClassName` over `querySelector` here?

submissions/NJul/dom-api-task/app.js
```diff
@@ -0,0 +1,98 @@
...
+    source: `<a href="https://www.oxfordlearning.com/how-to-study-effectively/" target="_blank">Source</a>`,
+  },
+];
+
+document.addEventListener('DOMContentLoaded', () => {
+  const hamburgerBtn = document.getElementById('hamburgerBtn');
+  const navBar = document.getElementById('navBar');
+
+  hamburgerBtn.addEventListener('click', () => {
+    navBar.classList.toggle('open');
```
AMashoshyna, line 33: This leads to a situation where `class` attribute may have no value at certain points. This is not not allowed by html specification, since only boolean attributes (like "checked") may have no value. Can you think of a different approach to class toggling?

submissions/NJul/dom-api-task/app.css
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  #navBar {
+    width: 100vw;
+    transform: translateX(-1000px);
+    transition: 500ms transform;
+  }
+
+  #navBar.open {
+    transform: translateX(0);
```
AMashoshyna, line 189: Let's make disappearance of the menu smooth as well (appearance looks nice)

AMashoshyna (changes requested):
Good job! Please take a look at my comments below

submissions/NJul/dom-api-task/app.css
NJul: Thank you for your review!

AMashoshyna:
Good job! Sorry for the delayed reply.

PR #264, opened 2020-12-11
JS post-OOP

zalevskyi:
### JS post-OOP

[Demo](https://zalevskyi.github.io/kottans-frontend/practice/a-tiny-js-world-oop/index.html)
[Code base](https://github.com/zalevskyi/kottans-frontend/tree/main/practice/a-tiny-js-world-oop)

Please, review.


OleksiyRudenko:
@zalevskyi OOP Exercise is a task on its own.
Code for every task should be submitted in its own directory.
Please fix.

OleksiyRudenko (changes requested):
@zalevskyi nice approach with composition.
Here is the problem. Species that do not have e.g. hands by natural design still have the property.
Imagine we add birds. Humans will then have undefined number of wings, birds will have undefined number of hands, etc. Then we add some more species and other props specific to particular species etc.
At the end every inhabitant will bear a burden of props they do not really have.
The point of OOP and SOLID, whether this is a class inheritance or composition pattern, is that every object has only those properties it really needs.
Can you resolve this issue within current scope?

zalevskyi:
Thank you for pointing out this issue.
Resolved.

submissions/zalevskyi/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,148 @@
...
+      this.genus=genus
+   }
+   descriptionArray() {
+      return [`species: ${this.genus}`]
+   }
+}
+class MammalNonMarine extends Species {
+   constructor(genus) {
+      super(genus)
+      this.legs = 4
```
OleksiyRudenko, line 22: No good reasons to have any default value here. There is no good reasons to favour four-legged animals.
Obviously this should be initialized from constructor's arguments.

You will not have "override" it child class - no just good reason for this.

submissions/zalevskyi/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,148 @@
...
+   descriptionArray() {
+      return [...super.descriptionArray(), `sound: ${this.sound}`]
+   }
+}
+class Hominoid extends MammalNonMarine {
+   constructor(genus, sound) {
+      super(genus)
+      this.legs = 2 // overides property value of the prototype
+      this.hands = 2
+      this.sound = sound
```
OleksiyRudenko, line 42: Both Hominoid and CarnivoraNonMarine have `sound`, which means it is a shared property and thus should be generalized to their common ancestor.

OleksiyRudenko (changes requested):
@zalevskyi well done!

A bit excessive taxonomy. Whenever any class is extended by a single child class then can be effectively merged. Keep this in mind.

In this task let assume it is OK.

Let's rather focus on more important things.

zalevskyi:
@OleksiyRudenko  Thank you for comments. Corrected. Please review once more

OleksiyRudenko (approved):
@zalevskyi well done!

PR #276, opened 2020-12-14
Friends app task

ses3332002:
### Friends app

[Demo](https://ses3332002.github.io/kottans-frontend/task_friends_app/)
[Code base](https://github.com/ses3332002/kottans-frontend/tree/main/task_friends_app)

Please, review.


Aldegid:
Good job! See me comments, please.

submissions/ses3332002/friends-app/app.js
```diff
@@ -0,0 +1,76 @@
...
+const hamburgerMenu = document.querySelector(".hamburger_menu");
+const mainArea = document.querySelector("main");
+const optionForm = document.forms.sorting_n_filtering;
+let receivedData = [];
+let dataToRender = [];
+let filterCriterionSex = () => true;
+let filterCriterionAge = () => true;
+let filterCriterionName = () => true;
+let sortCriterion = () => true;
+as_is.checked = true;
```
Aldegid, line 11: Where are these variables declared and used?

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
```diff
@@ -0,0 +1,76 @@
...
+hamburgerMenu.addEventListener("click", () => hamburgerMenu.classList.toggle("hamburger_menu-opened"));
+optionForm.addEventListener("input", changeData);
+
+fetch("https://randomuser.me/api/?nat=gb,us,fr&results=60&inc=gender,name,location,email,picture,dob")
+  .then((response) => {
+    return response.json();
+  })
+  .then((data) => {
+    receivedData = data.results;
+    dataToRender = receivedData.slice().filter(filterCriterionSex).filter(filterCriterionAge).filter(filterCriterionName).sort(sortCriterion);
```
Aldegid, line 25: Do you really need to do all this filtering and sorting on the first render?

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
ses3332002: Yes. If the first render is going without any filters or sorting then after user choose any "neutral" filter (like "as is" or "all") cards array is changing.

submissions/ses3332002/friends-app/app.js
ses3332002: I'm addressing DOM elements through their ID's.
Such a hack like using a button in a form without any listener to reload the page.
I use it to clear filters and sorting fields in form, because with common "reload" in browser they still keep old values, despite the HTML attributes.

Aldegid: Please don’t use id-named global variables to access elements.
This behavior is described in the specification, so it’s kind of standard. But it is supported mainly for compatibility.

The browser tries to help us by mixing namespaces of JS and DOM. That’s fine for simple scripts, inlined into HTML, but generally isn’t a good thing. There may be naming conflicts. Also, when one reads JS code and doesn’t have HTML in view, it’s not obvious where the variable comes from.

In real life document.getElementById is the preferred method.

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
ses3332002: Ok, I'll remake it.

submissions/ses3332002/friends-app/app.js
ses3332002: Yes

submissions/ses3332002/friends-app/app.js
```diff
@@ -0,0 +1,76 @@
...
+    <div class="friend_card__text friend_card__sex">${item.gender == "male" ? "мужчина" : "женщина"}</div>\
+    <div class="friend_card__text friend_card__address">${item.location.country}, ${item.location.city}</div>\
+    <div class="friend_card__text friend_card__email">${item.email}</div>`;
+    contentToRender.append(card);
+  });
+  mainArea.append(contentToRender);
+};
+
+function changeData({target}) {
+  if (target.tagName != "INPUT") {
```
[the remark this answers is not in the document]
ses3332002: Ok. It's just my preference - to use "if/else" but not "case".

submissions/ses3332002/friends-app/app.js
```diff
@@ -0,0 +1,76 @@
...
+function render() {
+  mainArea.innerHTML = "";
+  const contentToRender = document.createDocumentFragment();
+  dataToRender.forEach((item) => {
+    let card = document.createElement("div");
+    let lastNumberOfAge = item.dob.age.toString().slice(-1);
+    card.classList.add("friend_card");
+    card.innerHTML = `<img src=${item.picture.large} alt="photo" class="friend_card__photo">\
+    <div class="friend_card__text friend_card__name">${item.name.first} ${item.name.last}</div>\
+    <div class="friend_card__text friend_card__age">${item.dob.age} ${(lastNumberOfAge == "1") ? "год" : (lastNumberOfAge == "2")||(lastNumberOfAge == "3")||(lastNumberOfAge == "4") ? "года" : "лет"}</div>\
```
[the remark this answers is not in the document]
ses3332002: Ok, I'll remake it.

submissions/ses3332002/friends-app/app.js
Aldegid: so why you do all this filtering and sorting on the first render?

ses3332002: As I said before, if the first render is going without any filters or sorting then after user choose any "neutral" filter (like "as is" or "all") cards array is changing. I couldn't find out why, so I solved this problem "at the root".
Now I removed from Demo this filtering on the first render, so you can see it yourself: reload page, remember some faces, choose "man" and then - choose "all". They will be another "all", then after reload :)

submissions/ses3332002/friends-app/app.js
ses3332002: I remade it.

submissions/ses3332002/friends-app/app.js
ses3332002: I've found a mistake and remade it. Now there's no filtering and sorting on the first render.

submissions/ses3332002/friends-app/app.js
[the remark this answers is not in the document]
ses3332002: I remade it.

submissions/ses3332002/friends-app/app.js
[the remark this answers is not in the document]
ses3332002: I remade it.

submissions/ses3332002/friends-app/app.js
```diff
@@ -2,27 +2,33 @@
...
-all.checked = true;
-age_from.value = "";
-age_to.value = "";
-search_name.value = "";
+let sortCriterion = () => false;
+asIsOption.checked = "true";
+allOption.checked = "true";
+ageFromOption.value = "";
+ageToOption.value = "";
+searchNameOption.value = "";
```
Aldegid, line 21: Do you really need this assignments? 
input text/number value is empty string by default
for input type radio you can set html attribute `cheched`

submissions/ses3332002/friends-app/app.js
```diff
@@ -2,27 +2,33 @@
...
+const allOption = document.querySelector("#all");
+const ageFromOption = document.querySelector("#age_from");
+const ageToOption = document.querySelector("#age_to");
+const searchNameOption = document.querySelector(".filter_name");
+const requestString = "https://randomuser.me/api/?nat=gb,us,fr&results=60&inc=gender,name,location,email,picture,dob";
 let receivedData = [];
 let dataToRender = [];
 let filterCriterionSex = () => true;
 let filterCriterionAge = () => true;
 let filterCriterionName = () => true;
```
Aldegid: you don't need this function expression here as a global variables. You only use them in the `changeData` function

submissions/ses3332002/friends-app/app.js
```diff
@@ -2,27 +2,33 @@
...
 
-fetch("https://randomuser.me/api/?nat=gb,us,fr&results=60&inc=gender,name,location,email,picture,dob")
+fetch(requestString)
   .then((response) => {
     return response.json();
   })
   .then((data) => {
     receivedData = data.results;
-    dataToRender = receivedData.slice().filter(filterCriterionSex).filter(filterCriterionAge).filter(filterCriterionName).sort(sortCriterion);
+    dataToRender = receivedData.slice();
```
Aldegid: Please, add error handling

submissions/ses3332002/friends-app/app.js
```diff
@@ -31,11 +37,14 @@ function render() {
   const contentToRender = document.createDocumentFragment();
   dataToRender.forEach((item) => {
     let card = document.createElement("div");
-    let lastNumberOfAge = item.dob.age.toString().slice(-1);
+    function sexString(age) {
+      const lastNumberOfAge = age.toString().slice(-1);
+      return (lastNumberOfAge == "1") ? "год" : (lastNumberOfAge == "2")||(lastNumberOfAge == "3")||(lastNumberOfAge == "4") ? "года" : "лет";
```
Aldegid, line 42: Please, use constants for a 'magic numbers'

Aldegid (changes requested):

submissions/ses3332002/friends-app/index.html
```diff
@@ -0,0 +1,59 @@
...
+      <div class="filter_caption">по возрасту</div>
+      <div class="input_option">
+        <label for="age_from">от</label><input type="number" name="age_from" id="age_from" class="filter_age"/>
+        <label for="age_to">до</label><input type="number" name="age_to" id="age_to" class="filter_age"/>
+      </div>
+      <div class="filter_caption">Поиск по имени</div>
+      <div class="input_option">
+        <input type="text" name="search_name" id="search_name" class="filter_name"/>
+      </div>
+      <button class="reset_button">Обновить список</button>
```
Aldegid, line 52: Refreshing page by submit is not a good idea. Please, write a functionality to reset the list of users and reset the form fields to their original state.

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
```diff
@@ -48,28 +57,39 @@ function changeData({target}) {
...
+      filterCriterionSex = (item) => item.gender == "female";
+      break;
+    case "age_from":
+    case "age_to":
+      if ((ageFromOption.value)&&(ageToOption.value)) {
+        filterCriterionAge = (item) => (item.dob.age >= ageFromOption.value)&&(item.dob.age <= ageToOption.value);
+      };
+      break;
+    case "search_name":
+      filterCriterionName = (item) => (item.name.first.toLowerCase().includes(searchNameOption.value.toLowerCase()))||(item.name.last.toLowerCase().includes(searchNameOption.value.toLowerCase()));
```
Aldegid, line 92: It's not reqirements? but recomendation at this stage. Your function does everything at once. In the future, try to separate logic into simpler parts. This will help you read and maintain your code better.

submissions/ses3332002/friends-app/app.js
ses3332002: I code on Linux and use Firefox for developing and testing. There is a bug, that Firefox does not clear form inputs on page reload, so I need to clear them this way.
I will use autocomplete="off" on a form to solve this issue.

submissions/ses3332002/friends-app/app.js
ses3332002: changeData function is called every time on "input" event on form. But I don't need to reset all the filters and sorting options every time, I need to keep the options that user already set. So I need this options to be stored somewhere outside changeData function, so I need to use global variables.

submissions/ses3332002/friends-app/app.js
ses3332002: I remade it.

submissions/ses3332002/friends-app/app.js
ses3332002: I added simple error handler. Is it OK?

submissions/ses3332002/friends-app/app.js
ses3332002: I remade it.

submissions/ses3332002/friends-app/index.html
ses3332002: I remade it.

submissions/ses3332002/friends-app/app.js
```diff
@@ -39,7 +40,9 @@ function render() {
     let card = document.createElement("div");
     function sexString(age) {
       const lastNumberOfAge = age.toString().slice(-1);
-      return (lastNumberOfAge == "1") ? "год" : (lastNumberOfAge == "2")||(lastNumberOfAge == "3")||(lastNumberOfAge == "4") ? "года" : "лет";
+      const numberForGod = "1";
+      const numbersForGoda = ["2", "3", "4"];
```
Aldegid, line 44: whatIsThisNeimingTakoy?)

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
```diff
@@ -2,24 +2,21 @@
...
 let filterCriterionAge = () => true;
 let filterCriterionName = () => true;
 let sortCriterion = () => false;
-asIsOption.checked = "true";
-allOption.checked = "true";
-ageFromOption.value = "";
-ageToOption.value = "";
-searchNameOption.value = "";
 hamburgerMenu.addEventListener("click", () => hamburgerMenu.classList.toggle("hamburger_menu-opened"));
+resetButton.addEventListener("click", () => document.location.reload());
```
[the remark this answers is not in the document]
ses3332002: Ok, I'll remake it. But it was the fastest way!   :)

submissions/ses3332002/friends-app/app.js
ses3332002: :) I can use more correct and understandable names for this variables, but this names will be long as a train, like numberForYearOptionGod or numberForSuffixlessRussianYearWord. Is it Ok?

submissions/ses3332002/friends-app/app.js
Aldegid: i'm talking about this `filterCriterionSex filterCriterionAge filterCriterionName` vars

ses3332002: I'm talking about them too. This vars (and sortCriterion too) contains filter/sorting functions that are to be implemented to array of cards. In case the user choose two ore more filters (like "man", age from 25 to 52, name contains character "a") - I need to keep content of this vars outside of changeData function (where I change this filter/sorting functions depending of user's choice), so they are global vars.
Or maybe I don't understand what you mean.

Aldegid: what happens if you move them inside the `changeData` function?

submissions/ses3332002/friends-app/app.js
Aldegid: the variable should reflect the essence of its content)

submissions/ses3332002/friends-app/app.js
ses3332002: For example: 
user set sotring "by age" in form;
changeData function runs, overrides all the filtering and sorting settings, then sets a sorting order by age, prepares array of cards;
render function displays cards sorting by age;
user decides to see only woman sorting by age and set option "women" in form (in addition to already set option to sort by age);
changeData function runs, overrides all the filtering and sorting settings, then sets a filter to show only women, prepares array of cards;
render function displays cards filtered by sex but without sorting by age (because this setting was reassigned in changeData function when it worked last time);
user says "What the hell!?"
I can make a demo if you want to see this version.

submissions/ses3332002/friends-app/app.js
ses3332002: I renamed this vars.

submissions/ses3332002/friends-app/app.js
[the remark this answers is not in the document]
ses3332002: I made the separate function to reset main area, and I put there my filtering/sorting functions reassigning.

submissions/ses3332002/friends-app/app.js
```diff
@@ -2,24 +2,30 @@
...
-
-fetch(requestString)
+resetApp();
+  
+function resetApp() {
+  filterCriterionSex = () => true;
+  filterCriterionAge = () => true;
+  filterCriterionName = () => true;
+  sortCriterion = () => false;
+  optionForm.reset();
```
Aldegid, line 26: the `resetApp` function should reset the form and call `render` function the original users array. And nothing else.
**hint**: function `render` can accept users as argument.

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
```diff
@@ -40,9 +47,9 @@ function render() {
     let card = document.createElement("div");
     function sexString(age) {
```
Aldegid, line 48: you really need declare function `sexString` for every `render` function call?
Please, move it go a global scope.

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
ses3332002: Ok, but I use it only in render function.
Explain me, please, when which rule to apply?
Some language guides say that variables and functions should be defined where they are used (as class methods are defined in a class), and that this allows for "modularity" and the ability to reuse entire pieces of code as ready-made modules for other projects, for example: in the eventListeners callback describe all the variables and functions that are used in this callback. This does not "trash" the global scope, but this leads to the fact that these variables and functions are redefined every time each new callback instance is called, and this is obviously bad.
What is interesting: no matter what strategy I use for the location of such "accessory" functions and variables, the reviewers ask to change the approach to the opposite.
So, tell me - when is it necessary to "take out", and when is it not necessary to "take out"?

Aldegid: Write me in tg `@andrew_fedin`

ses3332002: Ok

ses3332002: I moved sexString function to a global scope.

submissions/ses3332002/friends-app/app.js
ses3332002: I added one more button - exclusively to reset form (and then it shows cards without filters and sorting). 
The old button loads new set of data (and call form reset to show new data without filters and sorting). 
I use render function with arguments.

How it works now:
when app starts it needs to load data - so it calls function prepareList;
when data is ready app needs to show it without any filters/sorting - so it calls function prepareForm;
when form resets app needs to render cards with reseted filters - so it calls function render;
when some input happens in form - function changeData runs, changes some filter and calls render;
and so on...

submissions/ses3332002/friends-app/app.js
```diff
@@ -6,51 +6,55 @@ const ageFromOption = document.querySelector("#age_from");
...
 
-function render() {
+function sexString(age) {
+  const lastNumberOfAge = age.toString().slice(-1);
+  const numberForYearWord = "1";
+  const numbersForYearsWord = ["2", "3", "4"];
+  return (lastNumberOfAge == numberForYearWord) ? "год" : (numbersForYearsWord.includes(lastNumberOfAge)) ? "года" : "лет";
+};
+
+function render(arrayToShow, sortingFunction, filterFunction1, filterFunction2, filterFunction3) {
```
Aldegid, line 53: The render function should only accept a list of users. Will it be filtered or sorted for it does not matter.
And you don't need to call filtering and sorting on the first render. I already told you about this.

submissions/ses3332002/friends-app/app.js
```diff
@@ -6,51 +6,55 @@ const ageFromOption = document.querySelector("#age_from");
...
-function resetApp() {
+function prepareForm() {
   filterCriterionSex = () => true;
   filterCriterionAge = () => true;
   filterCriterionName = () => true;
   sortCriterion = () => false;
   optionForm.reset();
-  
+  render(receivedData.slice(), sortCriterion, filterCriterionSex, filterCriterionAge, filterCriterionName);
+};
```
Aldegid: Use a relevant function name in relation to what it does

submissions/ses3332002/friends-app/app.js
```diff
@@ -6,51 +6,55 @@ const ageFromOption = document.querySelector("#age_from");
...
     receivedData = data.results;
-    dataToRender = receivedData.slice();
-    render();
+    prepareForm();
   })
   .catch((err) => {
     mainArea.innerHTML = "There was some error, try again!";
     console.error("there was some error:", err);
   });
 };
```
Aldegid: What do you know about Single Responsibility Principle? Your `prepareList` function both loads the data and displays it. And finally call `prepareForm()`. It's not ok.

Aldegid (changes requested):

submissions/ses3332002/friends-app/app.js
ses3332002: I just wanted to make my functions multipurpose. That is why there is no matter for render function which time it is called - the first time or not. But in the first time the sorting and filtering functions are "clear" (they return true or false, not real parameters).
What bad will happen in this way - if function runs the same the first time and the next, but the first time - with the initial filters?
Ok, I will remake it.

ses3332002: I remade it

submissions/ses3332002/friends-app/app.js
ses3332002: I remade it

submissions/ses3332002/friends-app/app.js
ses3332002: I remade it

Aldegid:
Well done! Resolve all outdated conversation and we go ahead.

ses3332002:
I resolved. Let's go ahead.

MaxKalashnyk (approved):
LGTM!

MaxKalashnyk:
@Aldegid , are we good with this PR?

Aldegid (approved):

PR #279, opened 2020-12-14
Frogger game

smillims:
### Object-Oriented JavaScript

[Demo](https://smillims.github.io/frontend-nanodegree-arcade-game/)
[Code base](https://github.com/smillims/frontend-nanodegree-arcade-game)

Please, review.

### DIFF
--- a/submissions/smillims/oop/app.js
+++ b/submissions/smillims/oop/app.js
@@ -1,0 +1,123 @@
+const block = {
+	width: 101,
+	height: 85
+}
+
+// Enemies our player must avoid
+const Enemy = function(x, y, player) {
+    // Variables applied to each of our instances go here,
+    // we've provided one for you to get started
+
+	 this.x = x;
+	 this.y = y;
+	 this.player = player;
+	 this.speed = 150 + (Math.random() * 100); //speed between 100 and 250;
+
+    // The image/sprite for our enemies, this uses
+	 // a helper we've provided to easily load images
+	 
+    this.sprite = 'images/enemy-bug.png';
+};
+
+// Update the enemy's position, required method for game
+// Parameter: dt, a time delta between ticks
+Enemy.prototype.update = function(dt) {
+    // You should multiply any movement by the dt parameter
+    // which will ensure the game runs at the same speed for
+	 // all computers.
+	 this.x += this.speed * dt;
+	 if (this.x > ctx.canvas.width + block.width){
+		 this.x = -101 * 2;
+	 }
+};
+
+// Draw the enemy on the screen, required method for game
+Enemy.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const checkCollisions = () => {
+	allEnemies.forEach(enemy => {
+		if(Math.abs(enemy.x - player.x) < 50 &&
+			Math.abs(enemy.y - player.y) < 50 ||
+			player.y < block.height - 85){
+			player.startOver();
+		}
+	})
+}
+
+// Now write your own player class
+// This class requires an update(), render() and
+// a handleInput() method.
+
+const Player = function(x, y, player) {
+	this.x = x;
+	this.y = y;
+	this.player = player;
+	this.sprite = 'images/char-boy.png';
+};
+
+Player.prototype.update = function() {
+
+};
+
+Player.prototype.render = function() {
+	ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+Player.prototype.handleInput = function(key) {
+	switch(key) {
+		case "up":
+			if(this.y > 0) {
+				this.y -= block.height;
+			}
+			break;
+		case "right":
+			if (this.x < block.width*3){
+				this.x += block.width;
+			}
+			break;
+		case "down":
+			if(this.y < block.height*4){
+				this.y += block.height;
+			}
+			break;
+		case "left":
+			if (this.x > 0){
+				this.x -= block.width;
+			}
+			break;
+	}
+};
+
+Player.prototype.startOver = function() {
+	this.x = 200;
+	this.y = 400;
+};
+
+// Now instantiate your objects.
+// Place all enemy objects in an array called allEnemies
+// Place the player object in a variable called player
+const player = new Player(200, 400);
+
+const enemy1 = new Enemy(0, 60);
+const enemy2 = new Enemy(0, 140);
+const enemy3 = new Enemy(0, 220);
+const allEnemies = [
+	enemy1,
+	enemy2,
+	enemy3
+];
+
+// This listens for key presses and sends the keys to your
+// Player.handleInput() method. You don't need to modify this.
+document.addEventListener('keyup', function(e) {
+    var allowedKeys = {
+        37: 'left',
+        38: 'up',
+        39: 'right',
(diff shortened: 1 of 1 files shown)


submissions/smillims/oop/app.js
```diff
@@ -0,0 +1,123 @@
...
+
+// Update the enemy's position, required method for game
+// Parameter: dt, a time delta between ticks
+Enemy.prototype.update = function(dt) {
+    // You should multiply any movement by the dt parameter
+    // which will ensure the game runs at the same speed for
+	 // all computers.
+	 this.x += this.speed * dt;
+	 if (this.x > ctx.canvas.width + block.width){
+		 this.x = -101 * 2;
```
OleksiyRudenko, line 30: `-101` resembles negated `block.width` or it is just a coincidence?

Check your entire code for mysterious numbers. They definitely have semantic meaning, thus worth defining as contants.

submissions/smillims/oop/app.js
```diff
@@ -0,0 +1,123 @@
...
+const player = new Player(200, 400);
+
+const enemy1 = new Enemy(0, 60);
+const enemy2 = new Enemy(0, 140);
+const enemy3 = new Enemy(0, 220);
+const allEnemies = [
+	enemy1,
+	enemy2,
+	enemy3
+];
```
OleksiyRudenko, line 110: Too much repetition. Imagine, we need creating 10 enemies.
Consider the following example:
```
const objects = [100, 150, 200].map(pos => new Circle(pos, 0));
```
Intentionally oversimplified, please do not copypaste and reconsider all variables naming.

submissions/smillims/oop/app.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// Now write your own player class
+// This class requires an update(), render() and
+// a handleInput() method.
+
+const Player = function(x, y, player) {
+	this.x = x;
+	this.y = y;
+	this.player = player;
```
OleksiyRudenko, line 56: Why do you need `this.player`? It doesn't seem to be used anywhere in `Player`'s methods. Besides if this is a selfreference then `this` is already a selfreference.

submissions/smillims/oop/app.js
```diff
@@ -0,0 +1,123 @@
...
+	 }
+};
+
+// Draw the enemy on the screen, required method for game
+Enemy.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const checkCollisions = () => {
+	allEnemies.forEach(enemy => {
```
OleksiyRudenko, line 40: Classes should not refer to any global variables, like global variable `player`, which is an instance of `Player` class (referring to global constants and globals provided by the gaming platform like `Resources` is OK).

OleksiyRudenko (changes requested):
@smillims good job.
However there is certain room for improvement.

submissions/smillims/oop/app.js
smillims: Yeah, it's resembles negated `block.width`

submissions/smillims/oop/app.js
smillims: Resolved

submissions/smillims/oop/app.js
smillims: Yes, it was self reference, so I removed it

submissions/smillims/oop/app.js
smillims: Thanks for your advice, resolved

submissions/smillims/oop/app.js
```diff
@@ -0,0 +1,121 @@
...
+Enemy.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+Enemy.prototype.checkCollisions = function() {
+	if(Math.abs(this.x - player.x) < widthEnemies &&
+		Math.abs(this.y - player.y) < widthEnemies ||
+		player.y < 0){
+			setTimeout(() => player.startOver(), 120);
+	};
```
OleksiyRudenko, line 50: Classes should not refer to any global variables directly, like global variable `player`, which is an instance of `Player` class (referring to global constants and globals provided by the gaming platform like `Resources` is OK).

https://github.com/kottans/frontend-2021-homeworks/pull/279#discussion_r546909025

Hint: Check the way `Player` gets to know about `positionPlayer.playerX`

OleksiyRudenko (changes requested):
@smillims well improved.
However, one problem was solved too literraly, so that it just turned out in different location with different "participants". But the problem is all the same in its essence.

submissions/smillims/oop/app.js
smillims: I tried to understand how to correct it's problem and solve of this problem was refer to `player` and setup `positionPlayer.playerX`, `positionPlayer.playerX`; if it's not solve for my problem so I don't understand how to fix it if I not need to refer to `player`

OleksiyRudenko: @smillims let's make a step at a time.
How does Enemy class learn about its `y` position that is defined somewhere outside the class?

smillims: @OleksiyRudenko it's defined in `positionEnemyY` which defined in `allEnemies` by creating `new Enemy`

OleksiyRudenko: Not exactly. The more accurate description of the situation would be:
Each Enemy's y position is stored in a global constant `positionEnemyY`, which is passed to an `Enemy` constructor as a parameter by means of `new Enemy`, and in the `Enemy` class constructor the argument for y position gets assigned to an `Enemy`'s own `y` property (which is then referred to as `this.y` in all `Enemy`'s methods).

Do you see a way to handle global `player` properly?

smillims: by using `this.player.x/y`?

OleksiyRudenko: Just pass global `player` to each enemy and store it in `this.player`

smillims:
@OleksiyRudenko I think that it's finally done

submissions/smillims/oop/app.js
```diff
@@ -0,0 +1,120 @@
...
+	width: 101,
+	height: 85
+};
+const positionEnemyY = [60, 140, 220];
+const positionPlayer = {
+	playerX: 200,
+	playerY: 400,
+};
+
+const widthEnemies = [80];
```
OleksiyRudenko, line 11: Is there any good reason to have an array here? What will happen if we add yet another value in the list?

OleksiyRudenko (changes requested):
@smillims pretty well done.
Noticed a bug: it is not safe for the player to remain on the middle bricks row even when there is no bug approaching along this row - a bug moving a level lower kills the player.

submissions/smillims/oop/app.js
smillims: Oh, then would be some troubles. I'm:
1) Decrease value
2) Change variable

OleksiyRudenko (approved):
@smillims great job!
Please test your app at every update and double check your code before submission.

smillims:
> 
> 
> @smillims great job!
> Please test your app at every update and double check your code before submission.

Yeah, of course, big thank you for wasting your time, you are so help me in this deal. Now I be back to oop exercise

PR #280, opened 2020-12-15
A tiny JS world post-OOP

natashafir:
### A tiny JS world post 

[Demo](https://natashafir.github.io/a-tiny-JS-world/)
[Code base](https://github.com/natashafir/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/natashafir/post-oop/index.js
+++ b/submissions/natashafir/post-oop/index.js
@@ -14,10 +14,46 @@ class Inhabitants {
     }
 }
 
-const woman = new Inhabitants('human', 'Emilia', 'female', 2, 2, 'Hello, World!');
-const man = new Inhabitants('human', 'Efrain', 'male', 2, 2, 'Nice to meet you!');
-const cat = new Inhabitants('cat', 'Fluffy', 'female', 4, 0, 'Meow!');
-const dog = new Inhabitants('dog', 'Buddy', 'male', 4, 0, 'Woof-woof!');
+class Human extends Inhabitants {
+    constructor(name, gender, saying) {
+        super('human', name, gender, 2, 2, saying)
+    }
+}
+
+class Woman extends Human {
+    constructor(name, saying) {
+        super(name, 'female', saying)
+    }
+}
+
+class Man extends Human {
+    constructor(name, saying) {
+        super(name, 'male', saying)
+    }
+}
+
+class Animal extends Inhabitants {
+    constructor(species, name, gender, saying) {
+        super(species, name, gender, 4, 0, saying)
+    }
+}
+
+class Cat extends Animal {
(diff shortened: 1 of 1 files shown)


OleksiyRudenko (changes requested):
@natashafir nice migration from object literals to es6 class syntax.
However, OOP and SOLID principles can help you write more scalable and maintainable code.
Imagine, we want hundreds of inhabitants, and we want to add ostrichs, chickens and other birds.
They do not have hands by natural design, like humans, unfortunately, do not have wings.
OOP/SOLID principles can help you design objects structures so that every object has the properties that exactly reflect their natural design.
Consider revisiting [further materials in the original project](https://github.com/OleksiyRudenko/a-tiny-JS-world#learn-on-your-own) to learn something new and helpful, and improve your code then.

submissions/natashafir/post-oop/index.js
```diff
@@ -0,0 +1,60 @@
+class Inhabitants {
+    constructor(species, name, gender, legs, hands, saying) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.legs = legs;
+        this.hands = hands;
```
OleksiyRudenko, line 7: Not all inhabitants have hands by natural design. Not that they just have `0` hands or `null` hands or `undefined` number of hands, but they do not hands property at all.
Imagine, we add birds. Under current paradigm every inhabitant would have wings, just some of them would have zero wings. This way the `Inhabitant` class becomes too general, and not easy to maintain.

The beauty of OOP is that every base class owns and handles only shared props, and every child class introduces and handles props unique to that child class.

Handling of props:
a child class can have its own method (e.g. `methodName`) where it calls parent's method (e.g. `super.methodName()`) and supplements it with some data from unique properties it **owns** directly.

submissions/natashafir/post-oop/index.js
```diff
@@ -0,0 +1,60 @@
...
+        this.saying = saying;
+    }
+
+    sayHi() {
+        return "My name is " + this.name + ". I'm " + this.species + ". My gender is " + this.gender +
+            ". I have " + this.legs + " legs and " + this.hands + " hands. " + this.saying
```
OleksiyRudenko, line 13: Quite a lengthy string composition. Imagine, we add a dozen of props more.
How about splitting it in logical groups in an array and then join them with `Array.join`?

OleksiyRudenko (changes requested):
@natashafir a really good job!
Let's make it even better.
We want OOP and SOLID principles followed as well as improve readability and maintainability of the code.

OleksiyRudenko (approved):
@natashafir great job!

PR #291, opened 2020-12-16
A Tiny JS World post-OOP

ORybka:
### OOP exercise

[Demo](https://orybka.github.io/a-tiny-JS-world/)
[Code base](https://github.com/ORybka/a-tiny-JS-world)

Please, review.


submissions/ORybka/post-a-tiny-js-world/index.js
```diff
@@ -0,0 +1,61 @@
...
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
```
OleksiyRudenko, line 15: No all inhabitants have hands by natural design. Not that they just have `0` hands or `null` hands or `undefined` number of hands, but they do not hands property at all.
Imagine, we add birds. Under current paradigm every inhabitant would have wings, just some of them would have zero wings. This way the `Inhabitant` class becomes too general, and not easy to maintain.

The beauty of OOP is that every base class owns and handles only shared props, and every child class introduces and handles props unique to that child class.

Handling of props:
a child class can have its own method (e.g. `methodName`) where it calls parent's method (e.g. `super.methodName()`) and supplements it with some data from unique properties it **owns** directly.

OleksiyRudenko (changes requested):
@ORybka well done!
Let's benefit from OOP and SOLID principles.

submissions/ORybka/post-a-tiny-js-world/index.js
ORybka: @OleksiyRudenko  Thank you for this advice!
I am not sure whether my solution is good, but I can't solve it better now :( 
It is very helpful that you can make your own methods for children based on the parent's methods but for now I can't understand how to add `hands` to `super.getValues()`

OleksiyRudenko: @ORybka you do not need to pass anything to a super method if it doesn't require.
```
class A {
  constructor() {
  }
  method() {
    return 'a'; // "a"
  }
}

class B extends A {
  constructor() {
    super();
  }
}


class C extends B {
  constructor() {
    super();
  }
  method() {
    return super.method() + 'c'; // "ac"
  }
}
```

Consider `super` a substitute to the name of a class nearest to current class that has `method` defined.
It's a shortcut that allows you e.g. insert interim classes without rewriting (which would be required if we used notation like `A.super()`)

`super()` is a shortcut for `super.constructor()`

submissions/ORybka/post-a-tiny-js-world/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+class Human extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super('human', name, gender, 2, saying, friends);
+    this.hands = 2;
+  }
+
+  getValues() {
+    return super.getValues(this.hands);
+  }
```
OleksiyRudenko, line 40:
```suggestion
  getValues() {
    return [...super.getValues, this.hands];
  }
```

OleksiyRudenko (changes requested):
@ORybka a hint in addition to the comment above

submissions/ORybka/post-a-tiny-js-world/index.js
ORybka: Ah! Spread operator! Of course! And function as an argument
@OleksiyRudenko  Thank you very much for the hint. This is a great solution!
I will use it now 😄

OleksiyRudenko: @ORybka There was a typo. Not as agrument, but rather a normal function call just as it is now in the code suggestion

ORybka: Yes, function call, thanks :)

OleksiyRudenko (approved):
@ORybka great job!

PR #292, opened 2020-12-16
Dom task

yaripey:
### DOM

[Demo](https://yaripey.github.io/menu-task/)
[Code base](https://github.com/yaripey/menu-task)

Please, review.


submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  let currentTab = document.querySelector('#about')
+  const mainElem = document.querySelector('main')
+  const toggler = document.querySelector('.toggle')
```
Aldegid, line 13: so what is more suitable, let or const?

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const changeText = ({ target }) => {
+    if (target.id !== '') {
```
Aldegid, line 16: u can use  `if (target.id)` instead

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+      if (target.id !== currentTab.id) {
+        currentTab.classList.toggle('focused')
+        mainElem.innerHTML = texts[target.id]
+        currentTab = document.querySelector(`#${target.id}`)
+        currentTab.classList.toggle('focused')
+        document.querySelector('aside').classList.toggle('toggled')
```
Aldegid, line 22: remove 'togled' class only when needed

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+  let currentTab = document.querySelector('#about')
+  const mainElem = document.querySelector('main')
+  const toggler = document.querySelector('.toggle')
+
+  const changeText = ({ target }) => {
+    if (target.id !== '') {
+      if (target.id !== currentTab.id) {
+        currentTab.classList.toggle('focused')
+        mainElem.innerHTML = texts[target.id]
+        currentTab = document.querySelector(`#${target.id}`)
```
Aldegid, line 20: `document.querySelector(`#${target.id}`)` -  this is the same as your `target`

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+        currentTab.classList.toggle('focused')
```
Aldegid, line 21: try adding the active class to the target element and remove from the rest

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+        document.querySelector('aside').classList.toggle('toggled')
+      }
+    }
+    target.blur()
```
Aldegid, line 25: why are you calling this method here?

Aldegid (changes requested):

submissions/yaripey/dom-task/js/main.js
Aldegid: Good work! See my comments, please.

submissions/yaripey/dom-task/js/main.js
yaripey: Const for elements that will lot change. currentTab is changing during execution, that's why it's let.

submissions/yaripey/dom-task/js/main.js
yaripey: I am calling this to drop focus from the button after a mouse click. I have styles defined for buttons in focus just in case someone would want to navigate the menu with keyboard. But I don't want those styles to remain to the button after a simple mouse click.

submissions/yaripey/dom-task/js/main.js
yaripey: Can you please describe why would it be better? Wouldn't it do more actions than just untoggling the current one and toggling the target one?

submissions/yaripey/dom-task/js/main.js
yaripey: I already check if the button that is pressed not the current one on the 17th string. Or did you mean something else?

submissions/yaripey/dom-task/js/main.js
yaripey: Replaced it, thanks.

submissions/yaripey/dom-task/js/main.js
yaripey: Replaced, thanks

submissions/yaripey/dom-task/js/main.js
Aldegid: ` document.querySelector('aside').classList.toggle('toggled')` you do this every time the `changeText` function is called. do you really need it?

submissions/yaripey/dom-task/js/main.js
Aldegid: the focus/blur event is triggered both on click and on pressing the tab key

submissions/yaripey/dom-task/js/main.js
Aldegid: Dont reassign `currentTab`
Remove `if (target.id !== currentTab.id)`, this is an extra check in the next approach
For removing active class: ```document.querySelector('button.focused').classList.remove('focused') ```
For adding active class: ```target.classList.add('focused')```

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
+// Texts
```
Aldegid, line 1: redundant comment.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
...
+}
```
Aldegid, line 8: Please keep data without html tags, only text and image src..
Create markup using `createElement`, `createDocumentFragment`, `append` method, etc.

submissions/yaripey/dom-task/index.html
```diff
@@ -0,0 +1,75 @@
...
+          <li><button id="finn">Finn the Human</button></li>
+          <li><button id="jake">Jake the Dog</button></li>
+          <li><button id="marcy">Marceline</button></li>
+          <li><button id="pb">Princess Bubblegum</button></li>
+        </ul>
+      </nav>
+    </aside>
+
+    <main>
+      Adventure Time is an American fantasy animated television series created by Pendleton Ward for Cartoon Network.
```
Aldegid, line 35: Why you need this, if you generate content by js?

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,38 @@
+// Texts
+let texts = {
```
Aldegid, line 2: const?

Aldegid (changes requested):

submissions/yaripey/dom-task/js/main.js
yaripey: But if I do this, wouldn't the page rerender everytime I press a button? Even if I press the button that is active right now?

submissions/yaripey/dom-task/js/main.js
yaripey: Fixed

submissions/yaripey/dom-task/index.html
yaripey: Removed it, thanks.

submissions/yaripey/dom-task/js/main.js
yaripey: Refactored, thanks for a good advice.

submissions/yaripey/dom-task/js/main.js
yaripey: Understood, thanks.

submissions/yaripey/dom-task/js/main.js
yaripey: Now I got what you meant, thanks.

submissions/yaripey/dom-task/js/main.js
yaripey: I don't, thanks for pointing this out.

submissions/yaripey/dom-task/js/main.js
yaripey: Turned out all those element were unnecessary, you were right, thanks.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -1,37 +1,101 @@
...
 
-  const changeText = ({ target }) => {
-    if (target.id) {
-      if (target.id !== currentTab.id) {
-        currentTab.classList.toggle('focused')
-        mainElem.innerHTML = texts[target.id]
-        currentTab = target
-        currentTab.classList.toggle('focused')
+  // Function for changing text in main
+  const changeText = (textPart) => {
```
Aldegid, line 49: I think that passing the target as an argument is better. This will make it possible to avoid unnecessary operations like this: ```const newCurrentButton = document.querySelector(`#${textPart}`)```

submissions/yaripey/dom-task/js/main.js
```diff
@@ -1,37 +1,101 @@
...
+    const currentMain = document.querySelector('main');
+
+    if (currentMain.classList[0] !== textPart) { // if this page isn't open already
+      const newElem = document.createElement('main')
+      newElem.setAttribute('class', textPart)
+
+      const newImage = document.createElement('img')
+      newImage.setAttribute('src', texts[textPart].img)
+      newImage.setAttribute('class', 'picture')
```
Aldegid, line 58: try to use [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for prepare markup

submissions/yaripey/dom-task/js/main.js
```diff
@@ -1,37 +1,101 @@
...
+
+      const newText = document.createDocumentFragment()
+      texts[textPart].text.forEach(text => {
+        const anotherP = document.createElement('p')
+        anotherP.textContent = text
+        newText.appendChild(anotherP)
+        newText.appendChild(document.createElement('br'))
```
Aldegid, line 65: use template string and `append` method instead  textContent and appendChild

submissions/yaripey/dom-task/js/main.js
```diff
@@ -1,37 +1,101 @@
...
+      })
+
+      newElem.appendChild(newImage)
+      newElem.appendChild(newText)
+      currentMain.parentNode.replaceChild(newElem, currentMain)
+
+      // Changing styles of the corresponding button
+      const currentButton = document.querySelector('.focused')
+      currentButton.classList.toggle('focused')
+      const newCurrentButton = document.querySelector(`#${textPart}`)
```
Aldegid, line 75: ```const newCurrentButton = document.querySelector(`#${textPart}`)``` as i said before, it's  the same as your target element

Aldegid (changes requested):

submissions/yaripey/dom-task/js/main.js
```diff
@@ -97,5 +83,6 @@ document.addEventListener('DOMContentLoaded', () => {
     document.querySelector('aside').classList.toggle('toggled')
```
Aldegid, line 83: [DRY](https://ru.wikipedia.org/wiki/Don%E2%80%99t_repeat_yourself)

Aldegid (changes requested):
Please, remove all comments from you final code

submissions/yaripey/dom-task/js/main.js
```diff
@@ -45,33 +45,23 @@ const texts = {
...
-      const currentButton = document.querySelector('.focused')
-      currentButton.classList.toggle('focused')
-      target.classList.toggle('focused')
-
-      // For mobile, closing the menu after button tap
-      if (document.querySelector('aside').classList.contains('toggled')) {
-        document.querySelector('aside').classList.toggle('toggled')
-      }
-    }
+    document.querySelector('aside').classList.remove('toggled')
```
Aldegid: why did you remove this logic? If you use the same selector several times, it is better to put it in a separate variable.I meant it.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -45,33 +45,23 @@ const texts = {
 document.addEventListener('DOMContentLoaded', () => {
   const toggler = document.querySelector('.toggle')
 
-  // Function for changing text in main
   const changeText = (target) => {
     const currentMain = document.querySelector('main');
 
-    if (currentMain.classList[0] !== target.id) { // if this page isn't open already
```
Aldegid: Why did you remove this check? Please see my previous comment and make this logic work.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -83,6 +73,5 @@ document.addEventListener('DOMContentLoaded', () => {
     document.querySelector('aside').classList.toggle('toggled')
   })
 
-  // This imitates click for the content on page when first opened.
   document.querySelector('#about').click()
```
Aldegid, line 76: I dont recommend using click emulation. Better to make the first item active by default.

Aldegid (changes requested):

submissions/yaripey/dom-task/js/main.js
yaripey: I didn't remove the logic. `.remove()` does almost the same as what my logic did, so there's no point in this `if` statement. But I still moved the selector in the constant, since I indeed use it several times in code, thanks for pointing that out.

submissions/yaripey/dom-task/js/main.js
yaripey: I removed the logic and shifted some responsibilities in functions, now page rerenders only if user hits a button that is different than what is focused right now.

submissions/yaripey/dom-task/js/main.js
yaripey: After shifting responsibilities in functions, now it was easy to render the first page right away, without `click`, thanks.

OleksiyRudenko:
@yaripey please remove a file that is not required for a review.

image

yaripey:
@OleksiyRudenko removed it.

submissions/yaripey/dom-task/index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Document</title>
+  <link rel="stylesheet" href="css/main.css">
+  <script src="js/main.js"></script>
```
MaxKalashnyk, line 9: Why did you place this script inside the `head` tag?

submissions/yaripey/dom-task/index.html
```diff
@@ -0,0 +1,41 @@
...
+</head>
+
+
+<body>
+
+  <header>
+    <a href="#" class="toggle"></a>
```
MaxKalashnyk, line 16: It's better to make this element as `button`. Semantically it will be more correct

submissions/yaripey/dom-task/index.html
```diff
@@ -0,0 +1,41 @@
...
+    <img class="logo" src="img/logo.png" alt="logo">
+  </header>
+
+  <div class="middle">
```
MaxKalashnyk, line 20: Please use more informative class names. `middle` is too generic name, to my mind.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,78 @@
...
+    ]
+  }
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  const toggler = document.querySelector('.toggle')
+  const aside = document.querySelector('aside')
+
+  const changeText = (id) => {
+    const currentMain = document.querySelector('main');
```
MaxKalashnyk, line 50: You can move this element to the global scope. It would help you to decrease DOM assertions count.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,78 @@
...
+    const newText = texts[id].text.map(paragraph => `<p>${[paragraph]}</p>`)
+    currentMain.innerHTML = `
+          <img class="picture" src="${texts[id].img}">
+          ${newText.join('\n')}
+      `
+  }
+
+  const handleChangeText = ({ target }) => {
+    if (target.id) {
```
MaxKalashnyk, line 59: you can create some variable for storing `target.id` value because I see you used it multiple times in your code.

MaxKalashnyk (changes requested):
Good job!
Check my comments below.

Also, I'd like to share my recommendation with you.
It's not a big blocker in this task. You can use your data structure for this current simple project.
But for future I'd recommend you making your `texts` object more clear and close to the real life data objects.
Now your data object is not so dynamic.
Maybe, this data structure would help you:
``` javascript
[
   {
     image: ''someImageSrc",
     text: "someText",
     id: "someId"
   }, ...
]
```

submissions/yaripey/dom-task/index.html
yaripey: Moved it to the bottom.

submissions/yaripey/dom-task/index.html
yaripey: Made it a button. Also added some styles to compensate button's default estyles.

submissions/yaripey/dom-task/index.html
yaripey: Changed this particular one, do you see any other classes too generic?

submissions/yaripey/dom-task/js/main.js
yaripey: Moved it to the body of the function that is called on `DOMContentLoaded`.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,79 @@
...
+
+document.addEventListener('DOMContentLoaded', () => {
+  const toggler = document.querySelector('.toggle')
+  const aside = document.querySelector('aside')
+  const currentMain = document.querySelector('main');
+
+  const changeText = (id) => {
+    const newText = texts[id].text.map(paragraph => `<p>${[paragraph]}</p>`)
+    currentMain.innerHTML = `
+          <img class="picture" src="${texts[id].img}">
```
MaxKalashnyk, line 53: Don't forget about adding the `alt` attribute to the images.

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,79 @@
...
+          ${newText.join('\n')}
+      `
+  }
+
+  const handleChangeText = ({ target }) => {
+    const requestedID = target.id
```
MaxKalashnyk, line 59: 👍🏻

submissions/yaripey/dom-task/js/main.js
```diff
@@ -0,0 +1,79 @@
...
+      if (currentButton.id !== requestedID) {
+        changeText(requestedID)
+        currentButton.classList.toggle('focused')
+        target.classList.toggle('focused')
+      }
+    }
+    aside.classList.remove('toggled')
+  }
+
+  const menu = document.querySelector('.main-navigation')
```
MaxKalashnyk, line 71: you can move this variable to the top (where you store other variables)

MaxKalashnyk (approved):
LGTM! I think we can merge this PR. 
Check my now comments as well. They are not very critical for approving this PR.

PR #294, opened 2020-12-17
Object-Oriented JS task

kt-std:
### Object Oriented JS

[Demo](https://kt-std.github.io/arcade-game/)
[Code base](https://github.com/kt-std/arcade-game)

Please, review.

### DIFF
--- a/submissions/kt-std/object-oriented-js/app.js
+++ b/submissions/kt-std/object-oriented-js/app.js
@@ -1,160 +1,162 @@
 const allEnemies = [],
-    ENEMIES_AMOUNT = 3,
-    ROW_NUM = 6,
-    COL_NUM = 7,
-    ROW_HEIGHT = 70,
-    Y_STEP = 70,
-    X_STEP = 70,
-    INITIAL_X = X_STEP * Math.floor(COL_NUM / 2),
-    INITIAL_Y = (ROW_NUM - 1) * Y_STEP,
-    CANVAS_START = 0,
-    CANVAS_WIDTH = COL_NUM * X_STEP,
-    CANVAS_HEIGHT = ROW_NUM * X_STEP + 20,
-    LIVES_AMOUNT = 3,
-    LIVE_SPRITE = "images/heart.svg",
-    LIVES_CONTAINER = document.getElementById("hearts"),
-    RANDOM_BASE = 100,
-    SPEED_INC = 0.01,
-    X_GAP = X_STEP / 2,
-    Y_GAP = Y_STEP / 2,
-    messages = {
-        win: {
-            text: `You've just successfully saved Xmas!`,
-            path: `images/bell.svg`,
-            button: `Play again!`,
-            heading: `Congrats!`,
-        },
-        lose: {
-            text: ``,
-            path: `images/bell.svg`,
-            button: `Try again!`,
-            heading: `You lose \u26C4`,
-        },
-    };
+	ENEMIES_AMOUNT = 3,
+	ROW_NUM = 6,
+	COL_NUM = 7,
+	ROW_HEIGHT = 70,
+	Y_STEP = 70,
+	X_STEP = 70,
+	INITIAL_X = X_STEP * Math.floor(COL_NUM / 2),
+	INITIAL_Y = (ROW_NUM - 1) * Y_STEP,
+	CANVAS_START = 0,
+	CANVAS_WIDTH = COL_NUM * X_STEP,
+	CANVAS_HEIGHT = ROW_NUM * X_STEP + 20,
+	LIVES_AMOUNT = 3,
+	LIVE_SPRITE = "images/heart.svg",
+	LIVES_CONTAINER = document.getElementById("hearts"),
+	RANDOM_BASE = 100,
+	SPEED_INC = 0.01,
+	X_GAP = X_STEP / 2,
+	Y_GAP = Y_STEP / 2,
+	messages = {
+		win: {
+			text: `You've just successfully saved Xmas!`,
+			path: `images/bell.svg`,
+			button: `Play again!`,
+			heading: `Congrats!`,
+		},
+		lose: {
+			text: ``,
+			path: `images/bell.svg`,
+			button: `Try again!`,
+			heading: `You lose \u26C4`,
+		},
+	};
 
-const Enemy = function (index) {
-    this.x = CANVAS_START;
-    this.y = Y_STEP * index;
-    this.speed = Math.random() * RANDOM_BASE;
-    this.sprite = "images/yeti.png";
-};
+class Entity {
+	constructor(x, y, sprite) {
+		this.x = x;
+		this.y = y;
+		this.sprite = sprite;
+	}
 
-/* This function displays each Enemy instance sprite on the canvas
- * due to its current X and Y coodinates
- */
-Enemy.prototype.render = function () {
-    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
-};
+	render() {
+		ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+	}
 
-Enemy.prototype.update = function (dt) {
-    // Speed up the enemy during the game
-    this.speed += SPEED_INC;
-    // Update Enemy's horizontal position
-    this.x = this.updateHorizontalPosition(this.x + dt * this.speed);
-};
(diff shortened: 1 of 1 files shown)


andrewklmn:
@kt-std  
Good job!
Your game looks very relevant for this time of year! )))
cool_stuff
Very cool solution with teleportation of Santa from one side of the game area to another.
And of course, the idea of three extra lives for the player is very nice!

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+    this.lives = LIVES_AMOUNT;
+    this.sprite = "images/santa.png";
+};
+
+/* This function displays Player sprite on the canvas
+ * due to its current X and Y coodinates
+ */
+Player.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
```
andrewklmn, line 76: let's move it to parent class...

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+        displayMessage("lose", "flex", messages);
+    } else {
+        allEnemies.forEach((enemy) => {
+            if (
+                Math.abs(this.x - enemy.x) < X_GAP &&
+                Math.abs(this.y - enemy.y) < Y_GAP
+            ) {
+                this.lives--;
+                removeLivesFromContainer();
+                resetPosition(this, INITIAL_X, INITIAL_Y);
```
andrewklmn, line 95: `
this.setPosition(INITIAL_X, INITIAL_Y);
`

what about this instead of global func resetPosition()?

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+/*
+ * This function resets Player and Enemies position, Player's lives
+ * and Enemies speed to restart the game if the reset button is pressed
+ */
+document.querySelector("body").addEventListener("click", function (e) {
+    if (e.target.id === "reset") {
+        displayMessage(e.target.offsetParent.id, "none");
+        resetPosition(player, INITIAL_X, INITIAL_Y);
+        player.resetLives(LIVES_AMOUNT);
+        allEnemies.map((enemy, index) => {
```
andrewklmn, line 169: do we really need .map() here? 
It creates a new array here, but this new array is not used in the code later.
maybe .forEach() will be better?

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+            resetPosition(enemy, CANVAS_START, Y_STEP * (index + 1));
```
andrewklmn, line 170: if we move this function to parent class Actor then we can use just like this:

``
enemy.setPosition(CANVAS_START, Y_STEP * (index + 1));
``

andrewklmn (changes requested):
@kt-std 
Very clean code! 
It's easy to read because your code is well-formatted. Thank you! 
Let's make this good code more object-oriented.

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+     * move right
+     */
+    return position > CANVAS_WIDTH + X_GAP ? CANVAS_START - X_GAP : position;
+};
+
+const Player = function () {
+    this.y = INITIAL_Y;
+    this.x = INITIAL_X;
+    this.lives = LIVES_AMOUNT;
+    this.sprite = "images/santa.png";
```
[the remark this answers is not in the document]
kt-std: Thank you for your time and reviewing this one! :)
I've fixed this part as you proposed. Sincerely hope, the code looks better and much DRYer now.

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+            resetPosition(enemy, CANVAS_START, Y_STEP * (index + 1));
+            enemy.speed = Math.random() * 100;
+        });
+    }
+});
+
+function resetPosition(object, positionX, positionY) {
+    object.x = positionX;
+    object.y = positionY;
+}
```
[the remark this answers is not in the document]
kt-std: I've updated the function naming

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,255 @@
...
+
+/*
+ * This function removes all lives from the LIVES_CONTAINER.
+ */
+function clearLivesContainer() {
+    while (LIVES_CONTAINER.firstChild) {
+        LIVES_CONTAINER.removeChild(LIVES_CONTAINER.lastChild);
+    }
+}
```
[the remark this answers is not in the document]
kt-std: Explicit usage of global variable was replaced to the function parameter as it was proposed. Thank you!

submissions/kt-std/object-oriented-js/app.js
kt-std: I've moved it to the Entity class

submissions/kt-std/object-oriented-js/app.js
kt-std: Not annoying at all! :)
Thanks a lot for your detailed review.

submissions/kt-std/object-oriented-js/app.js
kt-std: Nope, map is not needed as the array that function returns has never being assigned to any other variable. I've updated the code with forEach method.

andrewklmn (changes requested):
LGTM!

Great job! 
You listened to all my suggestions.
I approve.

andrewklmn:
@MaxKalashnyk 
@OleksiyRudenko 
@zonzujiro 

Could you check, approve, and merge this PR, please?

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,254 @@
...
+					Math.abs(this.x - enemy.x) < X_GAP &&
+					Math.abs(this.y - enemy.y) < Y_GAP
+				) {
+					this.lives--;
+					removeItemFromContainer(LIVES_CONTAINER);
+					this.setPosition(INITIAL_X, INITIAL_Y);
+				}
+			});
+		}
+	}
```
OleksiyRudenko, line 110: Separation of Concerns principle should be followed
(e.g. `update` method does only rendering and doesn't contain any unrelated **inline** code; for example collision check is defined as a dedicated method and only called from inside `update`)

submissions/kt-std/object-oriented-js/app.js
```diff
@@ -0,0 +1,254 @@
...
+	 * If Player still left lives and it has crossed the border of any Enemy,
+	 * then decrease Player lives amount, delete live item from the LIVES_CONTAINER
+	 * and reset Player position to the initial state.
+	 */
+
+	update() {
+		if (!this.lives) {
+			displayMessage("lose", "flex", messages);
+		} else {
+			allEnemies.forEach((enemy) => {
```
OleksiyRudenko, line 99: Classes should not refer to any global variables, like global variable `player`, which is an instance of `Player` class. Similar is applicable to any global variable.

Referring to global constants and globals provided by the gaming platform like `Resources` is OK.

Hint: pass a global variable that is not a constant or an engine-provided resource to a class' constructor, save it in class props and use it across class method.

OleksiyRudenko (changes requested):
@kt-std well done!

@andrewklmn thank you for the help. You did a great job!

However there is a couple important things to fix yet.

andrewklmn (approved):

submissions/kt-std/object-oriented-js/app.js
kt-std: @OleksiyRudenko, thank you for reviewing and giving vital advices! :) 
Fixed this one via adding one more method to Player class. I do hope the code is better now.

submissions/kt-std/object-oriented-js/app.js
kt-std: @OleksiyRudenko, thank you for the tip! I've added enemies array as a Player's class property to pass it inside the checkCollisions method.

OleksiyRudenko (approved):
@kt-std great job!

PR #300, opened 2020-12-17
Added solution for JS Post OOP task

Ant-C-tech:
### OOP exercise

[Demo](https://ant-c-tech.github.io/a-tiny-JS-world/)
[Code base](https://github.com/Ant-C-tech/a-tiny-JS-world/blob/populate-world/index.js)

Please, review.


submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+   Code repository: https://github.com/Ant-C-tech/a-tiny-JS-world
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitant {
+    constructor(name, gender, friends) {
+        //At the moment I can't imagine what the creatures will inhabit our world, that is why I assumed that these will be some people and different kinds of their pets. In this way I can assume that they all would have these common properties and one method to perform the task.
```
OleksiyRudenko, line 13: The OOP compliant code is easier to update when requirements change.
Stick to current requirements, i.e. make code good to meet them.

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+        super(...args);
+        this.legs = 4;
+    };
+
+    getVoice() {
+        return super.getVoice() + ` I have ${this.legs} legs.`;
+    };
+}
+
+class Human extends Inhabitant {
```
OleksiyRudenko, line 38: Humans are animals.
Properties that two child classes share should be generalized in common predecessor.

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+    constructor(saying, ...args) {
+        super(...args);
```
OleksiyRudenko, line 40: This is legal, but for code readability stick to named arguments and pass to a function only what is required.
This way flaws in hierarchy design will become more obvious.

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+        //I assumed that people can say different phrases
+        this.saying = saying;
+        this.species = 'human';
+        this.hands = 2;
+        this.legs = 2;
+    };
+
+    getVoice() {
+        return `I am a ${this.species}. ` + super.getVoice() + ` I have ${this.legs} legs and ${this.hands} hands. I want to say you: "${this.saying}".`;
+    };
```
OleksiyRudenko, line 50: 1. If you had more optimal classes design you might benefit from method overloading.
```
methodName() {
  return [super.methodName(), 'bar'].join(", "); // calls methodName from a parent class
}
```

2. Lengthy strings are optimizable. Example:
```
[ `My name is ${name}`,
  'I am X yrs old',
  'My friends are',
].join(". ");
```
This also makes maintenance easier. If we want to add a new fragment, we can easily spot a right place for it.

OleksiyRudenko (changes requested):
@Ant-C-tech well done.
Let's make it more OOP and SOLID compliant.

Ant-C-tech:
Thanks for you review and work with my code! 
According to your notice "Humans are animals..." I changed the predecessor of Human class to Animal instead of Inhabitant. To not lose meaning of exist class Animal, which now did not have any unique properties (like "legs" before), I decided to imagine that only Animal can have gender while Inhabitant can be a flower for example, without any gender. So I transfered "gender" property to Animal class.
I was acting assumed that inheritance mean create new child classes always when we have some new property that supplement the parent class. In this case (not in the real world) Human class not have to be the child of  Animal class. Or maybe that is because we assume appearing new class like the Flower for example.
Anyway I have just started to learn OOP so my knowledges don't let me to understand many things and I don't know how to do it in the right way so I did it in this way. Did I do it properly? 
It seems to me that cause of my difficulties with understanding this task was named in article "Composition over Inheritance" "People can't predict the future even if they feel like can". 
It is the most difficult task for me in the course even though it is very interesting.

OleksiyRudenko:
@Ant-C-tech 
> People can't predict the future even if they feel like can
Exactly. So we do not need to take into consideration what requirements **can be**, we need to stick to what requirements **are**. It will be easy to change well designed code when new requirements arrive.

Now, all of your animals have legs, just different number of those (like all ihabitants have names, just different). So, the property `legs` is shared and thus should be generalized into common predecessor, just lioke name.

Ant-C-tech:
Thank you for your explonation. I have tried to add property ''legs" for class Animal in the correct way. Please, give me a bit of your time again.

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,126 @@
...
+class Inhabitant {
+    constructor(name, friends) {
+        this.name = name;
+        this.friends = friends;
+        this.templete = `My name is ${this.name}`;
+    };
+
+    getVoice() {
+        let pluralEnding;
+        (this.friends && this.friends.length > 1) ? pluralEnding = 's': pluralEnding = '';
```
OleksiyRudenko, line 20: A typical misuse of a ternary.
Logic like `let variable; if (condition) { variable = vIfConditionTrue } else { variable = vIfConditionFalse }` translates into `let variable = condition ? vIfTrue : vIfFalse` using a ternary

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,126 @@
...
+        ].join('. ');
+    };
+}
+
+class Dog extends Animal {
+    constructor(gender, name, friends, legs) {
+        super(gender, name, friends, legs);
+        this.species = 'dog';
+        this.saying = 'Woof-Woof!';
+        this.legs = 4;
```
OleksiyRudenko, line 52: Why sending `legs` over to `Animal` via `super` call is not sufficient?

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,126 @@
...
+            super.getVoice(),
+            `I want to say you: "${this.saying}".`
+        ].join('. ');
+    };
+}
+
+class Cat extends Animal {
+    constructor(gender, name, friends, legs) {
+        super(gender, name, friends, legs);
+        this.species = 'cat';
```
OleksiyRudenko, line 66: All inhabitants have prop `species` just like they all have e.g. `name`.
Why handling `species` is different from the way you handle `name`?

OleksiyRudenko (changes requested):
@Ant-C-tech looks better.
However there are certain inconsistencies in handling the props.
Please apply every question to every prop in your code and fix as appropariate.

Ant-C-tech:
Thanks for your review and your patience!
When I was changing my code according your comment "A typical misuse of a ternary..." I have noticed that to form the correct message depending on the existence of an array of friends, I have already checked its existence.
I made a mistake when I re-nested this check in the ternary operator. I got rid of double check
when i moved the ternary operator below. I hope it had't make the reading of my code difficulty and I did all things in the right way.
Also following your comments, I moved the "species" and "legs" properties to the parent class "Inhabitant". I hope I am setting their values correctly in the child classes.

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,113 @@
...
+        this.name = name;
+        this.legs = legs;
+        this.friends = friends;
+        this.saying = saying;
+
+        this.templete = [`I am a ${this.species}`,
+            `My name is ${ this.name }`,
+            `I have ${this.legs} legs`,
+            `I want to say you: "${this.saying}"`
+        ].join('. ');
```
OleksiyRudenko, line 23: Not illegal but there may be an issue.
Imagine a property (e.g. `saying`) gets changed during the object life cycle and we want this change reflected when `getVoice` gets called next time.

`friends` prop has this privilege. Any good reason to handle friends and the group of "simpler" props differently?

Answer to this question leads to a proper approach.

submissions/Ant-C-tech/oop-exercise/index.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    getVoice() {
+        return (this.friends) ? [
+            this.templete,
+            `I have friend${(this.friends.length > 1) ? 's' : ''}: ${this.friends.join(', ')}`
+        ].join('. ') : [
+            this.templete,
+            'I am looking for friends'
+        ].join('. ');
```
OleksiyRudenko, line 33: This fragment, simplified, looks as:
```
return condition
  ? [variableA, stringifiedFriends].join()
  : [variableA, 'string literal'].join()
```

In fact `condition` manages one of the resulting arrays elements. So just move it there. This way scalability and maintainability get improved, and the code gets shorter as quite a number of repetitions get eliminated.

Example above refactored:
```
return [variableA, 
  condition ? stringifiedFriends : 'string literal'].join()
```
or added formatting for even better readability:
```
return [
  variableA, 
  condition
    ? stringifiedFriends 
    : 'string literal'
  ].join()
```

OleksiyRudenko (approved):
@Ant-C-tech well done!
Check the comment below as the learning.

submissions/Ant-C-tech/oop-exercise/index.js
OleksiyRudenko: Alternatively you might have had
```
const stringifiedFriends = ... all the magic with friends here...;
return [variableA, stringifiedFriends].join();
```

Ternaries inside template strings make reading painful

PR #309, opened 2020-12-18
OOP exercise

kt-std:
### OOP exercise

[Demo](https://kt-std.github.io/a-tiny-JS-world/)
[Code base](https://github.com/kt-std/kottans-frontend/tree/main/task_tiny_js_world)

Please, review.

### DIFF
--- a/submissions/kt-std/oop-exercise/index.js
+++ b/submissions/kt-std/oop-exercise/index.js
@@ -1,0 +1,115 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+  Complete the below for code reviewers' convenience:
+
+  Code repository: https://github.com/kt-std/kottans-frontend/tree/main/task_tiny_js_world
+  Web app: https://kt-std.github.io/a-tiny-JS-world/
+  */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+/** Class representing an inhabitant. */
+class Inhabitant {
+  /**
+   * Create an inhabitant
+   * @param {string} species - The species of the inhabitant.
+   * @param {string} name - The name of the inhabitant.
+   * @param {string} gender - The gender of the inhabitant.
+   * @param {number} legs - The amount of legs of the inhabitant.
+   * @param {string} sayings - The text of inhabitant greetings.
+   * @param {[]} friends - The list of inhabitant friends.
+   */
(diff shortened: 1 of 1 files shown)


submissions/kt-std/oop-exercise/index.js
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Check if an inhabitant has friends
+   * @param {[]} friends - The list of inhabitant friends
+   * @return {string} List of friends or sad quote
+   */
+  haveFriends(friends) {
+    return Array.isArray(friends) && friends.length ?
+      friends.join(', ') :
+      'Forever alone...';
+  }
```
OleksiyRudenko, line 40: Storing friends as a string has a drawback:
- adding more friends during lifetime will be comlicated
~- what if we want to print not only friends' names but their other properties?~
Nevermind you use strings as friends rather than objects.

Just call this method in stringify instead of just referring `this.friends`

submissions/kt-std/oop-exercise/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Stringifies an inhabitant properties
+   * @return {string} String consistings of an inhabitant
+properties separated with semicolon
+   */
+  stringify() {
+    return [this.species, this.name, this.gender, this.sayings,
+      this.friends, this.legs].join('; ');
+  }
```
OleksiyRudenko, line 50: Bonus learning.
`Object` offers standard magic method `toString`. It is `toString` that is responsible for `[Object object]` output in some use cases.
All objects implictly inherit from global `Object`.
`toString` is overloadable (as any object method).

It will be called implicitly in string context. `print` ensures string context. So `print(obj.stringify())` can be shortened to `print(obj)`

OleksiyRudenko (changes requested):
@kt-std great job!
One fix and one bonus learning

submissions/kt-std/oop-exercise/index.js
kt-std: Thank you so much for reviewing!  I've just fixed it.

submissions/kt-std/oop-exercise/index.js
kt-std: Wow :open_mouth:  Kinda magic! Lots of thanks for sharing! Guess, it's my St. Nicholas gift :)

OleksiyRudenko (approved):
@kt-std great job!

PR #316, opened 2020-12-19
Object-Oriented JS

vakulinina:
### Object-Oriented JS

[Demo](https://vakulinina.github.io/frontend-nanodegree-arcade-game/)
[Code base](https://github.com/vakulinina/frontend-nanodegree-arcade-game)

Please, review.

### DIFF
--- a/submissions/vakulinina/js-oop/app.js
+++ b/submissions/vakulinina/js-oop/app.js
@@ -36,9 +36,10 @@ Enemy.prototype.render = function () {
     ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
 };
 
-const Player = function (x, y) {
+const Player = function (x, y, enemies) {
     this.x = x;
     this.y = y;
+    this.enemies = enemies;
     this.sprite = 'images/char-princess-girl.png';
 };
 
@@ -47,7 +48,7 @@ Player.prototype.update = function () {
 };
 
 Player.prototype.checkCollisions = function () {
-    allEnemies.forEach(enemy => {
+    this.enemies.forEach(enemy => {
         if (this.y === enemy.y && this.x >= (enemy.x - enemySize * 0.7) && this.x <= enemy.x + enemySize) {
             this.x = playerStartX;
             this.y = playerStartY;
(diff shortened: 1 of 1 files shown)


submissions/vakulinina/js-oop/app.js
```diff
@@ -0,0 +1,117 @@
...
+    this.y = y;
+    this.sprite = 'images/char-princess-girl.png';
+};
+
+Player.prototype.update = function () {
+    this.checkCollisions();
+};
+
+Player.prototype.checkCollisions = function () {
+    allEnemies.forEach(enemy => {
```
OleksiyRudenko, line 50: Classes should not not refer to any global variables, like global variable `player`, which is an instance of `Player` class (referring to global constants and globals provided by the gaming platform like `Resources` is OK)

OleksiyRudenko (changes requested):
@vakulinina well done!
One important thing to fix and we're done.

submissions/vakulinina/js-oop/app.js
vakulinina: I'm not sure I completely understand what I should do here. Make `player` a constant?

OleksiyRudenko: @vakulinina 
`player` was referred to here just for an example. In this particular case "illegal" global variable is `allEnemies`.
Do you see a way to make it so that `Player` class dosn't refer to it directly?
In other words, how we can make an instance of `Player` learn about `allEnemies`?
Hint: how `Player` instance learns about its initial X and Y positions and wouldn't we use `playerStartX` for example directly in `Player` class' methods?

vakulinina: Check out the changes, please.
If it's correct, then I have one more question. 
If I assigned a global variable to a property in the constructor directly without passing it as an argument, would it be 'illegal' as well?
Like so: 
`const Player = function (x, y) {
    this.x = x;
    this.y = y;
    this.enemies = allEnemies;
};`

OleksiyRudenko: Yes, it gonna be illegal as well.
The legal way is just the same Player learns about its X and Y positions.

vakulinina: Thank you a lot. You did not approve the PR, is there anything else?

OleksiyRudenko (approved):
@vakulinina great job!

PR #335, opened 2020-12-22
JS OOP

Shrai-dev:
### JS OOP (Frogger game)

[Demo](https://shrai-dev.github.io/frontend-nanodegree-arcade-game/)
[Code base](https://github.com/Shrai-dev/frontend-nanodegree-arcade-game)

Please, review.


OleksiyRudenko:
@Shrai-dev you still refer to outdated submission instructions.
image

Not a big problem, just get used to a better git flow.

OleksiyRudenko (changes requested):
@Shrai-dev good job.

Please check yourself against the checklist below and confirm your code base is changed to comply.

 * Requirements re **OOP**:

submissions/Shrai-dev/frogger-game/app.js
```diff
@@ -0,0 +1,128 @@
...
+  allEnemies.forEach((enemy) => {
+    if (
+      this.x < enemy.x + spriteSize &&
+      this.x + spriteSize > enemy.x &&
+      this.y < enemy.y + spriteSize * 0.75 &&
+      this.y + spriteSize * 0.75 > enemy.y
+    ) {
+      this.x = playerStartX;
+      this.y = playerStartY;
+    }
```
OleksiyRudenko, line 74: image

OleksiyRudenko (changes requested):
@Shrai-dev pretty well done.
Please do not expect mentors to point at exactly every mistake, but rather deduce problematic areas in your code from principles and examples.

OleksiyRudenko:
Does the app work as expected?

Shrai-dev:
Now it works.

OleksiyRudenko (approved):
@Shrai-dev well done!

Please test your app at all times before requesting review.

PR #336, opened 2020-12-22
Memory pair game

kt-std:
### Memory Pair Game

[Demo](https://kt-std.github.io/memory-pair-game/)
[Code base](https://github.com/kt-std/memory-pair-game)

Please, review.

### DIFF
--- a/submissions/kt-std/memory-pair-game/js/script.js
+++ b/submissions/kt-std/memory-pair-game/js/script.js
@@ -98,10 +98,7 @@ function changeSize(card) {
 	}, 500);
 }
 
-function setOpacity(property, value, delay) {
-	if (delay === undefined) {
-		delay = 1000;
-	}
+function setOpacity(property, value, delay = 1000) {
 	setTimeout(() => (property.style.opacity = value), delay);
 }
 
@@ -128,10 +125,9 @@ function showWinMessage() {
 }
 
 function reset() {
-	const clearVars = [checkedCards, checkedCardsContainers, foundMatches].map(
-		(variable) => (variable = [])
-	);
-	[checkedCards, checkedCardsContainers, foundMatches] = clearVars;
+	checkedCards = [];
+	checkedCardsContainers = [];
+	foundMatches = [];
 	shuffledCards = shuffleArray(cardsData);
 	container.innerHTML = "";
 	document.querySelector(".win__container").style.display = "none";
--- a/submissions/kt-std/memory-pair-game/js/script.js
+++ b/submissions/kt-std/memory-pair-game/js/script.js
@@ -1,18 +1,13 @@
 const container = document.querySelector(".container"),
-	numCards = 9,
-	cardsData = ((assetsPath) => {
-		const cardsDataArray = [];
-		for (let i = 1; i <= numCards; i++) {
-			const card = {path: `${assetsPath}p${i}.svg`, name: `p${i}`};
-			cardsDataArray.push(card, card);
-		}
-		return cardsDataArray;
-	})("assets/");
-
-let checkedCards = [],
-	checkedCardsContainers = [],
-	shuffledCards = shuffleArray(cardsData),
-	foundMatches = [];
+	numCards = 9;
+let cardsDataInitial = Array(numCards)
+	.fill()
+	.map((_, i) => ({ path: `assets/p${i + 1}.svg`, name: `p${i + 1}` })),
+	checkedCards = [],
+	foundMatches = [],
+	checkedCardsContainers = [];
+const cardsData = cardsDataInitial.concat(cardsDataInitial);
+let shuffledCards = shuffleArray(cardsData);
 
 function shuffleArray(o) {
 	for (
--- a/submissions/kt-std/memory-pair-game/js/script.js
+++ b/submissions/kt-std/memory-pair-game/js/script.js
@@ -1,12 +1,12 @@
 const container = document.querySelector(".container"),
 	numCards = 9,
 	cardsData = ((assetsPath) => {
-		const arr = [];
+		const cardsDataArray = [];
 		for (let i = 1; i <= numCards; i++) {
 			const card = {path: `${assetsPath}p${i}.svg`, name: `p${i}`};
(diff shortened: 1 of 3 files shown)


submissions/kt-std/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,145 @@
+const container = document.querySelector(".container"),
+	numCards = 9,
+	cardsData = ((assetsPath) => {
+		const arr = [];
```
AMashoshyna, line 4: `arr` is a very generic name. Let's find a name that describes what this variable does.

submissions/kt-std/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,145 @@
...
+		for (let i = 1; i <= numCards; i++) {
```
AMashoshyna, line 5: let's try to use `Array.prototype.map` here

submissions/kt-std/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,145 @@
...
+	})("assets/");
+
+let checkedCards = [],
+	checkedCardsContainers = [],
+	shuffledCards = shuffleArray(cardsData),
+	foundMatches = [];
+
+function shuffleArray(o) {
+	for (
+		var j, x, i = o.length;
```
AMashoshyna, line 19: why `var`?

submissions/kt-std/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,145 @@
...
+		card.style.width = "0";
+		card.style.height = "0";
+		card.childNodes.forEach(
+			(cardInnerStyle) => (cardInnerStyle.style.display = "none")
+		);
+	}, 500);
+}
+
+function setOpacity(property, value, delay) {
+	if (delay === undefined) {
```
AMashoshyna, line 107: I would return the `return` value. Mutating function parameter is really confusing.

submissions/kt-std/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,145 @@
...
+	winMessage.appendChild(heading);
+	winMessage.appendChild(button);
+	winMessage.classList.add("win__container", "flex-center");
+	setOpacity(winMessage, 1);
+	document.body.appendChild(winMessage);
+}
+
+function reset() {
+	const clearVars = [checkedCards, checkedCardsContainers, foundMatches].map(
+		(variable) => (variable = [])
```
AMashoshyna, line 137: this looks weird. what does it do?

AMashoshyna (commented):
Good job! Please take a look at my comments

submissions/kt-std/memory-pair-game/js/script.js
kt-std: @AMashoshyna lots of thanks for your review especially during holidays season! I am so thankful!
I've updated the variable name with another one to clearly express what it actually does.

submissions/kt-std/memory-pair-game/js/script.js
kt-std: I’m afraid I’m a little confused how to do this as this function just creates a cardData array that contains "generated" names and pathes.

submissions/kt-std/memory-pair-game/js/script.js
kt-std: `var` is not needed, so I've updated this one with the `let`

submissions/kt-std/memory-pair-game/js/script.js
kt-std: This function clears arrays that were used during the game played after restart button was clicked.

submissions/kt-std/memory-pair-game/js/script.js
kt-std: Do we actually need to return it? This if statement is only used to set the default value of the delay if this parameter was omitted during the function call.

submissions/kt-std/memory-pair-game/js/script.js
kt-std: Would it be okay if I create an array of empty slots with dedicated length and then apply map to it to iterate through all its elements and return the new one?

kt-std: I've just updated the code with the map method. I do hope it is better now.

submissions/kt-std/memory-pair-game/js/script.js
AMashoshyna: I think you are looking for [default parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)

submissions/kt-std/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,140 @@
...
+	);
+	[checkedCards, checkedCardsContainers, foundMatches] = clearVars;
```
AMashoshyna, line 134: Why do you need these variables?

AMashoshyna: If you don't, then I would suggest to reset the arrays in a simpler manner, like so:
```javascript
checkedCards = []
checkedCardsContainers = []
foundMatches = []
```

submissions/kt-std/memory-pair-game/js/script.js
kt-std: @AMashoshyna, thank you for your time and helpful advises ! I've fixed it with the default parameter.

submissions/kt-std/memory-pair-game/js/script.js
kt-std: I've updated the code with simpler manner of resetting arrays.

kt-std:
@AMashoshyna, I am so sorry for disturbing you :bow:  If there is even a tiny possibility that you will have some free time, I would be unbelievably thankful for checking this PR
cat

AMashoshyna:
Sooo sorry! My bad that the review took so long. Well done!

PR #341, opened 2020-12-23
JS OOP

tavor118sn:
### Object Oriented JS

[Demo](https://tavor118sn.github.io/frontend-nanodegree-arcade-game/)
[Code base](https://github.com/tavor118sn/frontend-nanodegree-arcade-game)

Please, review.

### DIFF
--- a/submissions/tavor118sn/task_js_oop/app.js
+++ b/submissions/tavor118sn/task_js_oop/app.js
@@ -1,0 +1,146 @@
+const CELL = {
+  WIDTH: 101,
+  HEIGHT: 83,
+};
+
+const PLAYING_AREA = {
+  START_X: 0,
+  START_Y: 0,
+  END_X: 404,
+  END_Y: 500,
+};
+
+const PLAYER_INFO = {
+  SPRITE: 'images/char-boy.png',
+  START_X_POSITION: CELL.WIDTH * 2,
+  START_Y_POSITION: CELL.HEIGHT * 5 - 15,
+  WIDTH: 80,
+  HEIGHT: 40,
+};
+
+const ENEMY_Y_CORRECTION = 13;
+const ENEMY_INFO = {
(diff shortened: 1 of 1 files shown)


submissions/tavor118sn/task_js_oop/app.js
```diff
@@ -0,0 +1,146 @@
...
+    PLAYER_INFO.START_Y_POSITION,
+  );
+};
+
+Player.prototype = Object.create(Character.prototype);
+
+Player.prototype.update = function () {
+
+};
+
```
OleksiyRudenko, line 98: If every Character needs to have update and not all Characters need any sensible implementation thereof (no real operations inside the method), then it makes sense to have this "empty" method in `Character` class so that all Characters inherit it but only some of them re-implement it.

Not really required to fix here. Just keep in mind.

submissions/tavor118sn/task_js_oop/app.js
```diff
@@ -0,0 +1,146 @@
...
+  if (
+    this.x < this.player.x + PLAYER_INFO.WIDTH &&
+    this.x > this.player.x - ENEMY_INFO.WIDTH &&
+    this.y < this.player.y + PLAYER_INFO.HEIGHT &&
+    this.y > this.player.y - ENEMY_INFO.WIDTH
+  ) {
+    alert('You lose!');
+    this.player.reset();
+  }
+};
```
OleksiyRudenko, line 81: This method makes two notably different things:
1. Render a character icon
2. Check if any collisions occur

This breaks the SRP principle. No good.

OleksiyRudenko (changes requested):
@tavor118sn great job!
One important thing to fix though.

tavor118sn:
Thanks for review! Hope I understood comments properly.

OleksiyRudenko (approved):
@tavor118sn great job!

PR #345, opened 2020-12-23
Tiny js world | Anastasiia Yeromina

AsyaYeromina:
### Tiny JS world

[Demo](https://asyayeromina.github.io/a-tiny-JS-world/)
[Code base](https://github.com/AsyaYeromina/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/AsyaYeromina/Tiny-JS-world/index.js
+++ b/submissions/AsyaYeromina/Tiny-JS-world/index.js
@@ -1,0 +1,91 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/AsyaYeromina/a-tiny-JS-world
+   Web app: https://asyayeromina.github.io/a-tiny-JS-world/.
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+const dog = {
+   species: 'Canis',
+   name: 'Jessie',
+   gender: 'female',
+   legs: '4',
+   hands: '0',
+   speech: 'woof-woof!',
+   friends: ['Stepan', 'Ivanka']
+};
+
+const cat = {
+   species: 'Felis catus',
+   name: 'Python',
+   gender: 'male',
+   legs: '4',
+   hands: '0',
+   speech: 'Meow',
+   friends: []
+};
+
+const woman = {
+   species: 'Homo sapiens',
+   name: 'Ivanka',
+   gender: 'female',
+   legs: '2',
+   hands: '2',
+   speech: 'Hello!',
+   friends: ['Stepan', 'Python', 'Jessie']
+};
+
+const man = {
+   species: 'Homo sapiens',
+   name: 'Stepan',
+   gender: 'male',
+   legs: '2',
+   hands: '2',
+   speech: 'Hello!',
+   friends: ['Ivanka', 'Python', 'Jessie']
+};
+
+const catWoman = {
+   species: 'Felis sapiens',
+   name: 'Selina',
+   gender: 'female',
+   legs: '2',
+   hands: '2',
+   speech: cat.speech,
+   friends: ['Stepan', 'Python']
+};
+
+const inhabitants = [cat, dog, woman, man, catWoman];
+
+const getValues = function(obj) {
+   obj.friends = obj.friends.join(', ') || '0';
+   const string = Object.values(obj).join('; ');
+   return string;
+}
+
+inhabitants.forEach(el => print(getValues(el), 'div'));
+
+
+
+
+// ======== OUTPUT ========
+/* Use print(message) for output.
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
--- a/submissions/AsyaYeromina/Tiny-JS-world/index.js
+++ b/submissions/AsyaYeromina/Tiny-JS-world/index.js
@@ -59,15 +59,23 @@ const catWoman = {
 };
 
 const inhabitants = [cat, dog, woman, man, catWoman];
+const keys = ['species', 'name', 'gender', 'legs', 'hands', 'speech', 'friends'];
 
-const getValues = function(obj) {
-   obj.friends = obj.friends.join(', ') || '0';
-   const string = Object.values(obj).join('; ');
-   return string;
+const getValues = function(obj, arr) {
+   const stringArr = [];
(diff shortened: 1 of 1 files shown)


submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+   friends: ['Stepan', 'Python']
+};
+
+const inhabitants = [cat, dog, woman, man, catWoman];
+
+const getValues = function(obj) {
+   obj.friends = obj.friends.join(', ') || '0';
+   const string = Object.values(obj).join('; ');
+   return string;
+}
```
OleksiyRudenko, line 67: image

JFYI, `for...in` is a variation of `Object.keys()` and thus same restrictions applicable.

OleksiyRudenko (changes requested):
@AsyaYeromina nicely done

Please double-check the task requirements

submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,99 @@
...
+const getValues = function(obj, arr) {
+   const stringArr = [];
+   arr.forEach(key => {
+      const value = obj[key];
+      if (typeof value === 'object') {
+         stringArr.push(value.join(', '))
+      } else
+      stringArr.push(value);
+   });
+   return stringArr.join('; ');
```
OleksiyRudenko, line 73: Whenever you have a source array and resulting array of the same dimension, where each source array element translates into something else, then it is a use case for `Array.map`

Consider the following fragment:

```
['propNameA', 'propNameB']
  .map(propName => obj[propName]) // map keys onto values
  .map(value=> Array.isArray(value) ? value.join(", ") : value) // if any value is an array then merge its elements, otherwise leave as it is
  .join("; ");
```
The above `map`s could be merged. In a simple case it is acceptable. In complex cases it is better to have every step have its own distinct and readable operation.

submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+const inhabitants = [cat, dog, woman, man, catWoman];
+const keys = ['species', 'name', 'gender', 'legs', 'hands', 'speech', 'friends'];
+
+const getValues = function(obj, arr) {
+   const stringArr = [];
+   arr.forEach(key => {
+      const value = obj[key];
+      if (typeof value === 'object') {
```
OleksiyRudenko, line 68: This is an issue. If the value is not an iterable object (like literal objects above), then `value.join()` will throw an error.
The proper way to check if an entity is an array is to employ `Array.isArray(variable)`.

P.S. No need to handle cases where a value is a non-iterable object for now.

AsyaYeromina:
> @AsyaYeromina nicely done
> 
> Please double-check the task requirements

Sorry for misunderstanding, will check tasks twice before sending.

Please review changes in latest commit.

OleksiyRudenko (changes requested):
@AsyaYeromina well done!
If you didn't have a test of values against being any type of objects rather than against being an array specifically, then the other comment would be just for your information.

Please fix both issues.

submissions/AsyaYeromina/Tiny-JS-world/index.js
AsyaYeromina: I thought that `Array.isArray(variable)` can't be used in this home task
This is what I need. Added to code. Thanks

submissions/AsyaYeromina/Tiny-JS-world/index.js
AsyaYeromina: Please check the code with map in latest commit

submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,93 @@
...
+   name: 'Selina',
+   gender: 'female',
+   legs: '2',
+   hands: '2',
+   speech: cat.speech,
+   friends: ['Stepan', 'Python']
+};
+
+const inhabitants = [cat, dog, woman, man, catWoman];
+const keys = ['species', 'name', 'gender', 'legs', 'hands', 'speech', 'friends'];
```
OleksiyRudenko, line 62: `keys` is too generic name. What we've got here? - Inhabitants' property names.
So `inhabitantsPropertyNamesList` or `inhabitantsPropertyNames` or `propertyNames` would be better.

submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+const getValues = function(obj, arr) {
```
OleksiyRudenko, line 64: `getValues` is too generic.
What this function does? It composes an inhabitant's individual properties into the inhabitant's presentation. Something like `getInhabitantPresentation` or `build...` or `buildPresentation`.

`obj` and `array` are too generic.
What the function expects as its 1st argument? It is specialized to process inhabitants. So `inhabitant` tells a reader what the function expects here.

Not just any `array` is ok here but rather `propertyNames`.

This will also make use of IDE IntelliSense, when you start typing a function call and Intellisense gives you a hint on what are the function's arguments names are, so you know what params to pass and in what order. Imagine, if you had `function(arr1, arr2){...}` IntelliSense would give any sensible hint.

submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,93 @@
...
+   return arr.map(cur => obj[cur])
```
OleksiyRudenko, line 65: `cur` is not an obvious abbreviation. `propertyName` from a list of property names I suppose?

submissions/AsyaYeromina/Tiny-JS-world/index.js
```diff
@@ -0,0 +1,93 @@
...
+   .map(value => Array.isArray(value) ? value.join(", ") : value)
```
OleksiyRudenko, line 66: `value` is too generic. `propertyValue` I suppose?

OleksiyRudenko (approved):
@AsyaYeromina well optimized code! Approved

With reservations though.

Abbreviations and generic names are good for generic examples. Real code must read as much English as only possible. (`dog` for name you used is better than if we had `d` here instead)

Remember the code is written for other people to read. It's especially important in bigger tasks like Memory Pair Game and Friends that you may have ahead.

Small tasks are good to learn better practices.

Please take the comments below for consideration. That's how the code is read by an arbitrary reader that is not really well aware of the task itself.

submissions/AsyaYeromina/Tiny-JS-world/index.js
OleksiyRudenko: @AsyaYeromina what made you think so?

AsyaYeromina: At that time I didn't understand which methods can be used, and which shouldn't and why. 
It's a surprise for me now, that I couldn't comprehend this before, because the task description seems very clear now, and the job seems to me now very easy to complete.

AsyaYeromina:
> @AsyaYeromina well optimized code! Approved
> 
> With reservations though.
> 
> Abbreviations and generic names are good for generic examples. Real code must read as much English as only possible. (`dog` for name you used is better than if we had `d` here instead)
> 
> Remember the code is written for other people to read. It's especially important in bigger tasks like Memory Pair Game and Friends that you may have ahead.
> 
> Small tasks are good to learn better practices.
> 
> Please take the comments below for consideration. That's how the code is read by an arbitrary reader that is not really well aware of the task itself.

Thank you a lot for such detailed comments! Now I know how to name variables in real code. 
Although I've read some docs about naming before, I had the wrong idea about real naming. 
Maybe one of the reasons for that was that I saw a lot of code on educational platforms and some code examples from the internet. Variable names are usually short there which is comfortable for students in small training tasks but is a bad example for using it in real code.
I think I should read more real code and get used to writing the way you show it.

I'll change names in the next task, considering your comments. Thanks for your time!

PR #351, opened 2020-12-25
JS post-OOP

vakulinina:
### JS post-OOP

[Demo](https://vakulinina.github.io/a-tiny-JS-world/)
[Code base](https://github.com/vakulinina/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/vakulinina/js-post-oop/index.js
+++ b/submissions/vakulinina/js-post-oop/index.js
@@ -1,0 +1,56 @@
+class Inhabitant {
+   constructor(species, name, gender, legs, saying) {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.saying = saying;
+      this.legs = legs;
+      this.friends = [];
+   }
+
+   makeFriends(...friends) {
+      this.friends = [...friends];
+   }
+
+   getFriendsList(friends) {
+      return friends.length > 0 ? friends.map(friend => friend.name).join(', ') : 'no friends';
+   }
+
+   introduce() {
+      return [this.species, this.name, this.gender, this.legs, this.saying, this.getFriendsList(this.friends)].join('; ');
+   }
+}
+
+class Human extends Inhabitant {
+   constructor(name, gender, saying) {
+      super('human', name, gender, 2, saying);
+      this.hands = 2;
+   }
+
+   introduce() {
+      return [this.species, this.name, this.gender, this.legs, this.hands, this.saying, this.getFriendsList(this.friends)].join('; ');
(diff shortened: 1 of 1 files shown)


submissions/vakulinina/js-post-oop/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+class Human extends Inhabitant {
+   constructor(name, gender, saying) {
+      super('human', name, gender, 2, saying);
+      this.hands = 2;
+   }
+
+   introduce() {
+      return [this.species, this.name, this.gender, this.legs, this.hands, this.saying, this.getFriendsList(this.friends)].join('; ');
+   }
```
OleksiyRudenko, line 32: I see why you refer all the props here.
What would you change in order to have a scalable and maintainable implementation if allowed to expose properties in arbitrary order?
Imagine we need to establish dozens of species, so most likely we do not want every species `introduce` list all props as any change in the list of properties required for various species will require code changes in multiple classes and will quickly become a nightmare.

OleksiyRudenko (changes requested):
@vakulinina great job!
Just want to see if you're familiar with the concepts of ancestor's methods re-use.

submissions/vakulinina/js-post-oop/index.js
vakulinina: Thank you, I was struggling to come up with a neater solution, but failed 😄 
Hope new one is ok.

submissions/vakulinina/js-post-oop/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+class Human extends Inhabitant {
+   constructor(name, gender, saying) {
+      super('human', name, gender, 2, saying);
+      this.hands = 2;
+   }
+
+   getProperties() {
+      let properties = super.getProperties();
+      properties.splice(3, 0, this.hands);
```
OleksiyRudenko, line 36: This approach relies on an order of props and may easily become a bug when hierarchy grows.

OleksiyRudenko (approved):
@vakulinina good job!
I see why did you struggle. It was a compromise between objects' design and consistency of the output.

Your approaches are valid for this particular task and you exposed quite a plenty of creativity. 

However we have to keep maintainability and scalability at all times. No doubt you've taken your learnings from this expereince.

submissions/vakulinina/js-post-oop/index.js
vakulinina: Note taken. Thank you!

PR #353, opened 2020-12-25
JS OOP task

okkkko:
### JS OOP task

[Demo](https://okkkko.github.io/frontend-nanodegree-arcade-game/index.html)
[Code base](https://github.com/okkkko/frontend-nanodegree-arcade-game/tree/master/js)

Please, review.

### DIFF
--- a/submissions/okkkko/js_oop/app.js
+++ b/submissions/okkkko/js_oop/app.js
@@ -10,6 +10,7 @@ const MIN_SPEED = 100;
 const MAX_SPEED = 700;
 const CANVAS_WIDTH = 505;
 const BUG_SIZE = 70;
+const NUMBER_OF_ROWS = 3;
 
 const Character = function (x, y) {
   this.x = x;
@@ -76,15 +77,20 @@ Player.prototype.resetPosution = function () {
 };
 
 const player = new Player(PLAYER_START_POSITION.x, PLAYER_START_POSITION.y);
-const allEnemies = BUGS_START_POSITION.y.map(
-  (y) =>
-    new Enemy(
-      BUGS_START_POSITION.x,
-      y,
-      getRandomSpeed(MIN_SPEED, MAX_SPEED),
-      player
-    )
-);
+const enemyRows = (numberOfRows) =>
+  Array.from({ length: numberOfRows }, (_, i) => i + 1);
+const allEnemies = enemyRows(NUMBER_OF_ROWS).map((row) => {
+  return new Enemy(
+    BUGS_START_POSITION.x,
+    getBugsY(row),
+    getRandomSpeed(MIN_SPEED, MAX_SPEED),
+    player
+  );
+});
+
+function getBugsY(row) {
+  return row * STEP_Y - BUGS_PADDING;
+}
 
 function getRandomSpeed(min, max) {
   return Math.random() * (max - min) + min;
--- a/submissions/okkkko/js_oop/app.js
+++ b/submissions/okkkko/js_oop/app.js
@@ -66,7 +66,7 @@ Player.prototype.resetPosution = function(){
     this.x = PLAYER_START_POSITION.x;
 };
 
-let player = new Player(PLAYER_START_POSITION.x,PLAYER_START_POSITION.y);
+const player = new Player(PLAYER_START_POSITION.x,PLAYER_START_POSITION.y);
 BUGS_START_POSITION.y.forEach(function(y){
     allEnemies.push(new Enemy(BUGS_START_POSITION.x, y, getRandomSpeed(MIN_SPEED,MAX_SPEED),player))
  });
--- a/submissions/okkkko/js_oop/app.js
+++ b/submissions/okkkko/js_oop/app.js
@@ -1,11 +1,11 @@
 const PLAYER_START_POSITION = {x: 200, y: 380};
 const STEP_Y = 80;
-const STEP_X = 100;
+const STEP_X = 102;
 const FIRST_ROW = 60;
 const BUGS_START_POSITION = {x:-200, y: [FIRST_ROW, FIRST_ROW + STEP_Y, FIRST_ROW + STEP_Y*2, FIRST_ROW + STEP_Y*3]};
 const MIN_SPEED = 100;
 const MAX_SPEED = 700;
-const CANVAS_WIDTH = 500;
+const CANVAS_WIDTH = 505;
 const BUG_SIZE =70;
 
 const Enemy = function(x,y,speed,player){
--- a/submissions/okkkko/js_oop/app.js
+++ b/submissions/okkkko/js_oop/app.js
@@ -1,8 +1,8 @@
-const PLAYER_START_POSUTION = {x: 200, y: 380};
+const PLAYER_START_POSITION = {x: 200, y: 380};
 const STEP_Y = 80;
 const STEP_X = 100;
 const FIRST_ROW = 60;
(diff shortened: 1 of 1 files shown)


submissions/okkkko/js_oop/app.js
```diff
@@ -0,0 +1,87 @@
...
+    }if (key === "up" && this.y>0){
+        this.y-=STEP_Y;
+    };
+};
+Player.prototype.resetPosution = function(){
+    this.y = PLAYER_START_POSUTION.y;
+    this.x = PLAYER_START_POSUTION.x;
+};
+
+let player = new Player(PLAYER_START_POSUTION.x,PLAYER_START_POSUTION.y);
```
babayK0, line 69: why you use `let` here ?

submissions/okkkko/js_oop/app.js
```diff
@@ -0,0 +1,87 @@
+const PLAYER_START_POSUTION = {x: 200, y: 380};
+const STEP_Y = 80;
+const STEP_X = 100;
+const FIRST_ROW = 60;
+const BUGS_START_POSUTION = {x:-200, y: [FIRST_ROW, FIRST_ROW + STEP_Y, FIRST_ROW + STEP_Y*2, FIRST_ROW + STEP_Y*3]};
```
babayK0, line 5: What you will do if you will need 100500 bugs, you will add y position one by one to this array ?
Maybe made some function which will calculate `y` position for all bugs automatically ?

babayK0 (changes requested):
Good job!

First of all i really liked that you try to separate logic of update and collisions :+1: 

But i have some comments which i think can improve your code:
 1. Format your code. You can use tool like [prettier](https://prettier.io/)
 2. [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  Properties common for some classes you can generalized into a base class (e.g. there is Character base class, which is extended by Enemy and Player classes)
 3.  Watch my comments below)

submissions/okkkko/js_oop/app.js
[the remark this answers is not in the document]
okkkko: yeah, maybe that is too much :)

submissions/okkkko/js_oop/app.js
okkkko: just wasn't attentive enough, it should be const here

submissions/okkkko/js_oop/app.js
```diff
@@ -0,0 +1,87 @@
...
+const allEnemies = [];
+const MIN_SPEED = 100;
+const MAX_SPEED = 700;
+const CANVAS_WIDTH = 500;
```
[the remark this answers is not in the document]
okkkko: indeed, it is 505

submissions/okkkko/js_oop/app.js
okkkko: yes, I`ll fix it

okkkko:
@babayK0 Thanks for the review!

submissions/okkkko/js_oop/app.js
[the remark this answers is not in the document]
babayK0: not fixed. why bug is on grass ?)
Screenshot from 2020-12-28 16-01-14

submissions/okkkko/js_oop/app.js
```diff
@@ -1,87 +1,102 @@
...
 const STEP_Y = 80;
-const STEP_X = 100;
-const FIRST_ROW = 60;
-const BUGS_START_POSUTION = {x:-200, y: [FIRST_ROW, FIRST_ROW + STEP_Y, FIRST_ROW + STEP_Y*2, FIRST_ROW + STEP_Y*3]};
-const allEnemies = [];
+const STEP_X = 102;
+const BUGS_PADDING = 18;
+const BUGS_START_POSITION = {
+  x: -200,
+  y: [1, 3, 4].map((rowNumber) => rowNumber * STEP_Y - BUGS_PADDING),
```
babayK0, line 7: You didn't properly understand me. 
 
this must not be array of positions, because this variable don't want to know how many bugs there will be in game


just function which accept rowNumber and returns y position. It will be more abstract.

submissions/okkkko/js_oop/app.js
```diff
@@ -1,87 +1,102 @@
...
+  this.y = PLAYER_START_POSITION.y;
+  this.x = PLAYER_START_POSITION.x;
 };
 
-let player = new Player(PLAYER_START_POSUTION.x,PLAYER_START_POSUTION.y);
-BUGS_START_POSUTION.y.forEach(function(y){
-    allEnemies.push(new Enemy(BUGS_START_POSUTION.x, y, getRandomSpeed(MIN_SPEED,MAX_SPEED),player))
- });
+const player = new Player(PLAYER_START_POSITION.x, PLAYER_START_POSITION.y);
+const allEnemies = BUGS_START_POSITION.y.map(
```
babayK0, line 79: Here will be better to do map from rows

Also you can do automatic rows calculation)
example 
```
const enemyRows = numberOfRows => Array.from({ length: numberOfRows }, (_, i) => i + 1); 
// will return array of length you provide and fill it with numbers 1 to numberOfRows 

const allEnemies = enemyRows(10).map(row => {
   return new Enemy( BUGS_START_POSITION.x, BUGS_START_POSITION.y(row) .....)
})

```

babayK0 (changes requested):

submissions/okkkko/js_oop/app.js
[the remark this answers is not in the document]
okkkko: Understood) I just missed that there should be no bug on the grass

submissions/okkkko/js_oop/app.js
```diff
@@ -3,13 +3,13 @@ const STEP_Y = 80;
 const STEP_X = 102;
 const BUGS_PADDING = 18;
 const BUGS_START_POSITION = {
-  x: -200,
-  y: [1, 3, 4].map((rowNumber) => rowNumber * STEP_Y - BUGS_PADDING),
+  x: -200
 };
```
babayK0: If this just X coordinate why do you need store it in object ?
And it is not *bugs* this is *bug* variable (single *bug*) So, change naming.

submissions/okkkko/js_oop/app.js
```diff
@@ -3,13 +3,13 @@ const STEP_Y = 80;
...
 const MIN_SPEED = 100;
 const MAX_SPEED = 700;
 const CANVAS_WIDTH = 505;
 const BUG_SIZE = 70;
+const NUMBER_OF_ROWS = 3;
```
babayK0, line 12: Of what rows ?)

submissions/okkkko/js_oop/app.js
```diff
@@ -3,13 +3,13 @@ const STEP_Y = 80;
 const STEP_X = 102;
 const BUGS_PADDING = 18;
 const BUGS_START_POSITION = {
-  x: -200,
-  y: [1, 3, 4].map((rowNumber) => rowNumber * STEP_Y - BUGS_PADDING),
+  x: -200
 };
 const MIN_SPEED = 100;
 const MAX_SPEED = 700;
 const CANVAS_WIDTH = 505;
```
babayK0, line 10: But what you will do if width change ? you will need to change this value. If you have access to dynamically calculated value - use it instead of this one.
 
[Old comment](https://github.com/kottans/frontend-2021-homeworks/pull/353#discussion_r549315352)

babayK0 (changes requested):
Some minor changes needed, and i will approve it

babayK0 (approved):
Nice work !

okkkko:
Thanks!)

PR #362, opened 2020-12-26
add js-dom homework

kasionio:
### js-dom

[Demo](https://kasionio.github.io/js-dom/#done)
[Code base](https://github.com/kasionio/js-dom)

Please, review.


submissions/kasionio/js-dom/index.js
```diff
@@ -0,0 +1,79 @@
...
+const main = document.querySelector('#main');
+const nav = document.querySelector('#nav');
+
+planets.map(item => item.name)
+    .forEach(planet => {
+        let li = document.createElement('li');
+        li.className = 'nav_item';
+        li.innerHTML = `<a href='#'>${planet}</a>`;
+        const nav = document.getElementById('nav');
+        nav.appendChild(li);
```
Aldegid, line 54: Adding elements to DOM from a loop is a bad practice. A browser will run reflow and repaint for every element in the loop. Instead, you can: 1. Use append method, which can add several elements in one operation 2. Create some wrapper, add your items to the wrapper and then add it to DOM. It will be one operation

submissions/kasionio/js-dom/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+const main = document.querySelector('#main');
+const nav = document.querySelector('#nav');
+
+planets.map(item => item.name)
+    .forEach(planet => {
+        let li = document.createElement('li');
+        li.className = 'nav_item';
+        li.innerHTML = `<a href='#'>${planet}</a>`;
+        const nav = document.getElementById('nav');
```
Aldegid, line 53: `const nav = document.querySelector('#nav');` and then `const nav = document.getElementById('nav');`, why?

submissions/kasionio/js-dom/index.js
```diff
@@ -0,0 +1,79 @@
...
+        <figcaption class='figcaption'>${item.name} planet view</figcaption>
+    </figure>
+    </div>
+    <p>${item.description}</p>`;
+    return article;
+}
+
+nav.addEventListener('click', (event) => {
+    main.appendChild(createArticle(planets.find(item => item.name === event.target.innerText)));
+    main.removeChild(main.childNodes[0]);
```
Aldegid, line 78: Avoid using constructions like childNodes[0] children[0], firstElementChild, parentNode, nextSibling, etc.
In this case you can clear the content of the main before inserting it with a new one.

submissions/kasionio/js-dom/index.js
```diff
@@ -0,0 +1,79 @@
...
+    <figure>
+        <img class='image' src='${item.image}' alt='${item.name}'/>
+        <figcaption class='figcaption'>${item.name} planet view</figcaption>
+    </figure>
+    </div>
+    <p>${item.description}</p>`;
+    return article;
+}
+
+nav.addEventListener('click', (event) => {
```
Aldegid, line 76: you can use ES6 features and destructure `target` from event `({target})`

Aldegid (changes requested):

submissions/kasionio/js-dom/index.js
kasionio: Excuse me, my bad. The second one is redundant

submissions/kasionio/js-dom/index.js
kasionio: `let navLinks = planets.map(planet => '<li class ='nav_item'><a href='#'>${planet.name}</a></li>')
.join(''); `

`nav.innerHTML = navLinks;`

This works. Isn't this a bad practice?

submissions/kasionio/js-dom/index.js
kasionio: insert `main.textContent = '';` before appending a child

submissions/kasionio/js-dom/index.js
Aldegid: in this task, this is acceptable, but I recommend using  `createDocumentfragment` and an `append` method

submissions/kasionio/js-dom/index.js
Aldegid: `innerHTML` looks better

submissions/kasionio/js-dom/index.js
```diff
@@ -0,0 +1,74 @@
...
+        <figcaption class='figcaption'>${item.name} planet view</figcaption>
+    </figure>
+    </div>
+    <p>${item.description}</p>`;
+    return article;   
+}
+
+nav.addEventListener('click', ({target}) => {
+   main.textContent = '';
+   main.appendChild(createArticle(planets.find(item => item.name === target.innerText)));
```
Aldegid, line 73: you rerender the content of `main` every click, even if the active menu item does not change. Can you please fix it?

Aldegid (changes requested):

submissions/kasionio/js-dom/index.js
kasionio: Implemented checking. At least tried

Aldegid:
Well done!

PR #371, opened 2020-12-28
Task Friends app added

BlueLamantine:
### Task Friends app

[Demo](https://bluelamantine.github.io/kottans-community-app/)
[Code base](https://github.com/BlueLamantine/kottans-community-app)

Please, review.


submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
+class Render {
+  constructor(results, info) {
+    this.data = results;
```
zonzujiro, line 3: Give less generic name to the variable - https://learn.javascript.ru/ninja-code

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+        <input type="radio" id="male" name="sort-gender">
+        <label for="male" class="radio-lable">Male</label>
+        <input type="radio" id="female" name="sort-gender">
+        <label for="female" class="radio-lable">female</label>
+        <input type="radio" id="allGender" name="sort-gender">
+        <label for="allGender" class="radio-lable">all</label>
+        </div>
+      </div>
+
+      <div class="filter filter-name">
```
zonzujiro, line 121: Maybe you should wrap it into `fieldset`?

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+
+      
+      <div class="filter filter-search">
+        <p>Search user by email</p>
+        <input id="nameInput" type="text" size="30" class="searchEmail icon search">
+      </div>
+    `;
+  }
+
+  getFilters() {
```
zonzujiro, line 153: You called function `get` but it will return nothing :)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+  constructor() {
+    this.profilesContainer = document.querySelector('.profile-cards');
+    new Render().getFilters();
+    this.filtersPanel = document.querySelector('#filters');
+    this.searchInput = document.querySelector('#nameInput');
+    this.startFilter();
+    this.sortByNameTypes = {
+      type1: 'nameAZ',
+      type2: 'nameZA',
+      type3: 'nameAll',
```
zonzujiro, line 168: Magic numbers in variable names are bad. Please, name it more descriptive

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+    };
+  }
+
+  unckeckedRadioButton(nodeName) {
+    const activeRadioButtons = document.querySelectorAll(
+      'input[type=radio]:checked'
+    );
+    const checkedRadio = Array.from(activeRadioButtons).find(
+      node => node.name === nodeName
+    );
```
zonzujiro, line 183: You can replace it with one `querySelector` ;)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+  selectSort(selected) {
+    const profileNames = document.querySelectorAll('.user-mail');
+    this.searchInput.addEventListener('input', ({ target }) => {
+      let filter = target.value.toUpperCase();
+      profileNames.forEach(nameNode => {
+        if (nameNode.textContent.toLocaleUpperCase().startsWith(filter)) {
+          nameNode.closest('.user-card').hidden = false;
+        } else {
+          nameNode.closest('.user-card').hidden = true;
+        }
```
zonzujiro, line 196: You working on DOM itself which is crunchy. Better try to re-render array of filtered users.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+        this.unckeckedRadioButton('sort-age');
+        this.sortByContent('.first-name', selected.id, this.sortByNameTypes);
+      },
+      'sort-age': () => {
+        this.unckeckedRadioButton('sort-name');
+        this.sortByContent('.age', selected.id, this.sortByAgeTypes);
+      },
+      default: () => false,
+    };
+    return (sort[selected.name] || sort['default'])();
```
zonzujiro, line 212: `sort['default']` no need for array notation here

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+  sortBySelector(selectorName) {
+    const cards = document.querySelectorAll('.user-card');
+    if (selectorName === 'all') {
+      return Array.from(cards).sort((a, b) => a.id.localeCompare(b.id));
+    }
+    return Array.from(cards).sort((a, b) =>
+      a
+        .querySelector(selectorName)
+        .textContent.localeCompare(b.querySelector(selectorName).textContent)
+    );
```
zonzujiro, line 255: Same, you changing DOM itself and you should change array of sources. And then render it again

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+        this.insertSortedProfiles(sortedCards, 'firstChild'),
+      [sortTypes.type3]: () =>
+        this.insertSortedProfiles(this.sortBySelector('all'), 'lastChild'),
+      default: () => false,
+    };
+    return (typeOfSort[sortType] || typeOfSort['default'])();
+  }
+}
+
+class Api {
```
zonzujiro, line 273: `API` :)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  async getData(pageNum = 1) {
+    try {
+      const response = await fetch(
+        this.apiURL + `?page=${pageNum}&results=10&seed=abc`
+      );
+      if (!response.ok) throw new Error(errorMessage);
+      const data = await response.json();
+      new Render(data.results, data.info).getProfiles();
```
zonzujiro, line 286: Ok, you are recreating this class only to call its methods.

And class itself is stateless.

So you can replace it with singleton and pass instance as an argument to the constructor using dependency injection.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+      document.querySelector('.loading').classList.remove('active');
+    } catch (err) {
+      console.error(err);
+      alert(err);
```
zonzujiro, line 290: Please, no :) Not alerts

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+class Observer {
+  constructor() {
+    this.startPage = 1;
+    this.loading = () => {
+      if (
+        window.scrollY + window.innerHeight >=
+        document.documentElement.scrollHeight - 5
+      ) {
+        document.querySelector('.loading').classList.add('active');
+        new Api().getData(++this.startPage);
```
zonzujiro, line 304: Nope, dependency injection here also :)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+      window.removeEventListener('scroll', this.loading);
+      return;
+    }
+    window.addEventListener('scroll', this.loading);
+  }
+}
+
+new Api().getData();
+new Filter();
+const loadMoreData = new Observer();
```
zonzujiro, line 323: Why class `Observer` has instance named `loadMoreData`?

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+loadMoreData.loadDataByScroll();
+setTimeout(() => {
+  document.querySelector('#preload').remove();
+  document.querySelector('body').style.setProperty('--scroll', 'auto');
+  document.querySelector('.wrapper').classList.remove('hidden');
+}, 7000);
```
zonzujiro, line 329: And what will be if 7 seconds are pass and data will not be loaded? :)

Do not use such approaches - you don't know how much time downloading of data will take. You can start load and `then` do what you need.

zonzujiro (changes requested):

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, got it!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: thanks for a tip, now it is ```<form>``` with ```fieldsets```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: renamed to ```renderFilters()```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: my bad, already have renamed!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: thanks for a tip! Now with querySelector much better!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Got it! According to your recommendation I save data to array and then work with it, sorted and then render

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Got it! According to your recommendation I save data to array and then work with it, sorted and then render

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Now it is called class ```Users```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: ok

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: I removed from my app option to load more data with scroll. That's not necessary.

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: I feel like this ```new Users(new Render(), new Filter())``` might be the one. (dependency injection)

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok. got it!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Point of this function is to provide two kind of sort. One includes different sorts (by name, by age...) and second with search by email. So when we choose ```search``` our ```const sort``` used  variety ```default```: () => false. ```sort['default']``` is necessary,  that will be error without it!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: This chunk of code responsibility for preload animation only. I need 7 seconds for right rendering text and image that removed this from page after render. Downloading of data is parallel with animation in this case. There was not desire to hide downloading, it was just for render animation at the beginning, before we see database of users.
Maybe I need to swap this parts of code. First might be rendering of animation and than downloading data (now is opposite) ?

BlueLamantine:
@zonzujiro thanks for review! I followed your recommendations!

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: Sorry, I have confused you. I meant bracket notation - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors

Also, if I understand you correctly, default sort is by email?

If so, why function called default then?

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: API was fine. It's an abbreviation, so it should be written with Caps Lock

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: But then it will be not singleton.

How many filters you have in your app?

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: Looks like this is problem with names.

If you need this 7 seconds only because of animation you should:

1) save 7000 into constant with proper name
2) Separate animation and loading of data into separate functions
3) Call them inside function which will initialise your application

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Just a moment's inattention.  Now it is ```return (sort[selected.name] || sort.sortByEmail)();```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, I got it. My goal is a have single instance of class and use it every time I need. 
So I realized singleton of ```Filter``` and ```Render``` classes with this construction :
```class ClassName {
constructor() {

    if (ClassName.instance == null) {
    ...
     ClassName.instance = this;
    }
    return ClassName.instance;
}

File: submissions/BlueLamantine/friends app/js/main.js (hunk 14 above)
Comment (BlueLamantine) [reply submissions/BlueLamantine/friends app/js/main.js]: Agreed, I separated this to another ```class Animation```

**Comment by BlueLamantine** [comment]
@zonzujiro Changes to my app has been added! Hope, singleton pattern works fine!

File: submissions/BlueLamantine/friends app/js/main.js (hunk 14 above)
Comment (zonzujiro) [reply submissions/BlueLamantine/friends app/js/main.js]: And why it should be class? 🤔 Class is overkill for this. You can just use function

File: submissions/BlueLamantine/friends app/js/main.js [hunk 15]
```diff
@@ -0,0 +1,339 @@
...
+        usersData.results;
+      this.renderData.renderProfiles();
+      this.filterData.startFilter();
+    } catch (err) {
+      console.error(err);
+    }
+  }
+}
+
+class Observer {
```
zonzujiro, line 314: Same, why class and not plain function?

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+    this.errorMessage = `Uh oh, something has gone wrong. Please tweet us @randomapi about the issue. Thank you.`;
+  }
+
+  async getUsersData() {
+    try {
+      const response = await fetch(this.apiURL + `?results=100&seed=abc`);
+      if (!response.ok) throw new Error(errorMessage);
+      const usersData = await response.json();
+      this.renderData.usersDataForRender = this.filterData.usersDataForFilter =
+        usersData.results;
```
zonzujiro, line 305: Not a best way to assign variables :) Possible, but not very comfortable to read.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+        this.sortedUsers = [...this.usersDataForFilter];
+      },
+    };
+    return selectGender[gender]();
+  }
+
+  sortByContent(sortType, sortTypes, sortBy) {
+    if (this.sortedUsers.length !== 0)
+      this.usersDataForFilter = [...this.sortedUsers];
+    else this.sortedUsers = [...this.usersByDefault];
```
zonzujiro, line 261: JavaScript has brackets - `{}`. Please, use them.

They make code more readable

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+      },
+      sortByEmail: () => false,
+    };
+    return (sort[selected.name] || sort.sortByEmail)();
+  }
+
+  startFilter() {
+    this.usersByDefault = [...this.usersDataForFilter];
+    this.filtersPanel.addEventListener('click', ({ target }) => {
+      if (target.type === 'radio' || target.type === 'text') {
```
zonzujiro, line 228: Why you need this checks?

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+      },
+      'sort-name': () => {
+        this.unckeckedRadioButton('sort-age');
+        this.sortByContent(selected.id, this.sortByNameTypes, 'name');
+      },
+      'sort-age': () => {
+        this.unckeckedRadioButton('sort-name');
+        this.sortByContent(selected.id, this.sortByAgeTypes, 'age');
+      },
+      sortByEmail: () => false,
```
zonzujiro, line 220: I don't see any reason why it should be in this object.

All keys in object named in `kebab-case`, but this method is in `camelCase`.
Also, all keys in your object can be selected by `selected.name` except this one.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  unckeckedRadioButton(nodeName) {
+    const checkedRadio = document.querySelector(
+      `input[type=radio][name=${nodeName}]:checked `
+    );
+    if (checkedRadio) checkedRadio.checked = false;
+  }
+
+  selectSort(selected) {
```
zonzujiro, line 192: To general name. Selected what?

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+        inAscending: 'nameAZ',
+        inDescending: 'nameZA',
+        byDefault: 'nameAll',
+      };
+      this.sortByAgeTypes = {
+        inAscending: 'age09',
+        inDescending: 'age90',
+        byDefault: 'ageAll',
+      };
+      Filter.instance = this;
```
zonzujiro, line 180: And why you are doing this? What problem are you solving?

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok. Now it is ```function animation()```. That was just my desire to follow one code style. Like 'only classes used, no mix with functions'

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: I removed this class and move it's only one function to hide/show menu to constructor of class Render.

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Please, give me a hint for another way to assign the same value to different variable. I used this method and prettier format my code to one line in this place. But, I will try make It comfortable to read.

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: ok. fixed

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: ```sortByEmail: () => false``` removed. All works fine without it

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Now it is ```selectSortMethod(selectedRadioButton)```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: That is a singleton pattern's prinsep (i think, not sure). I removed it because it is unnecessary here, only one instanse of class Filter in the app at all

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,329 @@
...
+      type3: 'nameAll',
+    };
+    this.sortByAgeTypes = {
+      type1: 'age09',
+      type2: 'age90',
+      type3: 'ageAll',
+    };
+  }
+
+  unckeckedRadioButton(nodeName) {
```
zonzujiro, line 184: `unchecked`? Install spellchecker into your code editor :)

Also, name of method should be verb

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+      inDescending: 'age90',
+      byDefault: 'ageAll',
+    };
+  }
+
+  unckeckedRadioButton(nodeName) {
+    const checkedRadio = document.querySelector(
+      `input[type=radio][name=${nodeName}]:checked `
+    );
+    if (checkedRadio) checkedRadio.checked = false;
```
zonzujiro, line 188: Please use curly brackets - it is easier to read then one-liners

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  selectSortMethod(selectedRadioButton) {
+    const name = selectedRadioButton.name;
+    const sortMethod = {
+      'sort-gender': () => {
+        this.unckeckedRadioButton('sort-name');
```
zonzujiro, line 195: Right now this is a magic strings.

Please, put `'sort-gender'` etc into constants and then reuse them.

Also, you should name your arguments in function. ` this.unckeckedRadioButton('sort-age')` does not describes anything.

` this.unckeckedRadioButton({buttonName: 'sort-age' })` does

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+    `;
+  }
+
+  renderFilters() {
+    this.filtersContainer.insertAdjacentHTML('beforeend', this.getHtmlFilter());
+  }
+}
+
+class Filter {
+  constructor(render) {
```
zonzujiro, line 165: `render` - verb, but this is not a function. You need noun.

So `renderer`.

Please, pay attention to variable names

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  startFilter() {
+    this.usersByDefault = [...this.usersDataForFilter];
+
+    this.searchInput.addEventListener('input', ({ target }) => {
+      let usersByEmail = [];
+      let filter = target.value.toUpperCase();
+
+      this.usersDataForFilter.forEach(user => {
```
zonzujiro, line 222: Use `Array#filter` :)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+    };
+    return sortMethod[name]();
+  }
+
+  startFilter() {
+    this.usersByDefault = [...this.usersDataForFilter];
+
+    this.searchInput.addEventListener('input', ({ target }) => {
+      let usersByEmail = [];
+      let filter = target.value.toUpperCase();
```
zonzujiro, line 220: It is not a `filter`, it's `userInput`. Filter - is entity which filters something.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+        if (user.email.toUpperCase().startsWith(filter)) {
+          usersByEmail.push(user);
+        }
+      });
+      this.renderDataWithFilters.usersDataForRender = usersByEmail;
+      this.renderDataWithFilters.renderProfiles();
+    });
+
+    this.filtersPanel.addEventListener('click', ({ target }) => {
+      if (target.type === 'radio') {
```
zonzujiro, line 232: Radio input will fire change event when clicked and this even will bubble. You can handle it.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+        this.renderDataWithFilters.usersDataForRender = this.usersDataForFilter;
+        this.renderDataWithFilters.renderProfiles();
+      }
+    });
+  }
+
+  sortByGender(gender) {
+    this.usersDataForFilter = [...this.usersByDefault];
+    const selectGender = {
+      allGender: () => {
```
zonzujiro, line 243: `allGenders`

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+    } else {
+      this.sortedUsers = [...this.usersByDefault];
+    }
+
+    const typeOfSort = {
+      [sortTypes.byDefault]: () => {
+        this.usersDataForFilter = [...this.sortedUsers];
+      },
+      [sortTypes.inAscending]: () => {
+        if (sortBy == 'name')
```
zonzujiro, line 275: Magic strings

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  sortByContent(sortType, sortTypes, sortBy) {
+    if (this.sortedUsers.length !== 0) {
+      this.usersDataForFilter = [...this.sortedUsers];
+    } else {
+      this.sortedUsers = [...this.usersByDefault];
+    }
+
+    const typeOfSort = {
```
zonzujiro, line 270: `typesOfSort`

zonzujiro (changes requested):
More comments. Almost done :)

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: Yes it is. You are right.

But usually no one bothers with handling situation when constructor was called twice in FE. Creating of one instance usually enough.

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: Possible you can keep users in one place?

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: My bad, fixed!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: My bad, I've already installed spellchecker and renamed method to ```uncheckRadioButton```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: ok, I put all names of sort methods in ```const sortMethods```, and named arguments in function like you said

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, renamed and got it!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, fixed.

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, renamed.

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, used ```addEventListener('change', function (e) {})``` in this case

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, renamed!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Created ```const sortIndicators``` for this

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, renamed!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: I thought not to use global variable in this case. 
Problem that I have in data appearance after instances of classes created (classes use this data and this entails an error). 
For me solution that I have is acceptable, couse I've tried different ways, but inconclusively.

BlueLamantine:
@zonzujiro thanks for chance to improve my general understanding of making suitable code!

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,352 @@
...
+  constructor() {
+    if (Render.instance == null) {
+      this.profilesContainer = document.querySelector('.profile-cards');
+      this.filtersContainer = document.querySelector('#filters');
+      document.querySelector('#menu').addEventListener('click', () => {
+        document.querySelector('#filters').classList.toggle('active');
+      });
+      Render.instance = this;
+    }
+    return Render.instance;
```
zonzujiro, line 11: As I told, YAGNI. In this task can call constructor only once

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,352 @@
...
+      byName: 'sort-name',
+    };
+    this.sortByNameTypes = {
+      inAscending: 'nameAZ',
+      inDescending: 'nameZA',
+      byDefault: 'nameAll',
+    };
+    this.sortByAgeTypes = {
+      inAscending: 'age09',
+      inDescending: 'age90',
```
zonzujiro, line 188: Why not call them `ascending` and `descending`? In such a case you will have general logic to check.

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,352 @@
...
+      byDefault: 'ageAll',
+    };
+    this.sortByGenderTypes = {
+      male: 'male',
+      female: 'female',
+    };
+  }
+
+  uncheckRadioButton(node) {
```
zonzujiro, line 197: `uncheckSelectedFilter` better, IMO

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,352 @@
...
+          selectedRadioButton.id,
+          this.sortByAgeTypes,
+          this.sortIndicators.age
+        );
+      },
+    };
+    return sortMethod[methodName]();
+  }
+
+  startFilter() {
```
zonzujiro, line 234: You not starting filter process, you add listeners :)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,352 @@
...
+      },
+      [sortTypes.inDescending]: () => {
+        if (sortBy == this.sortIndicators.name)
+          this.usersDataForFilter.sort((a, b) =>
+            b.name.first.localeCompare(a.name.first)
+          );
+        if (sortBy == this.sortIndicators.age)
+          this.usersDataForFilter.sort((a, b) =>
+            b.dob.age.toString().localeCompare(a.dob.age.toString())
+          );
```
zonzujiro, line 306: Sorting logic the same as at lines 289-296. Only the order of arguments is changed.

You can extract it into two functions and pass arguments as you need in them.

submissions/BlueLamantine/friends app/js/main.js
zonzujiro: You can keep users in this class and have getters and setters for them.

Having one source of truth is very important :)

submissions/BlueLamantine/friends app/js/main.js
```diff
@@ -0,0 +1,352 @@
...
+      this.renderData.renderProfiles();
+      this.filterData.startFilter();
+    } catch (err) {
+      console.error(err);
+    }
+  }
+}
+
+function animation() {
+  const animationDuration = 7000;
```
zonzujiro, line 339: As far as I remember this is duration of intro? If so, change name please

zonzujiro (changes requested):

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: ok, renamed to ```animationDurationOfIntro```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, fixed it!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, renamed this method to ```addListenersForFilters() ```

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: let it be, renamed

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: ok, renamed!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: fixed!

submissions/BlueLamantine/friends app/js/main.js
BlueLamantine: Ok, I've done like you said with getter and setter

zonzujiro:
LGTM

PR #372, opened 2020-12-28
Js post oop

Harnytskyi:
### OOP exercise

[Demo](https://harnytskyi.github.io/a-tiny-JS-world)
[Code base](https://github.com/Harnytskyi/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/Harnytskyi/js-post-oop/index.js
+++ b/submissions/Harnytskyi/js-post-oop/index.js
@@ -1,0 +1,83 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://harnytskyi.github.io/a-tiny-JS-world
+   Web app: https://github.com/Harnytskyi/a-tiny-JS-world
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+class Inhabitant {
+    constructor(species, name, gender, saying) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.saying = saying;
+    };
+    toString() {
+        return [
+            this.saying,
+            `I am a ${this.species}.`,
+            `My name is ${this.name}.`,
+            `My gender is ${this.gender}.`
+        ].join(' ');
+    }
+}
+class Human extends Inhabitant {
+    constructor(name, gender, saying) {
+        super('human', name, gender, saying);
+        this.legs = 2;
+        this.hands = 2;
+    };
+    toString(){
+        return `{${super.toString()} I have ${this.legs} legs and ${this.hands} hands.`;
+    }
+}
+class Man extends Human {
+    constructor(name, saying) {
+        super(name, 'male', saying);
+        
+    }
+}
+class Woman extends Human {
+    constructor(name, saying) {
+        super(name, 'female', saying);
+    }
+}
+class Mammal extends Inhabitant{
+    constructor(species, name, gender, saying){
+        super(species, name, gender, saying);
+        this.paws = 4;
+    };
+    toString(){
+        return `{${super.toString()} I have ${this.paws} paws.`;
+    }
+}
+
+class Dog extends Mammal {
+    constructor(name, gender) {
+        super('dog', name, gender, 'Gav');
+    }
+}
+class Cat extends Mammal {
+    constructor(name, gender) {
+        super('cat', name, gender, 'Meow');
+    }
+}
+
+const man = new Man('John', 'Hello, everybody!');
+const woman = new Woman('Jessica', 'Good morning, gays!');
+const cat = new Cat('Tom', 'male');
+const dog = new Dog('Jerry', 'female');
+
+// ======== OUTPUT ========
+/* Use print(message) for output.
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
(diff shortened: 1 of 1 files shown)


OleksiyRudenko (changes requested):
@Harnytskyi well done.

Must be a bug
image

Perhaps it would be avoided if subclasses' `toString` method followed the pattern of `Inhabitant#toString`

P.S. It's a pity you opted for skipping `friends` feature.

submissions/Harnytskyi/js-post-oop/index.js
```diff
@@ -0,0 +1,104 @@
...
+    };
+    getListFriends(){
+        if (this.friends.length == 0){
+            return "I have not a friends."
+        }
+        else{
+            var listFriends = [];
+            this.friends.map(element => {
+                listFriends.push(element.name);
+            })
```
OleksiyRudenko, line 25: This way of using `Array#map` is exactly the way to use `Array#forEach`.
The beauty of `Array#map` is that it as able to transform every item in inbound array. It will make code shorter and better readable.

Check the following materials to master this powerful method. It will come to a rescue in many cases.

- https://dev.to/sarah_chima/the-javascript-array-map-method-44m0
- https://www.javascripttutorial.net/javascript-array-map/

submissions/Harnytskyi/js-post-oop/index.js
```diff
@@ -0,0 +1,104 @@
...
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+class Inhabitant {
+    constructor(species, name, gender, saying, friends = []) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.saying = saying;
+        this.friends = friends;
```
OleksiyRudenko, line 15: Use `addFriends` method here. This way if any changes in handling of inbound friends list occur you'd need to change handling rules in a single place - in `addFriends`.

submissions/Harnytskyi/js-post-oop/index.js
```diff
@@ -0,0 +1,104 @@
...
+/* Use print(message) for output.
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+const inhabitants = [man, woman, cat, dog];
+inhabitants.forEach(key => print(key));
```
OleksiyRudenko, line 103: `key` is too generic name. What `inhabitants` array contains? The answer will produce a more appropriate name.

OleksiyRudenko (changes requested):
@Harnytskyi well done
Let's polish it.

submissions/Harnytskyi/js-post-oop/index.js
Harnytskyi: If I understood you correctly, it should look like `this.friends = this.addFriends();` ?

OleksiyRudenko: Well, not exactly this way. As far as i understand the signature of the method it requires an argument and returns nothing. Just use it the way the method should be used, just within the constructor.

submissions/Harnytskyi/js-post-oop/index.js
```diff
@@ -7,28 +7,23 @@
...
-    constructor(species, name, gender, saying, friends = []) {
+    constructor(species, name, gender, saying) {
         this.species = species;
         this.name = name;
         this.gender = gender;
         this.saying = saying;
-        this.friends = friends;
+        this.addFriends = function (friends){
+            this.friends = friends;
+        };
```
OleksiyRudenko, line 17: > Just **use** it the way the method should be used per your design, just within the constructor.

Note verb **use**. Not **define**.

Bring `addFriends` method definition back. Bring `friends` argument back to the constructor signature - why wouldn't we allow assign friends at construction time?

Use the `addFriends` method to initialize friends in constructor.

Feel free asking questions.

OleksiyRudenko (changes requested):
@Harnytskyi need to make a step back on friends handling and try different approach.

submissions/Harnytskyi/js-post-oop/index.js
Harnytskyi: Do you mean, when creating an instance, "friends" are assigned?
And creating an instance would look like this:
`const man = new Man('John', 'Hello, everybody!', [woman, dog]);` ?

OleksiyRudenko: The way you use `addFriends` directly in your code is as follows:
```
man.addFriends([woman, dog]);
woman.addFriends([cat]);
dog.addFriends([man]);
```

OleksiyRudenko: Both options are basically valid.
One can add friends (those that already exist) at a new object instantiation. And then at any time after.

OleksiyRudenko: BTW, current implementation is rather `assignFriends` as it won't add friends at the next call but rather replace them.

OleksiyRudenko (changes requested):
@Harnytskyi let's have a single point of new friends handling.

Harnytskyi:
> @Harnytskyi let's have a single point of new friends handling.

`this.addFriends(friends);`

OleksiyRudenko (approved):
@Harnytskyi well done!

PR #377, opened 2020-12-29
add first OOP version of tiny world

Roophee:
### a tiny JS world OOP

[Demo](https://github.com/Roophee/js-tyni-world-noOOP/tree/oop-world)
[Code base](https://github.com/Roophee/frontend-2021-homeworks/tree/a-tiny-js-world-oop/submissions/Roophee/a-tiny-js-world-oop)

Please, review.

### DIFF
--- a/submissions/Roophee/a-tiny-js-world-oop/index.js
+++ b/submissions/Roophee/a-tiny-js-world-oop/index.js
@@ -1,85 +1,65 @@
 "use strict";
 
-class Animalia {
-    alive = true;
-    canSaying = false;
-    constructor(gender, legs, tail) {
-        this.legs = legs;
-        this.tail = tail;
+class Organism {
+    constructor(name, gender, saying, legs,hands, tail ){
+        this.name = name;
         this.gender = gender;
+        this.species = this.constructor.name.toLowerCase();
+        this.saying = saying;
+        this.legs = legs;
+        this.hands = hands;
+        this.tail = tail;
+        this.friends = [];
+    }
+
+    addFriends(...arrayOfFriends) {
+        arrayOfFriends.forEach(({name}) => this.friends.push(name));
+    }
+
+    description(){
+        let descriptionArray = [this.name, this.gender,this.species, this.saying, this.legs, this.hands, this.tail];
+        this.friends.length === 0? descriptionArray.push('none'):descriptionArray.push(this.friends);
+        return descriptionArray;
     }
 }
 
-class Mammalia extends Animalia {
-    canSaying = true;
-    gregarious = true;
-    description (arrayOfProperties) {
-        return arrayOfProperties.map(property =>this[property]).join(';')
-    }
-    constructor(gender, friends, legs, tail) {
-        super(gender, legs, tail);
-        this.friends = friends.split(',');
-
+class Dog extends Organism {
+    static saying = 'Woof-Woof!';
+    constructor(name, gender){
+        super(name, gender, Dog.saying, 4,0,1);
     }
 }
 
-class HomoSapiens extends Mammalia {
-    species = "human";
-    hasName = true;
-    constructor(name, gender, friends, sayingText) {
-        super(gender, friends);
-        this.name = name;
-        this.legs = 2;
-        this.hands = 2;
-        this.tail = 'none';
-        this.saying = sayingText;
+class Cat extends Organism {
+    static saying = 'Meow!';
+    constructor(name, gender){
+        super(name, gender, Cat.saying, 4,0,1);
     }
 }
 
-class Dog extends Mammalia {
-    species = "dog";
-    hasName = true;
-    static saying = "Woof-Woof!";
-    constructor(name, gender, friends) {
-        super(gender, friends);
-        this.legs = 4;
-        this.hands = 'none';
-        this.tail = 1;
-        this.name = name;
-        this.saying = Dog.saying;
+class Human extends Organism {
+    static saying = 'Cogito ergo sum!';
+    constructor(name, gender,saying){
+        saying? super(name, gender, saying, 2,2,0):super(name, gender, Human.saying, 2,2,0);
     }
 }
 
-class Cat extends Mammalia {
-    species = "cat";
-    hasName = true;
-    static saying = "Meow!";
-    constructor(name, gender, friends) {
-        super(gender, friends);
-        this.legs = 4;
-        this.hands = 'none';
-        this.tail = 1;
-        this.name = name;
-        this.saying = Cat.saying;
+class WomanCat extends Organism {
+    constructor(name, gender){
+        super(name, gender, Cat.saying, 2,2,0);
     }
 }
 
-class CatWoman extends HomoSapiens {
-    species = "cat-woman";
-    catSaying() {
-        return Cat.saying;
-    }
-    constructor(name, gender, friends) {
-        super(name, gender, friends);
-        this.saying = this.catSaying();
-    }
-}
-const properties = ['species', 'name', 'gender', 'legs', 'hands', 'tail', 'saying', 'friends'],
-    man = new HomoSapiens('Sam', 'male', 'Hanna, Milo', 'Honey, I\'m home'),
-    woman = new HomoSapiens('Hanna', 'female', 'Sam, Milo, Kitty', "I\'m traveling down the river."),
-    dog = new Dog('Milo', 'male', 'Sam, Hanna'),
-    cat = new Cat("Kitty", 'female', ''),
-    catWoman = new CatWoman('Susanna', 'female', 'Sam, Hanna, Milo, Kitty'),
-    tinyWorld = [man, woman, dog, cat, catWoman];
+const dog = new Dog('Milo','male'),
+    cat = new Cat('Kitty','female'),
+    man = new Human('Michael', 'male'),
+    woman = new Human('Helga', 'female', 'Per aspera ad astra!'),
+    hero = new WomanCat('Sara', 'female'),
+    population = [man, woman, dog, cat, hero];
 
-print(tinyWorld.map(item => item.description(properties)).join('\n'))
+dog.addFriends(man, woman);
+man.addFriends(woman);
+woman.addFriends(man, cat, dog);
+hero.addFriends(man, woman, cat, dog);
+
+population.forEach(item =>print(item.description().join(';')));
--- a/submissions/Roophee/a-tiny-js-world-oop/index.js
+++ b/submissions/Roophee/a-tiny-js-world-oop/index.js
@@ -1,7 +1,7 @@
 "use strict";
 
 class Organism {
-    constructor(name, gender, saying, legs,hands, tail ){
+    constructor(name, gender, saying, legs, hands, tail) {
         this.name = name;
         this.gender = gender;
         this.species = this.constructor.name.toLowerCase();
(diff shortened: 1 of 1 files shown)


Roophee:
This is OOP versioned of task. My pre-OOP task is approved and merged. Please, change the label. Thanks.

lenadgit:
could you please change a link on DEMO? currently it redirects to the task description

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,85 @@
...
+        this.saying = sayingText;
+    }
+}
+
+class Dog extends Mammalia {
+    species = "dog";
+    hasName = true;
+    static saying = "Woof-Woof!";
+    constructor(name, gender, friends) {
+        super(gender, friends);
```
OleksiyRudenko, line 44: `super` here calls constructor of `Mammalia` that generously offers to send it also legs and tail.
Why not using it?
Same is applicable to many other props.

OleksiyRudenko (changes requested):
@Roophee nice try.
However not much of benefits OOP offers are employed.
OOP/SOLID principles help building more scalable and maintanable code.
Currently classes exist very much in isolation, features are not used, too much code repetition.
You can do better. Please reconsider the structure.
Yet I like the use of the static props. These are helpful when used wisely. Please keep those.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+class Organism {
+    constructor(name, gender, saying, legs, hands, tail) {
+        this.name = name;
+        this.gender = gender;
+        this.species = this.constructor.name.toLowerCase();
+        this.saying = saying;
+        this.legs = legs;
+        this.hands = hands;
+        this.tail = tail;
```
OleksiyRudenko, line 11: Not all inhabitants have hands and/or tail by natural design. Not that they have zero of those but rather do not have at all. Imagine how overwhelmed would be this class if we add wings for birds, horns for rhinos etc

The beauty of OOP design is that it allows every class inherit what is shared and introduce what is unique.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,65 @@
...
+        this.friends = [];
+    }
+
+    addFriends(...arrayOfFriends) {
+        arrayOfFriends.forEach(({ name }) => this.friends.push(name));
+    }
+
+    description() {
+        let descriptionArray = [this.name, this.gender, this.species, this.saying, this.legs, this.hands, this.tail];
+        this.friends.length === 0 ? descriptionArray.push('none') : descriptionArray.push(this.friends);
```
OleksiyRudenko, line 21:
```suggestion
        let descriptionArray = [this.name, 
          this.gender,
          this.species,
          this.saying,
          this.legs,
          this.hands,
          this.tail,
          this.friends.length === 0 ? 'none' : this.friends,
        ];
```

But beware: a class cannot handle properties it doesn't own or inherit from ancestors.
In good OOP design a class handles only props it owns.
Method overloading is to the rescue.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,65 @@
...
+    static saying = 'Meow!';
+    constructor(name, gender) {
+        super(name, gender, Cat.saying, 4, 0, 1);
+    }
+}
+
+class Human extends Organism {
+    static saying = 'Cogito ergo sum!';
+    constructor(name, gender, saying) {
+        saying ? super(name, gender, saying, 2, 2, 0) : super(name, gender, Human.saying, 2, 2, 0);
```
OleksiyRudenko, line 43:
```suggestion
        super(name, gender, saying || Human.saying, 2, 2, 0);
```

or even more obvious (as default values are still worth using)

```
constructor(name, gender, saying = 'Cogito ergo sum!') {
  super(name, gender, saying, 2, 2, 0);
}
```
and you won't need `static saying`.
Perhaps, super's constructor signature will change as humans have "zero" tails, while actually do not have any.

OleksiyRudenko (changes requested):
@Roophee please consider revisiting the [original project's materials](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up)

You will be interested in basic OOP principles, inhertiance and encapsulation in particular, as well as method overloading.

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: I rewrote code and add to each property a label with name of property, because it will be difficult to understand what mean property value on method overloading and length of returned string will be different depending on class. I think, it will be more recognizably.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,74 @@
+"use strict";
+
+class Inhabitant  {
+    constructor(name, gender, saying, legs=4, tail= 1 ){
+        this.name = name;
+        this.gender = gender;
+        this.species = this.constructor.name.toLowerCase();
```
OleksiyRudenko, line 7: Nice touch. Only 1 in 100 digs deep enough to use this.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+    description(){
+        let descriptionArray = ['name: '+ this.name,
+            'gender: '+ this.gender,
+            'species: '+ this.species,
+            'saying: '+ this.saying,
+            'legs: '+ this.legs,
+            'tail: '+ this.tail,
+            'friends: '+ (this.friends.length === 0? 'none' : this.friends),
+        ];
```
OleksiyRudenko, line 26: Example:
```
['propA', 'propB'].map(propName => `${propName} = ${obj[propName}`)
```

You may still need to add handling of array props in the map callback. Trust you can. Hint: a ternary in place of `` ${obj[propName} `` with a condition `Array.isArray(x) && x.length === yyy`

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+class Human extends Inhabitant  {
+    static saying = 'Cogito ergo sum!';
+    constructor(name, gender,saying, legs= 2, tail = 'none'){
+        super(name, gender, saying || Human.saying, legs, tail)
+        this.hands = 2;
+    }
+    description() {
+        return [super.description(),'hands: '+this.hands].join(';');
```
OleksiyRudenko, line 52: Delimiter was `; ` (with a trailing space) one other time. Worth having a typical delimiter as a constant?

OleksiyRudenko (changes requested):
@Roophee great job!
Let's polish it.

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: Done.  It's clever approach. Thanks for it.
I decide not remove static properties, it is not necessary, but in my mind, it gives useful access to saying property not only for class instance.

OleksiyRudenko: A note: Ternary operator is not a replacement for `if...else`. It can be used exactly in a place you need.
So `saying ? super(name, gender, saying, 2, 2, 0) : super(name, gender, Human.saying, 2, 2, 0)` at least translates into `super(name, gender, saying ? saying : Human.saying, 2, 2, 0)`. The latter then makes `saying || Human.saying` pretty obvious.

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: Don't understand what for it needs?
You mean I should rewrite method that create string of properties and it had to create string not in way ofnamed enumeration of properties, but in way of automatic enumeration of properties by Array methods?

Right?

OleksiyRudenko: Right. Imagine you need to add dozens of props in the code base created by someone else (other you)

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: Done. Save delimiter in const and use it in join methods.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,68 @@
...
+        console.log(this.constructor)
+    }
+
+    addFriends(...arrayOfFriends) {
+        arrayOfFriends.forEach(({name}) => this.friends.push(name));
+    }
+
+    description(){
+        return Object.entries(this).map(property => `${property[0]}: ${property[1]}`).join(delimiter);
+    }
```
OleksiyRudenko, line 25: Let's take a step back.
The following from the [earlier requirements](https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md#additional-requirements) is valid at all times.

image

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: Done. Rewrite description method and this way more universal. That's why there is  no need to override the description method in inheriting objects.

OleksiyRudenko (changes requested):
@Roophee nice try :)
Let's stick to the learnings from earlier tasks.

submissions/Roophee/a-tiny-js-world-oop/index.js
OleksiyRudenko: Universal methods are often error prone and do not add to readability. Explcitness should win above implicitness in the most of cases. Take care of peer devs that may have to extend or amend functionality of your code.

Roophee: Can I use:
 for ( let property in obj) {...} ?

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: Can I use:
for ( let property in obj) {...} ?

Roophee: Or you mean I should create athe array of properties wich is nessary to output and I can use "map" function to enumerate properties with call obj.property?

OleksiyRudenko: Yes. But rather `obj[propertyName]`
https://github.com/kottans/frontend-2021-homeworks/pull/377#discussion_r550455922

Roophee: Done.

submissions/Roophee/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,68 @@
...
+        this.tail = tail;
+        this.friends = [];
+    }
+
+    addFriends(...arrayOfFriends) {
+        arrayOfFriends.forEach(({ name }) => this.friends.push(name));
+    }
+
+    description() {
+        return ['name', 'gender', 'species', 'saying', 'legs', 'tail', 'friends'].map(property => Array.isarray(this[property]) && this[property].length === 0 ? `${property}:none` : `${property}:${this[property]}`).join(delimiter);
```
OleksiyRudenko, line 21:
```suggestion
        return ['name', 'gender', 'species', 'saying', 'legs', 'tail', 'friends'].map(property => `${property}: ` + Array.isarray(this[property]) && this[property].length === 0 ? 'none' : this[property]).join(delimiter);
```
or
```suggestion
        return ['name', 'gender', 'species', 'saying', 'legs', 'tail', 'friends'].map(property => `${property}: ${Array.isarray(this[property]) && this[property].length === 0 ? 'none' : this[property]}`).join(delimiter);
```

to avoid duplication of `` `${property}: ` ``

OleksiyRudenko (approved):
@Roophee great job!
Consider the suggestions below. Hope it is clear that either of these approches avoids code duplication (yes, template strings are also code).

submissions/Roophee/a-tiny-js-world-oop/index.js
Roophee: Thanks, it is height level art of codding).
Thanks for this experience and knowledge!!!

PR #379, opened 2020-12-29
implement post oop

akulahere:
### OOP exercise

[Demo](https://akulahere.github.io/a-tiny-JS-world/)
[Code base](https://github.com/akulahere/a-tiny-JS-world/tree/populate-world)

Please, review.

### DIFF
--- a/submissions/akulahere/task-js-post-oop/index.js
+++ b/submissions/akulahere/task-js-post-oop/index.js
@@ -16,7 +16,7 @@ class Inhabitant {
     return ['species', 'name', 'gender', 'legs', 'saying', 'friends'];
   }
 
-  makeMessage() {
+  toString() {
     return this.getProperties().map(property => `${capitalizeString(property)}: ${this[property]}`).join(' ')
   }
 }
@@ -42,7 +42,7 @@ class CatWoman extends Cat {
   }
   getProperties() {
     let properties = super.getProperties();
-    properties.splice(4, 0, 'hands');
+    properties.splice(properties.indexOf('legs'), 0, 'hands');
     return properties;
   }
 }
@@ -65,7 +65,9 @@ const woman = new Human('Lero', 'female', 'Bye!');
 const dog = new Dog('Toby', 'male');
 const cat = new Cat('Sheldon', 'male');
 const catWoman = new CatWoman('Mary');
+console.log(catWoman.getProperties())
 
+console.log(man)
 dog.makeFriend(cat);
 dog.makeFriend(man);
 man.makeFriend(cat);
@@ -73,4 +75,4 @@ const inhabitantsArray = [dog, cat, man, woman, catWoman];
 
 const capitalizeString = (string) => string.charAt(0).toUpperCase() + string.slice(1);
 
-inhabitantsArray.forEach(inhabitant => print(inhabitant.makeMessage()));
+inhabitantsArray.forEach(inhabitant => print(inhabitant));
--- a/submissions/akulahere/task-js-post-oop/index.js
+++ b/submissions/akulahere/task-js-post-oop/index.js
@@ -1,39 +1,65 @@
-const dog = {
-  species: 'dog',
-  name: 'Toby',
-  gender: 'male',
-  legs: 4,
-  hands: 0,
-  saying: 'woof-woof!'
-};
-const cat = {
-  species: 'cat',
-  name: 'Sheldon',
-  gender: 'male',
-  legs: 4,
-  hands: 0,
-  saying: 'mrrr!'
-};
-const man = {
-  species: 'human',
-  name: 'Dmitry',
-  gender: 'male',
-  legs: 2,
-  hands: 2,
-  saying: 'Hello!'
-};
-const woman = {
-  species: 'human',
-  name: 'Lero',
-  gender: 'female',
-  legs: 2,
-  hands: 2,
-  saying: 'Bye!'
-};
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
+    this.saying = saying;
+    this.friends = Array.isArray(this.friends) ? this.friends : 'No friends' ;
+  }
 
-const inhabitantsArray = [dog, cat, man, woman];
-const properties = ['species', 'name', 'gender', 'legs', 'hands', 'saying']
+  makeFriend(friend) {
+    this.friends = this.friends === 'No friends' ? [] : this.friends;
+    this.friends.push(friend.name);
+  }
+
+  makeMessage() {
+    const properties = [];
+    for (const property of Object.keys(this)) {
+      properties.push(property.toString());
+    }
+    return properties.map(property => `${capitalizeString(property)}: ${this[property]}`).join(' ')
+  }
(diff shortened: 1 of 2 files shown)


MaxKalashnyk (changes requested):
Please, create a separate repository for this task. No need to change the file from the `tiny-js-world` task.

OleksiyRudenko:
@akulahere 
image

akulahere:
@OleksiyRudenko @MaxKalashnyk fixed

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,65 @@
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
+    this.saying = saying;
+    this.friends = Array.isArray(this.friends) ? this.friends : 'No friends' ;
```
OleksiyRudenko, line 9: At object construction time `this.friends` is not defined, so `Array.isArray(this.friends)` will always be `false`.
Is there a good reason to translate list of items into string at this stage?

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,65 @@
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
```
OleksiyRudenko, line 7: This makes all inhabitants have hands, not zero but no hands at all.
Imagine, we introduce birds. This way the code will become a nightmare soon, say apart constructors that need to keep `0` constants for hands, wings, horns etc

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,65 @@
...
+  makeFriend(friend) {
+    this.friends = this.friends === 'No friends' ? [] : this.friends;
+    this.friends.push(friend.name);
+  }
+
+  makeMessage() {
+    const properties = [];
+    for (const property of Object.keys(this)) {
+      properties.push(property.toString());
+    }
```
OleksiyRudenko, line 21: [This is still valid](https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md#additional-requirements)

OleksiyRudenko (changes requested):
@akulahere well done.
There is still room for improvement.
Feel free revisiting [helpful materials from the original project](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up)

submissions/akulahere/task-js-post-oop/index.js
akulahere: fixed

submissions/akulahere/task-js-post-oop/index.js
akulahere: refactored a little to avoid this problem

submissions/akulahere/task-js-post-oop/index.js
akulahere: fixed

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,76 @@
...
+class CatWoman extends Cat {
+  constructor(name) {
+    super(name, 'female');
+    this.species = 'CatWoman';
+    this.legs = 2;
+    this.hands = 2;
+  }
+  getProperties() {
+    let properties = super.getProperties();
+    properties.splice(4, 0, 'hands');
```
OleksiyRudenko, line 45: What will happen if a peer developer adds property `eyes` in `Inhabitants` at an arbitrary position?
You may want to add `hands` somewhere next to `legs` (if my assumption is coreect). JS offers instruments to make this fragment of code more obvious. Seek among `Array` methods.

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,76 @@
...
+const catWoman = new CatWoman('Mary');
+
+dog.makeFriend(cat);
+dog.makeFriend(man);
+man.makeFriend(cat);
+const inhabitantsArray = [dog, cat, man, woman, catWoman];
+
+const capitalizeString = (string) => string.charAt(0).toUpperCase() + string.slice(1);
+
+inhabitantsArray.forEach(inhabitant => print(inhabitant.makeMessage()));
```
OleksiyRudenko, line 76: Bonus:
Check out magic method `toString` that is defined for every object in JS and is also overloadable. It is called implictly in string context. `print` ensures string context for its arguments.

OleksiyRudenko (changes requested):
@akulahere well done!
However there is an issue to solve.

submissions/akulahere/task-js-post-oop/index.js
akulahere: Replaced exact index with indexOf method

submissions/akulahere/task-js-post-oop/index.js
akulahere: Thank you! Replace makeMessage method with toString

submissions/akulahere/task-js-post-oop/index.js
akulahere: Now hands always insert before legs.

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,76 @@
...
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.saying = saying;
+    this.friends = 'No friends' ;
+  }
+
+  makeFriend(friend) {
+    this.friends = this.friends === 'No friends' ? [] : this.friends;
```
OleksiyRudenko, line 12: What if a peer developer wants to have e.g. 'none' instead of 'No friends'?
At least this should be a constant.
Not a really good approach to have interpretation of state instead of state.
Having empty array at initialization and interpretation at the point when we need this would be a more constistent approach. 
Why that? We may want to interpet the list of friends in a different way, e.g. having a count of friends.
This means that the logic inside `toString` becomes more complex (e.g. addint interim `.map()`), but it will a single point of complexity.

submissions/akulahere/task-js-post-oop/index.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor(name, gender) {
+    super('cat', name, gender, 4, 'mrrr!');
+  }
+}
+
+class CatWoman extends Cat {
+  constructor(name) {
+    super(name, 'female');
+    this.species = 'CatWoman';
+    this.legs = 2;
```
OleksiyRudenko, line 40: If we need to access inherited props directly then this is a clear sign that something is not really OK in the design.
In fact the compositional pattern is better than OOP when "hybrids" are required.
You did your best within OOP paradigm.
This was an intentional pitfall to learn that OOP has its own deficiencies.

OleksiyRudenko (approved):
@akulahere well done!
There is a couple of points for you to reflect on.

OleksiyRudenko:
@MaxKalashnyk i'm done here. Please approve and merge if you are good with what we've got in this PR.

PR #382, opened 2020-12-29
Friends app

maxim-zabolotny:
### Friends App

[Demo](https://maxim-zabolotny.github.io/friends-app.github.io/)
[Code base](https://github.com/maxim-zabolotny/friends-app.github.io)

Please, review.

### DIFF
--- a/submissions/maxim-zabolotny/friends-app/app.js
+++ b/submissions/maxim-zabolotny/friends-app/app.js
@@ -44,12 +44,10 @@ const SORT = {
     })),
 }
 
-const capitalized = ({
-    first,
-    last
-}) => {
-    let str = `${first} ${last}`
-    return str.split(' ').map(s => s.charAt(0).toUpperCase() + s.slice(1)).join(' ');
+const capitalize = str => str.charAt(0).toUpperCase() + str.slice(1);
+
+const getCapitalizedName  = ({first, last}) => {
+    return `${capitalize(first)} ${capitalize(last)}`
 }
 
 const formatDate = (date) => {
@@ -101,14 +99,14 @@ class Card {
 
 const generateMarkupForCard = (name, photo, email, date, location, cell, password) => {
     return `<div class="details">
-    <p class="user_title">${capitalized(name)}</p>
+    <p class="user_title">${getCapitalizedName(name)}</p>
     <div class="user_photo horizontal_center">
         <img src="${photo}" alt="${name}">
     </div>
     <p class="user_value">${email}</p>
 </div>
 <ul class="values_list horizontal_center">
-    <li data-title="Hi, My name is" data-value="${capitalized(name)}" data-label="name" class=""></li>
+    <li data-title="Hi, My name is" data-value="${getCapitalizedName(name)}" data-label="name" class=""></li>
     <li data-title="My email address is" data-value="${email}" data-label="email" class="active"></li>
     <li data-title="My birthday is" data-value="${formatDate(date)}" data-label="birthday" class=""></li>
     <li data-title="My address is" data-value="${location}" data-label="location" class=""></li>
--- a/submissions/maxim-zabolotny/friends-app/app.js
+++ b/submissions/maxim-zabolotny/friends-app/app.js
@@ -16,38 +16,38 @@ const DOM_NODES = {
     btnWrapper: document.querySelector('.btn-wrapper'),
     resetBtn: document.querySelector('.filter .reset_button'),
 }
-
 const state = {
-    allUsers : [],
+    allUsers: [],
     filteredUsers: [],
     sortedUsers: []
 }
-
-window.state = state
 
 const SORT = {
     all: () => renderCards(state.sortedUsers = [...state.filteredUsers]),
     female: () => renderCards(state.sortedUsers = state.filteredUsers.filter(user => user.gender.toLowerCase() == 'female')),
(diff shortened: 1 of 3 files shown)


submissions/maxim-zabolotny/friends-app/app.js
```diff
@@ -0,0 +1,215 @@
...
+        else return -1
+    })),
+    ageDesc: () => renderCards(state.sortedUsers = state.filteredUsers.sort((a, b) => {
+        if(a.age < b.age) return 1
+        else return -1
+    })),
+}
+
+const capitalized = ({first, last}) => {
+    let str = `${first} ${last}`
```
AMashoshyna, line 51: Maybe it would be better to uppercase the first name and the last name directly. I don't see a reason to combine them in one string just to split it back in the next line. Am I missing something?

submissions/maxim-zabolotny/friends-app/app.js
```diff
@@ -0,0 +1,215 @@
...
+        this.password = props.password
+        this.gender = props.gender
+        this.cell = props.cell
+    }
+
+    render() {
+        let div = document.createElement('div')
+        div.classList.add('card')
+        div.id=this.id
+        const markup = `<div class="details">
```
AMashoshyna, line 78: I suggest putting this markup outside the class. It makes class code hard to read. Maybe a helper that will receive data and return html string? You can prepare the data (uppercase etc.) in your class and then call markup function to get final html.

submissions/maxim-zabolotny/friends-app/app.js
```diff
@@ -0,0 +1,215 @@
...
+                            <li data-title="My email address is" data-value="${this.email}" data-label="email" class="active"></li>
+                            <li data-title="My birthday is" data-value="${formatDate(this.date)}" data-label="birthday" class=""></li>
+                            <li data-title="My address is" data-value="${this.location}" data-label="location" class=""></li>
+                            <li data-title="My phone number is" data-value="${this.cell}" data-label="phone" class=""></li>
+                            <li data-title="My password is" data-value="${this.password}" data-label="pass" class=""></li>
+                        </ul>`;
+        div.innerHTML = markup
+        const list = Array.from(div.getElementsByTagName("li"));
+        list.forEach((e) => {
+            e.addEventListener("mouseover", function() {
```
AMashoshyna, line 96: This looks like a good candidate for event delegation.

submissions/maxim-zabolotny/friends-app/app.js
```diff
@@ -0,0 +1,215 @@
...
+const buildCardList = () => {
+    fetch(APP_CONFIG.url)
+    .then(data => data.json())
+    .then(data => state.allUsers = data.results.map(buildCard))
+    .then(() => state.sortedUsers = [...state.allUsers])
+    .then(() => state.filteredUsers = [...state.allUsers])
+    .then(() => renderCards(state.filteredUsers))
+    .catch (e => {console.log(e)})
+
+    function buildCard (user) {
```
AMashoshyna, line 116: This function should be outside `buildCardList` imo

submissions/maxim-zabolotny/friends-app/app.js
```diff
@@ -0,0 +1,215 @@
...
+    const target = e.target.type;
+    if(!state.sortedUsers.length) state.sortedUsers = [...state.filteredUsers];
+    
+    if (target == 'radio') SORT[e.target.value]();
+};
+
+DOM_NODES.inputText.addEventListener("change", function (e) {
+    filter(e.target.value)
+});
+// DOM_NODES.inputText.addEventListener("focus", resetRadioButtons);
```
AMashoshyna, line 198: Is this still needed? It's better to remove commented out code from final version

AMashoshyna (commented):
Nice works! Please see my comments below

submissions/maxim-zabolotny/friends-app/app.js
```diff
@@ -0,0 +1,215 @@
...
+    resetBtn: document.querySelector('.filter .reset_button'),
+}
+
+const state = {
+    allUsers : [],
+    filteredUsers: [],
+    sortedUsers: []
+}
+
+window.state = state
```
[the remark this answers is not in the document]
maxim-zabolotny: This was necessary to check the state of the application.
Forgot to delete

submissions/maxim-zabolotny/friends-app/app.js
maxim-zabolotny: This function is used to process the variant in which the first and last name with a small letter comes from the server

maxim-zabolotny:
thanks for taking the time to review my code.
I hope I understood your comments correctly and corrected them correctly.

submissions/maxim-zabolotny/friends-app/app.js
AMashoshyna: BTW I would change function name `getCapitalizedName`, because  `capitalized` sounds more like a value, than a function (function usually does something, so generally it should have a verb in it's name)

What I mean is that you create a string `str` only to split it in the next line. How about something like that:
```javascript
const capitalize = str => str.charAt(0).toUpperCase() + str.slice(1);

const getCapitalizedName  = ({first, last}) => {
    return `${capitalize(first)} ${capitalize(last)}`
}
```

maxim-zabolotny: Thank you. I understand what you mean

AMashoshyna:
Well done!

PR #385, opened 2020-12-29
Frogger-game

kasionio:
### frogger-game

[Demo](https://kasionio.github.io/frontend-nanodegree-arcade-game/)
[Code base](https://github.com/kasionio/frontend-2021-homeworks/blob/frogger-game/submissions/kasionio/frogger-game/app.js)

Please, review.

### DIFF
--- a/submissions/kasionio/frogger-game/app.js
+++ b/submissions/kasionio/frogger-game/app.js
@@ -1,13 +1,37 @@
-// Enemies our player must avoid
-const ENEMY_START_POSITION_X = -100;
+const CANVAS = {
+    BLOCK_WIDTH: 101,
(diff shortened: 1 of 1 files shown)


OleksiyRudenko (changes requested):
@kasionio good job!
Yet much room for improvement.

Please, confirm all items in the list below are positive. Please, ask a question if anything requires clarification.

Minimal requirements to meet:
 * Requirements re **OOP**:
 * Most common mistakes

kasionio:
I have seen AMA and corrected my code the best I can :)

OleksiyRudenko (approved):
@kasionio great job!

PR #388, opened 2020-12-30
JS post-OOP

NJul:
### JS post-OOP

[Demo](https://njul.github.io/a-tiny-JS-world/)
[Code base](https://github.com/NJul/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/NJul/a-tiny-JS-world-oop/index.js
+++ b/submissions/NJul/a-tiny-JS-world-oop/index.js
@@ -1,0 +1,91 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/NJul/a-tiny-JS-world
+   Web app: https://njul.github.io/a-tiny-JS-world/
+  */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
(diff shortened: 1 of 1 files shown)


submissions/NJul/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,91 @@
...
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
```
OleksiyRudenko, line 17: Not all inhabitants have hands by natural design. Not just they have zero hands, but no hands at all.

submissions/NJul/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,91 @@
...
+    this.saying = saying;
+    this.friends = Array.isArray(friends) ? friends.join(', ') : 'no one';
```
OleksiyRudenko, line 19: What if want to add a friend or two later? If this prop is stored as a string then method of adding would be much different from a method of assigning initial set of friends. Not really consistent approach.

OleksiyRudenko (changes requested):
@NJul well done.
However there is some room for improvement.
Imagine we introduce birds and we want them to have wings. Following current approach we'd need to rewrite every class to have zero wings. When OOP principles are followed we'd just add class `Bird` with all required mechanics and wouldn't touch any other classes definitions.

OleksiyRudenko (approved):
Great job!

PR #389, opened 2020-12-30
Add js-post-oop task

okkkko:
### JS post OOP

[Demo](https://okkkko.github.io/js-post-oop/)
[Code base](https://github.com/okkkko/js-post-oop)

Please, review.

### DIFF
--- a/submissions/okkkko/js-post-oop/index.js
+++ b/submissions/okkkko/js-post-oop/index.js
@@ -1,13 +1,14 @@
 class Inhabitant {
-  constructor(species, name, gender, saying, friends) {
+  constructor(species, name, gender, saying, friends, legs) {
     this.species = species;
     this.name = name;
     this.gender = gender;
     this.saying = saying;
     this.friends = friends;
+    this.legs = legs;
   }
   toString() {
-    const properties = ["species", "name", "gender", "saying", "friends"].map(
+    const properties = ["species", "name", "gender", "saying", "friends", "legs"].map(
       (prop) => `${prop}: ${this[prop]}`
     );
     return properties.join("; ");
@@ -15,14 +16,13 @@ class Inhabitant {
 }
 class Human extends Inhabitant {
   constructor(name, gender, saying, friends) {
-    super("human", name, gender, saying, friends);
-    this.legs = 2;
+    super("human", name, gender, saying, friends, 2);
     this.hands = 2;
   }
   toString() {
     return [
       super.toString(),
-      ...["hands", "legs"].map((prop) => `${prop}: ${this[prop]}`),
+      ...["hands"].map((prop) => `${prop}: ${this[prop]}`),
     ].join("; ");
   }
 }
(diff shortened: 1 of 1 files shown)


submissions/okkkko/js-post-oop/index.js
```diff
@@ -0,0 +1,74 @@
...
+class Woman extends Human {
+  constructor(name, saying, friends) {
+    super(name, "female", saying, friends);
+  }
+}
+class Dog extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super("dog", name, gender, saying, friends);
+    this.legs = 4;
+    this.hands = 0;
```
OleksiyRudenko, line 40: Dogs do not have hands by natural design. Please get rid of those.

submissions/okkkko/js-post-oop/index.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super("cat", name, gender, saying, friends);
+    this.legs = 4;
+    this.hands = 0;
+  }
+  toString() {
+    return `${super.toString()}; ` + `hands: ${this.hands}; legs:${this.legs};`;
```
OleksiyRudenko, line 53: Requirements to every task earlier are valid forever.
https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md#additional-requirements

submissions/okkkko/js-post-oop/index.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super("cat", name, gender, saying, friends);
+    this.legs = 4;
```
OleksiyRudenko, line 49: All inhabitants have legs. Just like they all have names and gender etc.
Why handling legs in a different way?

OleksiyRudenko (changes requested):
@okkkko well done!
However there is room for improvement.

okkkko:
Thanks for the review!

submissions/okkkko/js-post-oop/index.js
okkkko: There are could be inhabitants without legs, and it would be more difficult to maintain code

OleksiyRudenko: This doesn't answer the question.
There could be inhabitants without gender or names. But current tech spec provides for named, gendered and legged inhabitants.

submissions/okkkko/js-post-oop/index.js
```diff
@@ -0,0 +1,76 @@
...
+class Human extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super("human", name, gender, saying, friends);
+    this.legs = 2;
+    this.hands = 2;
+  }
+  toString() {
+    return `${super.toString()}; ${["hands", "legs"]
+      .map((prop) => `${prop}: ${this[prop]}`)
+      .join("; ")};`;
```
OleksiyRudenko, line 25: Nested literals are hard to read. Formatting also can add to a better readability.

```suggestion
    return [
        super.toString(),
        ...["hands", "legs"].map(prop => `${prop}: ${this[prop]}`)
      ].join("; ");
```

OleksiyRudenko (changes requested):
@okkkko still no good reason to handle shared props in different ways. Approach must be consistent.

submissions/okkkko/js-post-oop/index.js
okkkko: Understood, thank you!

submissions/okkkko/js-post-oop/index.js
okkkko: I give up) I don`t know, I thought that it could be too specific to add legs to Inhabitant class, but maybe in this context it is appropriate

submissions/okkkko/js-post-oop/index.js
okkkko: If I add legs to the Inhabitant class, there will be only hands in the array. Should I leave it this way or get rid of the map method?

OleksiyRudenko: Both approaches are ok. I would stick to the pattern with map (due to conversion into a standard `"propName: propValue"`) as less error prone.

submissions/okkkko/js-post-oop/index.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super("cat", name, gender, saying, friends, 4);
+  }
+}
+class CatWoman extends Inhabitant {
+  constructor(name, friends) {
+    super("cat-woman", name, "female", cat.saying, friends, 2);
```
OleksiyRudenko, line 51: `cat.saying` will work only if `cat` exists and declared before any cat-woman.
You may want to use static class properties instead.

OleksiyRudenko (approved):
@okkkko well done. Approved with a minor reservation.
Check static class properties and methods. You will find those useful.

submissions/okkkko/js-post-oop/index.js
okkkko: Thank you, I`ll read about it

PR #390, opened 2020-12-30
JS Dom from f3d0t

f3d0t:
### JS Dom

[Demo](https://f3d0t.github.io/f3d0t/task_js_dom/)
[Code base](https://github.com/f3d0t/f3d0t/tree/main/task_js_dom)

Please, review.

### DIFF
--- a/submissions/f3d0t/task_js_dom/index.html
+++ b/submissions/f3d0t/task_js_dom/index.html
@@ -14,14 +14,12 @@
 			<h1 class="header__title">Five strange Ghost type pokemons</h1>
 			<span id="nav_switch" class="nav_button"></span>
 		</header>
-		<div class="main_wrapper">
-			<nav class="nav active" id="navigation">
-				<span>Choose pokemon:</span>
-			</nav>
-			<main class="main">
-				<h2 id="name" class="name"></h2>
-			</main>
-		</div>
+		<nav class="nav active" id="navigation">
+			<span>Choose pokemon:</span>
+		</nav>
+		<main class="main" id="main">
+			<h2 id="name" class="name"></h2>
+		</main>
 		<script src="script.js"></script>
 	</body>
 </html>
--- a/submissions/f3d0t/task_js_dom/index.html
+++ b/submissions/f3d0t/task_js_dom/index.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>JS - DOM</title>
+		<link rel="stylesheet" type="text/css" href="style.css" />
+		<link rel="preconnect" href="https://fonts.gstatic.com" />
+		<link href="https://fonts.googleapis.com/css2?family=Hind:wght@400;600&display=swap" rel="stylesheet" />
+	</head>
+
+	<body>
+		<header>
+			<h1 class="header__title">Five strange Ghost type pokemons</h1>
+			<span id="nav_switch" class="nav_button"></span>
+		</header>
+		<div class="main_wrapper">
+			<nav class="nav active" id="navigation">
+				<span>Choose pokemon:</span>
+			</nav>
+			<main class="main">
+				<h2 id="name" class="name"></h2>
+			</main>
+		</div>
+		<script src="script.js"></script>
+	</body>
+</html>
--- a/submissions/f3d0t/task_js_dom/script.js
+++ b/submissions/f3d0t/task_js_dom/script.js
@@ -69,7 +69,7 @@ document.addEventListener("DOMContentLoaded", () => {
 		if (target.nodeName === "LI" && oldId !== currentId) {
 			nav.querySelector(".active")?.classList.toggle("active");
 			target.classList.toggle("active");
-			let newPokemon = pokemons.find((item) => item.id == currentId);
+			const newPokemon = pokemons.find((item) => item.id == currentId);
 			if (newPokemon) {
 				main.innerHTML = "";
 				main.appendChild(createPokemonContent(newPokemon));
--- a/submissions/f3d0t/task_js_dom/script.js
+++ b/submissions/f3d0t/task_js_dom/script.js
@@ -42,23 +42,20 @@ const createMenuList = (arrayOfPokemons, container) => {
 	container.appendChild(wrapper);
 };
 
-const getPokemonContent = (id) => {
+const createPokemonContent = (pokemon) => {
 	const wrapper = document.createElement("div");
-	const currentPokemon = pokemons.find((item) => item.id == id);
-	if (!currentPokemon) {
-		const h2 = document.createElement("h2");
-		h2.classList.add("main__name");
-		h2.innerText = currentPokemon.name;
-		const img = document.createElement("img");
-		img.classList.add("main__img");
-		img.src = currentPokemon.imagePath;
-		img.alt = currentPokemon.name;
-		const p = document.createElement("p");
-		p.classList.add("main__text");
-		p.innerText = currentPokemon.description;
-		wrapper.append(h2, img, p);
-		return wrapper;
-	} else return false;
+	const h2 = document.createElement("h2");
+	h2.classList.add("main__name");
+	h2.innerText = pokemon.name;
+	const img = document.createElement("img");
+	img.classList.add("main__img");
+	img.src = pokemon.imagePath;
+	img.alt = pokemon.name;
+	const p = document.createElement("p");
+	p.classList.add("main__text");
+	p.innerText = pokemon.description;
+	wrapper.append(h2, img, p);
+	return wrapper;
 };
 
 document.addEventListener("DOMContentLoaded", () => {
@@ -72,10 +69,10 @@ document.addEventListener("DOMContentLoaded", () => {
 		if (target.nodeName === "LI" && oldId !== currentId) {
 			nav.querySelector(".active")?.classList.toggle("active");
 			target.classList.toggle("active");
-			let newPokemon = getPokemonContent(currentId);
+			let newPokemon = pokemons.find((item) => item.id == currentId);
 			if (newPokemon) {
 				main.innerHTML = "";
-				main.appendChild(newPokemon);
+				main.appendChild(createPokemonContent(newPokemon));
 			}
 		}
 	});
--- a/submissions/f3d0t/task_js_dom/script.js
+++ b/submissions/f3d0t/task_js_dom/script.js
@@ -45,7 +45,7 @@ const createMenuList = (arrayOfPokemons, container) => {
 const getPokemonContent = (id) => {
 	const wrapper = document.createElement("div");
 	const currentPokemon = pokemons.find((item) => item.id == id);
-	if (currentPokemon !== undefined) {
+	if (!currentPokemon) {
 		const h2 = document.createElement("h2");
 		h2.classList.add("main__name");
 		h2.innerText = currentPokemon.name;
--- a/submissions/f3d0t/task_js_dom/script.js
+++ b/submissions/f3d0t/task_js_dom/script.js
@@ -31,9 +31,9 @@ const pokemons = [
 	},
 ];
 
-const createMenuLinks = (array, container) => {
+const createMenuList = (arrayOfPokemons, container) => {
 	const wrapper = document.createElement("ul");
-	array.forEach((element) => {
+	arrayOfPokemons.forEach((element) => {
 		const li = document.createElement("li");
 		li.textContent = element.name;
 		li.dataset.id = element.id;
@@ -44,25 +44,27 @@ const createMenuLinks = (array, container) => {
 
 const getPokemonContent = (id) => {
 	const wrapper = document.createElement("div");
-	const item = pokemons.find((item) => item.id == id);
-	const h2 = document.createElement("h2");
-	h2.classList.add("main__name");
-	h2.innerText = item.name;
-	const img = document.createElement("img");
-	img.classList.add("main__img");
-	img.src = item.imagePath;
-	img.alt = item.name;
-	const p = document.createElement("p");
-	p.classList.add("main__text");
-	p.innerText = item.description;
-	wrapper.append(h2, img, p);
-	return wrapper;
+	const currentPokemon = pokemons.find((item) => item.id == id);
+	if (currentPokemon !== undefined) {
+		const h2 = document.createElement("h2");
+		h2.classList.add("main__name");
+		h2.innerText = currentPokemon.name;
+		const img = document.createElement("img");
+		img.classList.add("main__img");
+		img.src = currentPokemon.imagePath;
+		img.alt = currentPokemon.name;
+		const p = document.createElement("p");
+		p.classList.add("main__text");
+		p.innerText = currentPokemon.description;
+		wrapper.append(h2, img, p);
+		return wrapper;
+	} else return false;
 };
 
 document.addEventListener("DOMContentLoaded", () => {
 	const nav = document.getElementById("navigation");
 	const main = document.getElementById("main");
-	createMenuLinks(pokemons, nav);
+	createMenuList(pokemons, nav);
 
 	nav.addEventListener("click", function ({ target }) {
 		const currentId = target.dataset.id;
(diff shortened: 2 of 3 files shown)


submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+document.addEventListener("DOMContentLoaded", () => {
+	const nav = document.getElementById("navigation");
+	const nameHeader = document.getElementById("name");
+	createMenuLinks(pokemons, nav);
+
+	nav.addEventListener("click", function (evt) {
+		let currentId = evt.target.dataset.id;
+		if (currentId != undefined) {
```
babayK0, line 73: Why not check like this ?
```suggestion
		if (currentId) {
```
Also this condition will return true if id will be `null`. try to use [Strict Equality Comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality#Description)

And add check for active menu item, because now if i click multiple times in same nav item - content will repaint

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+			nav.querySelector(".active") != null ? nav.querySelector(".active").classList.toggle("active") : null;
```
babayK0, line 74: [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) will do the same, i guess

```suggestion
			nav.querySelector(".active")?.classList.toggle("active");
```

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+			evt.target.classList.toggle("active");
+			let main = document.querySelector(".main");
+			main.parentElement.replaceChild(paintContent(currentId, pokemons), main);
```
babayK0, line 77: Don't rely on HTML structure.

try to implement with [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) or with something else.

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+	const nav = document.getElementById("navigation");
+	const nameHeader = document.getElementById("name");
+	createMenuLinks(pokemons, nav);
+
+	nav.addEventListener("click", function (evt) {
+		let currentId = evt.target.dataset.id;
+		if (currentId != undefined) {
+			nav.querySelector(".active") != null ? nav.querySelector(".active").classList.toggle("active") : null;
+			evt.target.classList.toggle("active");
+			let main = document.querySelector(".main");
```
babayK0, line 76: It is not blocker. But you create this variable on every click.

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+	},
+	{
+		id: 790,
+		name: "Cosmoem",
+		description: "The king who ruled Alola in times of antiquity called it the “cocoon of the stars” and built an altar to worship it.",
+		imagePath: "./img/790.png",
+	},
+];
+
+const createMenuLinks = (array, container) => {
```
babayK0, line 34: array is to generic name for variable

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+	let wrapper = document.createElement("ul");
+	array.forEach((element) => {
+		let li = document.createElement("li");
+		li.textContent = element.name;
+		li.dataset.id = element.id;
+		li.className = "";
```
babayK0, line 40: Do you need this ?

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+		wrapper.appendChild(li);
+	});
+	container.appendChild(wrapper);
+};
+
+const paintContent = (id, arr) => {
```
babayK0, line 46: which 'id' and 'arr' ?)
to general names for variables

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+	let wrapper = document.createElement("main");
```
babayK0, line 47: Why do you use `let` here (and in some other places) ? use `let` if variable value will be changed

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+	wrapper.className = "main";
+	let item = arr.find((item) => item.id == id);
```
babayK0, line 49: what you will do if find will return `undefined` ?

My suggestion: pass only item to this function. (If there is not such item don't call this function)

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+		li.dataset.id = element.id;
+		li.className = "";
+		wrapper.appendChild(li);
+	});
+	container.appendChild(wrapper);
+};
+
+const paintContent = (id, arr) => {
+	let wrapper = document.createElement("main");
+	wrapper.className = "main";
```
babayK0, line 48: Use `classList` please, it is more predictable than `className`;

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,84 @@
...
+	let img = document.createElement("img");
+	img.className = "main__img";
+	img.src = item.imagePath;
+	img.alt = item.name;
+	let p = document.createElement("p");
+	p.className = "main__text";
+	p.innerText = item.description;
+	wrapper.appendChild(h2);
+	wrapper.appendChild(img);
+	wrapper.appendChild(p);
```
babayK0, line 62: with [append](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append) you can do things like this 

```
el.append(el1, el2, ...)
```

submissions/f3d0t/task_js_dom/script.js
babayK0, line 46: Function have name `paintContent` but it just return element

getContentHtmlElement (not use this name it is just for example :rofl: )

submissions/f3d0t/task_js_dom/style.css
```diff
@@ -0,0 +1,125 @@
...
+.nav_button {
+	display: block;
+	position: absolute;
+	margin: 10px 20px;
+	width: 50px;
+	height: 50px;
+	top: 0;
+	left: 0;
+	border-radius: 50%;
+	background-image: linear-gradient(to bottom, rgb(237, 83, 77) 48%, rgb(51, 51, 51) 48%, rgb(51, 51, 51) 52%, rgb(255, 255, 255) 52%, rgb(255, 255, 255) 100%);
```
babayK0, line 20: Try to format your this code to make it more ease to read)

submissions/f3d0t/task_js_dom/index.html
```diff
@@ -0,0 +1,27 @@
...
+		<link rel="preconnect" href="https://fonts.gstatic.com" />
+		<link href="https://fonts.googleapis.com/css2?family=Hind:wght@400;600&display=swap" rel="stylesheet" />
+	</head>
+
+	<body>
+		<header>
+			<h1 class="header__title">Five strange Ghost type pokemons</h1>
+			<span id="nav_switch" class="nav_button"></span>
+		</header>
+		<div class="main_wrapper">
```
babayK0, line 17: Do you need this wrapper ?

babayK0 (changes requested):
Good job!

Watch my comments below please)

submissions/f3d0t/task_js_dom/script.js
f3d0t: I don't understand this. Why i can't rely on HTML structure?

submissions/f3d0t/task_js_dom/script.js
f3d0t: Thanks! =)

submissions/f3d0t/task_js_dom/script.js
babayK0: Because if someone will change your `html` your `js` will be broken)

Don't use constructions like children[0], firstElementChild, parentNode, nextSibling, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use querySelector or closest, if in event, instead.

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -37,44 +37,41 @@ const createMenuLinks = (array, container) => {
...
 
-const paintContent = (id, arr) => {
-	let wrapper = document.createElement("main");
-	wrapper.className = "main";
+const getMainContent = (id, arr) => {
+	let wrapper = document.createElement("div");
 	let item = arr.find((item) => item.id == id);
 	let h2 = document.createElement("h2");
-	h2.className = "main__name";
+	h2.classList = "main__name";
```
babayK0, line 50: Read how to use [classList](https://developer.mozilla.org/en/docs/Web/API/Element/classList)

submissions/f3d0t/task_js_dom/script.js
babayK0: @f3d0t not fixed

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -37,44 +37,41 @@ const createMenuLinks = (array, container) => {
 		let li = document.createElement("li");
 		li.textContent = element.name;
 		li.dataset.id = element.id;
-		li.className = "";
+		li.classList = "";
```
babayK0, line 40: ?

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -37,44 +37,41 @@ const createMenuLinks = (array, container) => {
...
 		wrapper.appendChild(li);
 	});
 	container.appendChild(wrapper);
 };
 
-const paintContent = (id, arr) => {
-	let wrapper = document.createElement("main");
-	wrapper.className = "main";
+const getMainContent = (id, arr) => {
```
babayK0, line 46: naming not fixed

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -37,44 +37,41 @@ const createMenuLinks = (array, container) => {
...
-	nav.addEventListener("click", function (evt) {
-		let currentId = evt.target.dataset.id;
-		if (currentId != undefined) {
-			nav.querySelector(".active") != null ? nav.querySelector(".active").classList.toggle("active") : null;
-			evt.target.classList.toggle("active");
-			let main = document.querySelector(".main");
-			main.parentElement.replaceChild(paintContent(currentId, pokemons), main);
+	nav.addEventListener("click", function ({ target }) {
+		let currentId = target.dataset.id;
+		let oldId = nav.querySelector(".active")?.dataset.id;
```
babayK0, line 70: Why do you use `let` ? Fix all `let`'s - declare with `const` instead of `let` and only if variable will change - use `let`

babayK0 (changes requested):
Not all was fixed

submissions/f3d0t/task_js_dom/script.js
babayK0: @f3d0t :arrow_up:

submissions/f3d0t/task_js_dom/script.js
f3d0t: I was added empty class because className.toggle() throw an error when there is no class attribute. For now it's deprecated.

f3d0t:
Fixed. Thank you, @babayK0

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -0,0 +1,81 @@
...
+		li.textContent = element.name;
+		li.dataset.id = element.id;
+		wrapper.appendChild(li);
+	});
+	container.appendChild(wrapper);
+};
+
+const getPokemonContent = (id) => {
+	const wrapper = document.createElement("div");
+	const item = pokemons.find((item) => item.id == id);
```
babayK0, line 47: @f3d0t  What you will do if find will return `undefined`? I wrote about that but seems you ignore that message

submissions/f3d0t/task_js_dom/script.js
babayK0: @f3d0t not fixed

babayK0 (changes requested):

submissions/f3d0t/task_js_dom/script.js
f3d0t: I will now add a check if it is not undefined

f3d0t: now getPokemonContent() returns false and there is checkout to not change content if it returns false

f3d0t:
okay in this time I think all would be fine =)

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -44,34 +44,39 @@ const createMenuLinks = (array, container) => {
...
-	img.classList.add("main__img");
-	img.src = item.imagePath;
-	img.alt = item.name;
-	const p = document.createElement("p");
-	p.classList.add("main__text");
-	p.innerText = item.description;
-	wrapper.append(h2, img, p);
-	return wrapper;
+	const currentPokemon = pokemons.find((item) => item.id == id);
+	if (currentPokemon !== undefined) {
```
babayK0, line 48:
```suggestion
	if (!currentPokemon) {
```
Also:
> My suggestion: pass only item to this function. (If there is not such item don't call this function)

[Old comment](https://github.com/kottans/frontend-2021-homeworks/pull/390#discussion_r550232175)

Why this is good: you prevent redundant function calls.

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -44,34 +44,39 @@ const createMenuLinks = (array, container) => {
...
 
 	nav.addEventListener("click", function ({ target }) {
 		const currentId = target.dataset.id;
 		const oldId = nav.querySelector(".active")?.dataset.id;
 		if (target.nodeName === "LI" && oldId !== currentId) {
 			nav.querySelector(".active")?.classList.toggle("active");
 			target.classList.toggle("active");
-			main.innerHTML = "";
-			main.appendChild(getPokemonContent(currentId));
+			let newPokemon = getPokemonContent(currentId);
```
babayK0, line 75: Explain why you using `let` here ?

babayK0 (changes requested):
Please, don't hurry up !
You are repeating old mistakes.

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -42,23 +42,20 @@ const createMenuList = (arrayOfPokemons, container) => {
 	container.appendChild(wrapper);
 };
 
-const getPokemonContent = (id) => {
+const createPokemonContent = (pokemon) => {
```
babayK0, line 45: Why you changed naming for function ?

submissions/f3d0t/task_js_dom/script.js
```diff
@@ -72,10 +69,10 @@ document.addEventListener("DOMContentLoaded", () => {
 		if (target.nodeName === "LI" && oldId !== currentId) {
 			nav.querySelector(".active")?.classList.toggle("active");
 			target.classList.toggle("active");
-			let newPokemon = getPokemonContent(currentId);
+			const newPokemon = pokemons.find((item) => item.id == currentId);
```
babayK0, line 72: why it calls `newPokemon` ? 
`pokemon` or `currentPokemon` is more meaningful

babayK0 (changes requested):
I hope these are the last minor changes that are expected to be made.

babayK0:
@ f3d0t, please answer the questions I asked in recent comments.

submissions/f3d0t/task_js_dom/script.js
f3d0t: I was thinking about that this function actually creates new element that then appends to the container. But now i see that previous naming was more suitable. Should I change it back?

submissions/f3d0t/task_js_dom/script.js
f3d0t: Because , I think that currentPocemon - it's a pokemon that is currently on the screen. This variable holds whole new pokemon from list, and that's why i called it so. Just pokemon is also good I think

f3d0t:
@babayK0 
Answered. Thank you for your great patience =)

submissions/f3d0t/task_js_dom/script.js
babayK0: there is no need for it now

babayK0 (approved):
LGTM!

PR #394, opened 2020-12-30
Tiny JS world (post-OOP)

AsyaYeromina:
### Tiny JS world (post-OOP)

[Demo](https://asyayeromina.github.io/a-tiny-JS-world/)
[Code base](https://github.com/AsyaYeromina/a-tiny-JS-world)

Please, review.


submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
```diff
@@ -0,0 +1,146 @@
...
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitant {
+  constructor(inhabitantProperties) {
+    this.species = inhabitantProperties.species;
+    this.name = inhabitantProperties.name;
+    this.gender = inhabitantProperties.gender;
+    this.legs = inhabitantProperties.legs;
+    this.hands = inhabitantProperties.hands;
```
OleksiyRudenko, line 17: Not all inhabitants have hands. Imagine how owehwhelmed becomes this class when we add birds, rhinos, fish etc
The beauty of OOP is that every class inherits shared props and adds what is unique to it and its children.
Revisit [the materials from the original project](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up).

Let also every class handle its own properties. Note that base class shouldn't know and handle props introduced by any descendants.

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
```diff
@@ -0,0 +1,146 @@
...
+const man = new Man({
+  name: "Stepan",
+  speech: "Hello!",
+  friends: ["Ivanka", "Python", "Jessie"],
+});
+
+const catWoman = new FelisSapiens({
+  name: "Selina",
+  gender: "female",
+  speech: new FelisCatus({}).catSpeech(),
```
OleksiyRudenko, line 119: Looks like we create a cat to borrow its speech and then... get rid of the cat. Sad.
Consider static properties for the Cat class.

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
```diff
@@ -0,0 +1,146 @@
...
+    return [
+      this.species,
+      this.name,
+      this.gender,
+      this.legs,
+      this.hands,
+      this.speech,
+      this.friends,
+    ]
+      .filter((propertyValue) => propertyValue !== null)
```
OleksiyRudenko, line 32: You won't need `filter` once this class gets rid of props that aren't inherited by all descendants.

OleksiyRudenko (changes requested):
@AsyaYeromina pretty well done.
However we can benefit from following OOP principles.
Imagine you need to add dozens of species, some of them with unique props.

OleksiyRudenko:
@AsyaYeromina the following may pose some problems when merging this PR
image

I'd recommend getting rid of irrelevant commits to avoid possible conflicts and unwanted changes in the code from past merged submissions.

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
AsyaYeromina: Ok, I've understood. I've added hands only for Sapiens class. 
I've also added tail only for Animals as an example of sharing props with all animals.

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
AsyaYeromina: Fixed

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
AsyaYeromina: Yes, thanks
Removed filter

AsyaYeromina:
@OleksiyRudenko 
Is it ok now that toString method is rewritten for Sapiens and Animals? 
I have no idea how to refactor this. Have you any prompt?

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
```diff
@@ -0,0 +1,167 @@
...
+      this.tail,
+      this.speech,
+      this.friends,
+    ]
+    .map((propertyValue) =>
+    propertyValue == this.tail && propertyValue === true
+      ? (propertyValue = "has tail")
+      : propertyValue == this.tail && propertyValue === false
+      ? (propertyValue = "no tail")
+      : propertyValue
```
OleksiyRudenko, line 43: What happens here:

Issue 1.
`propertyValue == this.tail && propertyValue === true`  -- if either of properties (`this.species`, `this.name` etc including `this.tail`) has the same value as `this.tail` and the value is `true`, then "has tail" is a result from evaluation (`propertyValue` gets assigned but it is never stored in fact, see below for details).
So if it happens we've got a property e.g. `this.hasHorn` and it is `true` and we also have `this.tail`, which is also `true` then we will receive "has tail" twice.

What you most likely meant here is that if property **name** is `tail`, then... etc.

To make the latter logic to work you'd rather need property **names** as strings in the initial array. Then you'd be able to compare each value in the array against `"tail"` and build relevant logic. And whatever property name is not `"tail"` you then extract its value by means of `this[propName]` (i.e. `this["name"]` is equivalent of `this.name`)

Issue 2. Ternary
Whatever `propertyValue` is assigned within this ternary is never used (this variable, or rather callback argument, is scoped to this callback only). `propertyValue` in the subsequent `map` is a completely different variable, just with a similar name.

So, why ternary still returns "has tail" and "no tail"? Because expressions in JS **always** return something. So statement `propertyValue = "has tail"` returns "has tail". Try removing `propertyValue =` after both of `?` and note that the outcome doesn't change (parentheses also become obsolete while do not block any expession within).

Basically, this ternary here returns either of "has tail", "no tail" or a value received from the initial array.

Just of curiosity try executing each statements below, one by one in browser console and spot the output from each  of the statements.
```
let a = "a", b = "b";

a = "a", b = "b" // yes, it is comma and it is generally valid

(propertyValue = "has tail")

[1,2,3].forEach(n => n + 1)

[1,2,3].map(n => n + 1)
```

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
```diff
@@ -0,0 +1,167 @@
...
+    )
+    .map((propertyValue) =>
+      Array.isArray(propertyValue) ? propertyValue.join(", ") : propertyValue
```
OleksiyRudenko, line 46: Note, how ternary works here. This is a good pattern of ternary use.

submissions/AsyaYeromina/Tiny-JS-world-OOP/index.js
```diff
@@ -0,0 +1,167 @@
...
+      this.gender,
+      this.legs,
+      this.hands,
+      this.speech,
+      this.friends,
+    ]
+      .map((propertyValue) =>
+        Array.isArray(propertyValue) ? propertyValue.join(", ") : propertyValue
+      )
+      .join("; ");
```
OleksiyRudenko, line 85: The previous similar example sourced too much information. So I will use this one to emphasize yet another issue.

Imagine, we add winged birds, horned cattle, finned fish, nautic mammals etc, all introducing their unique features.

Current approach will appear suboptimal pretty soon. Encapsulation and methods overloading (with calls to parents' methods) are to the rescue.

Encapsulation: let each class handle props it **owns** directly.
This means that `Inhabitant` class should implement its own version of `toString` to build representation from the props it owns directly.
Same is applicable to every child class that introduces its own unique property. How a child class can benefit from parent's `toString` implementation? `super.methodName()` calls an inherited version of a method.

E.g.
```
toString() {
  return
    [ super.toString(),
      this.uniqueProperty,
    ].join(delimiterConst);
}
```

Perhaps, in `Animal` class `this.uniqueProperty` will work even better as `this.uniqueProperty ? "has tail" : "no tail"`

Well, yes. Tails and hands will come after everything else. A trade-off. Or consider reordering properties in the list so they combined still create a sense of sanity.

Explicit labels in presentation string might also work well.

Consider the example below:
```
["propA", "propB"].map(propName => `${propName}: ${this[propName]}`).join(delimiter);
```

OleksiyRudenko (changes requested):
@AsyaYeromina pretty well done.
Let's make code even more scalable and maintainable.

AsyaYeromina:
@OleksiyRudenko Thank you for such a detailed explanation! It's incredibly cool that my code was so carefully and patiently reviewed. 
Does it seem ok now?

OleksiyRudenko (approved):
Great job!

PR #396, opened 2020-12-30
Game-frogger

ik-web:
### Object-Oriented JavaScript - practice

[Demo](https://ik-web.github.io/game-frogger/)
[Code base](https://github.com/ik-web/game-frogger)

Please, review.


OleksiyRudenko:
After a couple of wins level increments. Then at failure (collision with a bug) the alert reports same level (say, 4) but it actually becomes 1. Isn't it a bug?

submissions/ik-web/frogger/app.js
```diff
@@ -0,0 +1,198 @@
...
+Character.prototype.update = function() {};
+
+// Draw the character on the screen, required method for game
+Character.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+//----------------------------------------------
+// Constructor for create enemy
+const Enemy = function(x, y, sprite, speed, player) {
```
OleksiyRudenko, line 82: Enemy constructor receives the `player` argument but it never used. Other class' methods still refer to the global variable `player`

submissions/ik-web/frogger/app.js
```diff
@@ -0,0 +1,198 @@
...
+
+//Method for checking game progress
+Player.prototype.checkProgress = function() {
+    if (this.y < 0) {
+
+        setTimeout(() => {
+            this.x = playerStartPosX;
+            this.y = playerStartPosY;
+            this.lvl++;
+            levelCounter.innerHTML = `Level: ${this.lvl}`;
```
OleksiyRudenko, line 136: `levelCounter` innerHTML is assigned in multiple places and with different approaches. Introduce and use a function for this repetetive operation. This will reduce code repetition and will make code more maintainable.

OleksiyRudenko (changes requested):
@ik-web nice job!
However ther is some room for improvement.

submissions/ik-web/frogger/app.js
ik-web: I fixed this thing

submissions/ik-web/frogger/app.js
ik-web: I fixed it

ik-web:
> After a couple of wins level increments. Then at failure (collision with a bug) the alert reports same level (say, 4) but it actually becomes 1. Isn't it a bug?

I do not know English very well yet, perhaps I have incorrectly formulated the idea in the message displayed in the alert. The idea was to show in case of defeat what level was reached, but to reset the game itself to the beginning.

submissions/ik-web/frogger/app.js
OleksiyRudenko: Does the function adds score or sets/assigns score? Is it score or level? Some of the parameters att call times say `lvl` (i assume this abbreviation means 'level')

ik-web: I fixed it. Thanks for this useful information

submissions/ik-web/frogger/app.js
```diff
@@ -0,0 +1,202 @@
...
+});
+// ---------------------------------------------
+// ---------------------------------------------
+// ---------------------------------------------
+// Create player character
+const player = new Player(playerStartPosX, playerStartPosY, playerSprite);
+
+// Create first enemy
+let enemy = new Enemy(enemyPosX, enemyPosY(), enemySprite, enemySpeed(), player);
+let allEnemies = [enemy];
```
OleksiyRudenko, line 199:
```suggestion
let allEnemies = [];
addEnemy();
```
This way the approach to create enemies is consistent with the one you already use and also makes code shorter.

OleksiyRudenko (approved):
@ik-web great job!

PR #398, opened 2020-12-31
dom-js-task

SamVal007:
### JS-DOM

[demo](https://samval007.github.io/kottans-frontend/Task_DOM_SnookerPlayers/)
[codebase](https://github.com/SamVal007/kottans-frontend/tree/main/Task_DOM_SnookerPlayers)

Sorry, I understand my mistakes, i hope:)
Please, review.

### DIFF
--- a/submissions/SamVal007/task-JS-dom/dom.js
+++ b/submissions/SamVal007/task-JS-dom/dom.js
@@ -1,10 +1,12 @@
+const nav = document.querySelector('nav');
+
 const PLAYERS = [{
     name: "Ronnie O'Sullivan",
     born: "5 December 1975 (age 45) Wordsley, West Midlands, England",
     currentRanking: "3 (as of 21 December 2020)",
     careerWinnings: "£11,776,155",
     picture: "./img/Ronnie O'Sullivan.jpg",
-    descr: [
+    description: [
       "Ronald Antonio O'Sullivan OBE (born 5 December 1975) is an English professional snooker player from Essex. As a six-time (and reigning) world champion, a record seven-time Masters champion, and a record seven-time UK champion, he is the most successful player in the history of snooker's Triple Crown Series, with a record 20 titles. He also holds the record for the most ranking titles in professional snooker, with 37. His career prize money of over £11.7 million is the most by any player in snooker history.",
       "A noted snooker prodigy from an early age, O'Sullivan made his first competitive century break at age 10, won the British Under-16 Championship at age 13, achieved his first competitive maximum break at age 15, and won the IBSF World Under-21 Snooker Championship before turning professional in 1992, aged 16. He won his first ranking title at the 1993 UK Championship aged 17 born and 358 days, making him the youngest player ever to win a professional ranking event, a record he still holds. He is also the youngest player ever to win the Masters, which he first achieved in 1995, aged 19 born and 69 days. O'Sullivan is now also noted for his longevity in the sport, having competed in a record 28 consecutive World Championships at the Crucible between 1993 and 2020. Winning the 2020 World Championship aged 44 born and 254 days made him the second-oldest player (after Ray Reardon) to win a world title in snooker's modern era.",
       "A prolific break-builder, O'Sullivan is the only player to have achieved 1,000 career century breaks, a milestone he reached in the 2019 Players Championship final. He has also achieved the highest number of officially recognized maximum breaks in professional competition, with 15, and the fastest competitive maximum break, compiled in a time of 5 minutes and 8 seconds at the 1997 World Championship.",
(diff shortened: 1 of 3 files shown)


submissions/SamVal007/task-JS-dom/dom.js
```diff
@@ -0,0 +1,148 @@
...
+  },
+]
+
+const playerPhoto = document.querySelector('.photo');
+const playerName = document.querySelector('.player-name');
+const playerBorn = document.querySelector('.born');
+const playerCurrentRanking = document.querySelector('.current-ranking');
+const playerCareerWinnings = document.querySelector('.career-winnings');
+const playerDescription = document.querySelector('.descr');
+const nav = document.querySelector('nav');
```
AMashoshyna, line 100: Does the nav have a certain role in the application? I would try to find a better explanation for why it is here

submissions/SamVal007/task-JS-dom/dom.js
```diff
@@ -0,0 +1,148 @@
...
+
+let currentPlayerIndex = "0";
```
AMashoshyna, line 102: Why a string?

submissions/SamVal007/task-JS-dom/dom.js
```diff
@@ -0,0 +1,148 @@
...
+
+const generatePlayerDescription = (id) => {
+  const fragment = document.createDocumentFragment();
+  const div = document.createElement('div');
+
+  PLAYERS[id].descr.forEach(item => {
+    const p = document.createElement('p');
+
+    p.textContent = item;
+    div.appendChild(p);
```
AMashoshyna, line 117: DOM access is a rather expensive operation. Let's try to make it in one go. Suggestion: take a look at `document.fragment` API.

submissions/SamVal007/task-JS-dom/dom.js
```diff
@@ -0,0 +1,148 @@
...
+  const data = PLAYERS[activeElem.id];
+  playerPhoto.setAttribute('src', data.picture);
+  playerName.textContent = data.name;
+  playerBorn.textContent = data.born;
+  playerCurrentRanking.textContent = data.currentRanking;
+  playerCareerWinnings.textContent = data.careerWinnings;
+}
+
+
+const handler = ({target}) => {
```
AMashoshyna, line 135: `handler` sound too generic.  Let's find a better name

submissions/SamVal007/task-JS-dom/style.css
```diff
@@ -0,0 +1,236 @@
...
+  clip: rect(1px 1px 1px 1px);
+  clip: rect(1px, 1px, 1px, 1px);
+  padding: 0;
+  border: 0;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+}
+
+/* #menuGamburger {
```
AMashoshyna, line 117: ??

submissions/SamVal007/task-JS-dom/dom.js
```diff
@@ -0,0 +1,148 @@
...
+    ]
+  },
+]
+
+const playerPhoto = document.querySelector('.photo');
+const playerName = document.querySelector('.player-name');
+const playerBorn = document.querySelector('.born');
+const playerCurrentRanking = document.querySelector('.current-ranking');
+const playerCareerWinnings = document.querySelector('.career-winnings');
+const playerDescription = document.querySelector('.descr');
```
AMashoshyna, line 99: I really don't see a reason to shorten "description" to "desrc" here) but you can leave it if you like.

submissions/SamVal007/task-JS-dom/style.css
```diff
@@ -0,0 +1,236 @@
...
+
+body {
+  display: flex;
+  margin: 0;
+  height: 100vh;
+  font-family: Snell Roundhand, cursive;
+  background-color: #f5f5f5;
+}
+
+ul {
```
AMashoshyna, line 13: Styling by tag is risky: this will style ALL ul-s in a page. Not quite applicable for this small task, but better to keep in mind.

AMashoshyna (commented):
Good job! Please take a look at the comments

submissions/SamVal007/task-JS-dom/dom.js
SamVal007: Basically, dont know where to put it better:)
LEt's put it in the top? It's main element i would say.

submissions/SamVal007/task-JS-dom/dom.js
SamVal007: Yeah, really, good point! Thank you! Dont remeber why I put it as string fro the beginning!

submissions/SamVal007/task-JS-dom/dom.js
SamVal007: Ok, I will learn how to use it and would change the code. Thank You!

submissions/SamVal007/task-JS-dom/dom.js
SamVal007: May be let's try displayMain?

submissions/SamVal007/task-JS-dom/style.css
SamVal007: Yep, my bad. Deleted this

submissions/SamVal007/task-JS-dom/dom.js
SamVal007: Yes, I agree! Changed!

submissions/SamVal007/task-JS-dom/style.css
SamVal007: Also agree, usual dont do it like this, but there is only one UL, so that's why.
But ok, let's do it better!
Very Big Thanks for helping!!!
And Happy New Year!:)

submissions/SamVal007/task-JS-dom/dom.js
```diff
@@ -0,0 +1,151 @@
...
+let currentPlayerIndex = 0;
+
+const toggleActiveLink = (newActiveEl) => {
+  document.querySelector('.active').classList.toggle('active');
+  newActiveEl.classList.toggle('active');
+}
+
+const generatePlayerDescription = (id) => {
+  const fragment = document.createDocumentFragment();
+  const div = document.createElement('div');
```
AMashoshyna, line 113: You don't need this div =) You can return you document fragment directly, it can be used as innerHtml later

submissions/SamVal007/task-JS-dom/dom.js
AMashoshyna: I think `menuContainer`, `navigationNode` or something like this may be used, so that variable name explains why we need it. In terms of your application logic it does not matter if the DOM element is a nav or a div, that's why I was confused

submissions/SamVal007/task-JS-dom/dom.js
AMashoshyna: `displayContent` maybe? Because in this case (similar to nav earlier) it does not matter what HTML element is used (one day you can decide to change DOM structure and put it into a div or a section later), but it matters that this function changes displayed content.

submissions/SamVal007/task-JS-dom/dom.js
SamVal007: Ok, that's the best what I  found how to do it without this additional Div:)

AMashoshyna:
Good job!

PR #399, opened 2020-12-31
Js-OOP frogger by F3d0t

f3d0t:
### Object Oriented JS

[Demo](https://f3d0t.github.io/f3d0t/js_oop/)
[Code base](https://github.com/f3d0t/f3d0t/tree/main/js_oop/js)

Please, review.

### DIFF
--- a/submissions/f3d0t/js-oop/app.js
+++ b/submissions/f3d0t/js-oop/app.js
@@ -1,0 +1,139 @@
+const enemyInfo = {
+	minSpeed: 100,
+	maxSpeed: 300,
+	sprite: "images/enemy-bug.png",
+	width: 75,
+};
+const cellInfo = {
+	width: 101,
+	height: 83,
+	beingPositionCorrection: 28,
+};
+const playerInfo = {
+	sprite: "images/char-boy.png",
+	startPositionX: cellInfo.width * 2,
+	startPositionY: cellInfo.height * 5 - cellInfo.beingPositionCorrection,
+};
+const playingAreaInfo = {
+	width: cellInfo.width * 4,
+	height: cellInfo.height * 5,
+};
+const allEnemies = [];
+
+let generateEnemySpeed = function () {
+	return Math.floor(Math.random() * (enemyInfo.maxSpeed - enemyInfo.minSpeed)) + enemyInfo.minSpeed;
+};
+
+const Being = function (x, y, sprite) {
+	this.x = x;
+	this.y = y;
+	this.sprite = sprite;
+};
+
+Being.prototype.render = function () {
+	ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const Enemy = function (x, y) {
+	Being.call(this, x, y, enemyInfo.sprite);
+	this.speed = generateEnemySpeed();
+};
+
+Enemy.prototype = Object.create(Being.prototype);
+
+Enemy.prototype.update = function (dt) {
+	this.x += this.speed * dt;
+	if (this.x > cellInfo.width * 5) {
+		this.x = -cellInfo.width;
+		this.speed = generateEnemySpeed();
+	}
+	this.checkCollision();
+};
+
+Enemy.prototype.checkCollision = function () {
+	if (this.x + enemyInfo.width >= player.x && this.x < player.x + enemyInfo.width && this.y === player.y) {
+		this.win();
+	}
+};
+
+Enemy.prototype.win = function () {
+	alert("Bugs win\n(You DIE...)");
+	player.resetGame();
+};
+
+const Player = function (x, y) {
+	Being.call(this, x, y, playerInfo.sprite);
+};
+
+Player.prototype = Object.create(Being.prototype);
+
+Player.prototype.update = function () {};
+
+Player.prototype.win = function () {
+	setTimeout(() => {
+		alert("You win YEAH!");
+		this.resetGame();
+	}, 100);
+};
+
+Player.prototype.resetGame = function () {
+	player.x = playerInfo.startPositionX;
+	player.y = playerInfo.startPositionY;
+};
+
+Player.prototype.updatePosition = function (newX, newY) {
+	if (newX >= 0 && newX <= playingAreaInfo.width) {
+		this.x = newX;
+	}
+	if (newY >= 0 && newY <= playingAreaInfo.height) {
+		this.y = newY;
+	}
+	if (newY < 0) {
+		this.y = newY;
+		this.win();
+	}
+};
+
+Player.prototype.handleInput = function (keycode) {
+	let newX = this.x,
+		newY = this.y;
+	switch (keycode) {
+		case "up":
+			newY = this.y - cellInfo.height;
+			break;
+		case "right":
+			newX = this.x + cellInfo.width;
+			break;
+		case "down":
+			newY = this.y + cellInfo.height;
+			break;
+		case "left":
+			newX = this.x - cellInfo.width;
+			break;
+		default:
+			break;
+	}
+	this.updatePosition(newX, newY);
+};
+
+const enemyStartPostionsY = [];
+for (let i = 1; i <= 3; i++) {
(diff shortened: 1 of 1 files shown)


submissions/f3d0t/js-oop/app.js
```diff
@@ -0,0 +1,139 @@
...
+	}
+	this.updatePosition(newX, newY);
+};
+
+const enemyStartPostionsY = [];
+for (let i = 1; i <= 3; i++) {
+	enemyStartPostionsY.push(cellInfo.height * i - cellInfo.beingPositionCorrection);
+}
+enemyStartPostionsY.forEach((posY) => {
+	let enemy = new Enemy(-enemyInfo.width, posY);
```
babayK0, line 124: 1. typo;
2. use const if you want to declare variable.
One more point, there is no need to declare this var on every loop.
```suggestion
	const enemy = new Enemy(enemyInfo.width, posY);
```

submissions/f3d0t/js-oop/app.js
```diff
@@ -0,0 +1,139 @@
...
+			break;
+	}
+	this.updatePosition(newX, newY);
+};
+
+const enemyStartPostionsY = [];
+for (let i = 1; i <= 3; i++) {
+	enemyStartPostionsY.push(cellInfo.height * i - cellInfo.beingPositionCorrection);
+}
+enemyStartPostionsY.forEach((posY) => {
```
babayK0, line 123: Why not use `Array.map` ?

submissions/f3d0t/js-oop/app.js
```diff
@@ -0,0 +1,139 @@
...
+		default:
+			break;
+	}
+	this.updatePosition(newX, newY);
+};
+
+const enemyStartPostionsY = [];
+for (let i = 1; i <= 3; i++) {
+	enemyStartPostionsY.push(cellInfo.height * i - cellInfo.beingPositionCorrection);
+}
```
babayK0, line 122: map ?

babayK0 (changes requested):
Well done!
It's nice that you're trying to finish before deadline. (A lot of open PR in the last few days :+1: ) This is a huge result. But take your time, in real work it is better to provide clean and well-designed code so that avoid unexpected 'bugs' in the future. Exceptions happen, but try not to hurry in the future, it will save you and others time, thank you :smile: 

Please, confirm all items below are positive, having the code fixed as appropriate. Please, ask a question if anything below requires clarification.

Minimal requirements to meet:
 * Requirements re **OOP**:
 * Most common mistakes

submissions/f3d0t/js-oop/app.js
f3d0t: cool suggestion

submissions/f3d0t/js-oop/app.js
f3d0t: nope, there is empty array in beginning

f3d0t: oh, now i see, thanks

submissions/f3d0t/js-oop/app.js
babayK0: @f3d0t
my bad. Seems like you want to create enemy outside of canvas. Please provide more obvious variable 

example : `enemyInfo.startPositionX = -enemyInfo.width`

submissions/f3d0t/js-oop/app.js
```diff
@@ -116,17 +116,11 @@ Player.prototype.handleInput = function (keycode) {
...
-	enemyStartPostionsY.push(cellInfo.height * i - cellInfo.beingPositionCorrection);
-}
-enemyStartPostionsY.forEach((posY) => {
-	let enemy = new Enemy(-enemyInfo.width, posY);
-	allEnemies.push(enemy);
-});
-
 const player = new Player(playerInfo.startPositionX, playerInfo.startPositionY);
 
+const enemyStartPostionsY = [1, 2, 3].map((rowNumber) => cellInfo.height * rowNumber - cellInfo.beingPositionCorrection);
```
babayK0, line 121: Please made creation more automatically. What you will do if you will need 100500 bugs. 

`Array.from({ length: })` can help you i guess

submissions/f3d0t/js-oop/app.js
```diff
@@ -18,9 +18,8 @@ const playingAreaInfo = {
 	width: cellInfo.width * 4,
 	height: cellInfo.height * 5,
```
babayK0: You have access to `ctx` variable from engine.js.  If you have access to dynamically calculated value - use it instead of this ones.
```
/* Assign the canvas' context object to the global variable (the window
   * object when run in a browser) so that developers can use it more easily
   * from within their app.js files.
   */
  global.ctx = ctx;
  ```

**Review by babayK0 — CHANGES_REQUESTED** [review]

File: submissions/f3d0t/js-oop/app.js (hunk 5 above)
Comment (f3d0t) [reply submissions/f3d0t/js-oop/app.js]: I see that window.ctx returns undefined..

File: submissions/f3d0t/js-oop/app.js (hunk 4 above)
Comment (f3d0t) [reply submissions/f3d0t/js-oop/app.js]: This way is really cool

File: submissions/f3d0t/js-oop/app.js [hunk 6]
```diff
@@ -0,0 +1,133 @@
...
+			break;
+		default:
+			break;
+	}
+	this.updatePosition(newX, newY);
+};
+
+const player = new Player(playerInfo.startPositionX, playerInfo.startPositionY);
+
+const enemyStartPostionsY = Array.from({ length: 3 }, (v, rowNum) => rowNum + 1).map((rowNumber) => cellInfo.height * rowNumber - cellInfo.beingPositionCorrection);
```
babayK0, line 122: You can do the same with shorter code.

Parameters of [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)
```
 1. arrayLike
    An array-like or iterable object to convert to an array.
    
 2. mapFn Optional
    Map function to call on every element of the array.
```

Also change naming this is not `rowNum` and why you call variable  `v` ?  
```suggestion
const enemyStartPostionsY = Array.from({ length: 3 }, (element, index) => cellInfo.height * (index + 1) - cellInfo.beingPositionCorrection);
```

Why you always use word `being` ?) 
It is don't make any sense i guess. Looks like wrong translation or misuse of it. Explain meaning of that word please)

submissions/f3d0t/js-oop/app.js
```diff
@@ -0,0 +1,133 @@
...
+const playerInfo = {
+	sprite: "images/char-boy.png",
+	startPositionX: cellInfo.width * 2,
+	startPositionY: cellInfo.height * 5 - cellInfo.beingPositionCorrection,
+};
+const playingAreaInfo = {
+	width: cellInfo.width * 5,
+	height: cellInfo.height * 5,
+};
+
```
babayK0, line 22: By convention, the constant identifiers are in uppercase

```
const CONSTANT_NAME = value;
```

babayK0 (changes requested):

submissions/f3d0t/js-oop/app.js
f3d0t: @babayK0 
being - noun (Cambridge Academic Content Dictionary ©)
a person or thing that exists, or the state of existing:
[ C ] human/living beings
For me it is some kind living existence... (Существо по-русски)

babayK0: In my opinion, this word is not very common, especially for not native speakers, in this sense, the word in principle is rarely used) So i suggest you use more obvious word like “creature” or something else.

f3d0t: Thank You for your suggestion) Creature word sounds definitely more clearly

f3d0t:
Bugs and mistakes are now fixed=) check pls

f3d0t:
@babayK0 Are you here?...=)

babayK0:
@f3d0t yep, sorry for long response

babayK0 (approved):
LGTM!

PR #402, opened 2020-12-31
OOP-exercise

kasionio:
### oop-exercise

[Demo](https://kasionio.github.io/a-tiny-JS-world)
[Code base](https://github.com/kasionio/a-tiny-JS-world/blob/gh-pages/index.js)

Please, review.

### DIFF
--- a/submissions/kasionio/oop-exercise/index.js
+++ b/submissions/kasionio/oop-exercise/index.js
@@ -17,7 +17,7 @@ class Inhabitant {
       this.friends = [];
   }
 
-  addFriends = (...friendsInstances) => this.friends = [...this.friends, ...friendsInstances];
+  addFriends = (...friendsInstances) => this.friends.includes(...friendsInstances) ? false : this.friends = [...this.friends, ...friendsInstances];
 
   getFriendsNames = () => this.friends.length > 0 ? this.friends.map(friend => friend.name) : 'forever alone...';
 
@@ -33,8 +33,8 @@ class HumanBeing extends Inhabitant {
       super(species, name, gender, 2, saying);
       this.hands = 2;
   }
-  getOutput() {
-      return super.getOutput() + `; hands: ${this.hands};`
+  toString() {
+      return super.toString() + `; hands: ${this.hands};`
   }
 }
 
@@ -98,3 +98,4 @@ catWoman.addFriends(cat);
 const inhabitants = [man, woman, catWoman, cat, dog];
 
 inhabitants.forEach(inhabitant => print(inhabitant));
+
--- a/submissions/kasionio/oop-exercise/index.js
+++ b/submissions/kasionio/oop-exercise/index.js
@@ -8,56 +8,56 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 class Inhabitant {
-  constructor(species, name, gender, legs, hands, saying) {
-  this.species = species,
-  this.name = name,
-  this.gender = gender,
-  this.legs = legs,
-  this.hands = hands,
-  this.saying = saying
+  constructor(species, name, gender, legs, saying) {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.legs = legs;
+      this.saying = saying;
   }
 }
 
 class Human extends Inhabitant {
-  constructor(name, gender, saying) {
-    super('human', name, gender, 2, 2, saying);
-    }
+  constructor(name, gender, saying, hands) {
+      super('human', name, gender, 2, saying);
+      this.hands = 2;
+  }
 }
 
 class Man extends Human {
   constructor(name, saying) {
-    super(name, 'male', saying);
-    }
+      super(name, 'male', saying);
+  }
 }
 
 class Woman extends Human {
   constructor(name, saying) {
-    super(name, 'female', saying);
-    }
+      super(name, 'female', saying);
+  }
 }
 
 class Animal extends Inhabitant {
   constructor(species, name, gender, saying) {
-    super(species, name, gender, 4, 0, saying);
-    }
+      super(species, name, gender, 4, saying);
+  }
 }
 
 class Dog extends Animal {
   constructor(name, gender) {
-    super('dog', name, gender, 'woof-woof');
-    }
+      super('dog', name, gender, 'woof-woof');
+  }
 }
 
 class Cat extends Animal {
   constructor(name, gender) {
-    super('cat', name, gender, 'meow');
-    }
+      super('cat', name, gender, 'meow');
+  }
 }
 
 class CatWoman extends Inhabitant {
   constructor(name) {
-    super('catWoman', name, 'female', 2, 2, cat.saying);
-    }
+      super('catWoman', name, 'female', 2, cat.saying);
+  }
 }
 
 let man = new Man('Arnold', 'Hasta la vista, baby!');
@@ -66,5 +66,11 @@ let cat = new Cat('Bobby', 'male');
 let dog = new Dog('Toby', 'male');
 let catWoman = new CatWoman('Betty');
 
-const inhabitants = [man, woman, catWoman, cat, dog]; 
-inhabitants.forEach(item => print(Object.values(item)));
+const inhabitants = [man, woman, catWoman, cat, dog];
+const properties = ['species', 'name', 'gender', 'legs', 'hands', 'saying'];
+
+const message = (inhabitants) => inhabitants.map(inhabitant => properties.map(prop => inhabitant[prop] ? inhabitant[prop] : 'undefined')
+  .join('; '))
+  .join('\n');
+
+print(message(inhabitants));
--- a/submissions/kasionio/oop-exercise/index.js
+++ b/submissions/kasionio/oop-exercise/index.js
@@ -17,9 +17,14 @@ class Inhabitant {
       this.friends = [];
   }
 
-  addFriends = (...friendsInstances) => this.friends.includes(...friendsInstances) ? false : this.friends = [...this.friends, ...friendsInstances];
+  addFriends(...friendsInstances) {
+      this.friends = [...this.friends, ...friendsInstances.filter(friend => !this.friends.includes(friend))];
+  }
 
-  getFriendsNames = () => this.friends.length > 0 ? this.friends.map(friend => friend.name) : 'forever alone...';
+  getFriendsNames() {
+       let output = this.friends.length > 0 ? this.friends.map(friend => friend.name) : 'alone';     
+       return output;
+  }
 
   toString() {
       const properties = ['species', 'name', 'gender', 'legs', 'saying'];
@@ -98,4 +103,3 @@ catWoman.addFriends(cat);
 const inhabitants = [man, woman, catWoman, cat, dog];
 
 inhabitants.forEach(inhabitant => print(inhabitant));
-
--- a/submissions/kasionio/oop-exercise/index.js
+++ b/submissions/kasionio/oop-exercise/index.js
@@ -1,0 +1,70 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/kasionio/a-tiny-JS-world/tree/gh-pages
+   Web app: https://kasionio.github.io/a-tiny-JS-world
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying) {
+  this.species = species,
+  this.name = name,
+  this.gender = gender,
+  this.legs = legs,
+  this.hands = hands,
+  this.saying = saying
+  }
+}
+
+class Human extends Inhabitant {
+  constructor(name, gender, saying) {
+    super('human', name, gender, 2, 2, saying);
+    }
+}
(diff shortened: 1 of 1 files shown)


submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(species, name, gender, legs, hands, saying) {
+  this.species = species,
+  this.name = name,
+  this.gender = gender,
+  this.legs = legs,
+  this.hands = hands,
```
OleksiyRudenko, line 16: Why commas?

OleksiyRudenko, line 16: Not all inhabitants have hands by natural design. The beauty of OOP is that every base class defines and handles only the props that are shared by its descendants, and every child class introduces and handles props that are unique to it and its descendants.

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+let man = new Man('Arnold', 'Hasta la vista, baby!');
+let woman = new Woman('Kate', 'Merry Christmas!');
+let cat = new Cat('Bobby', 'male');
+let dog = new Dog('Toby', 'male');
+let catWoman = new CatWoman('Betty');
+
+const inhabitants = [man, woman, catWoman, cat, dog]; 
+inhabitants.forEach(item => print(Object.values(item)));
```
OleksiyRudenko, line 70: https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md#additional-requirements are still valid.

Also let classes handle their own representation.

OleksiyRudenko (changes requested):
@kasionio pretty well done
However there is still some room for improtant improvements

submissions/kasionio/oop-exercise/index.js
kasionio: My bad. Accidentally It comes from the objects

kasionio: Thanks, now I understand. This time I delete `hands` property from the parent class

submissions/kasionio/oop-exercise/index.js
kasionio: Re-write this part. But not sure that the code is correct for the requirements

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+class Cat extends Animal {
+  constructor(name, gender) {
+      super('cat', name, gender, 'meow');
+  }
+}
+
+class CatWoman extends Inhabitant {
+  constructor(name) {
+      super('catWoman', name, 'female', 2, cat.saying);
```
OleksiyRudenko, line 59: What if `cat` doesn't exist yet?
Besides we want that a cat-woman borrows saying not from a prticular cat but rather from all cats.
Take look into static class properties/methods (and define one in `Cat` class). These are available at all times, exists a particular class instance or not.

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,76 @@
...
+let dog = new Dog('Toby', 'male');
+let catWoman = new CatWoman('Betty');
+
+const inhabitants = [man, woman, catWoman, cat, dog];
+const properties = ['species', 'name', 'gender', 'legs', 'hands', 'saying'];
+
+const message = (inhabitants) => inhabitants.map(inhabitant => properties.map(prop => inhabitant[prop] ? inhabitant[prop] : 'undefined')
+  .join('; '))
+  .join('\n');
+
```
OleksiyRudenko, line 75: Let classes handle their own presentation. This is how OOP/SOLID principles get used at their full power.
You will need method overloading so that the top-most class handles its own props and `Human` class adds what is unique to it.
Hint: `super.methodName()`

OleksiyRudenko (changes requested):
@kasionio pretty well done!
Let's make the code more OOP/SOLID compliant.

submissions/kasionio/oop-exercise/index.js
kasionio: Thanks for the explanation! It's difficult for me when I go from theory to practice.

submissions/kasionio/oop-exercise/index.js
kasionio: Added a `HumanBeing` class cause of CatWoman. She has 2 hands and 2 legs as a human, but at the same time has another `species` value.

OleksiyRudenko (changes requested):
@kasionio please implement support for friends

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(species, name, gender, legs, saying, friends = 'forever alone') {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.legs = legs;
+      this.saying = saying;
+      this.friends = friends;
```
OleksiyRudenko, line 17: Can we re-use `addFriends` method here for consistency?
Besides within this class `friends` are treated as string in `constructor` and objects in `addFriends`. Confusing.

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/kasionio/a-tiny-JS-world/tree/gh-pages
+   Web app: https://kasionio.github.io/a-tiny-JS-world
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(species, name, gender, legs, saying, friends = 'forever alone') {
```
OleksiyRudenko, line 11: Neither of immediate child classes of `Inhabitant` ever pass friends via `super`. Is it a redundant param?

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.legs = legs;
+      this.saying = saying;
+      this.friends = friends;
+  }
+
+  addFriends(...friends)  {
+    this.friends = [...friends].map(friend => friend.name).join(', ');
```
OleksiyRudenko, line 21: If I do e.g. `cat.addFriends(dog, man); cat.addFriends(woman, dog);` will friends be really **added**?

What will happen if I do e.g. `cat.addFriends([dog, man])` instead of `cat.addFriends(dog, man)`? Will the code work as expected?

If we pack method arguments in an array (`addFriends(...friends) {...}`) why do we need unpacking array and packing it back into array (`[...friends]`) in the function body?

OleksiyRudenko (changes requested):
@kasionio good job. You will find some questions to reflect on and fix based on the insights from the reflection

submissions/kasionio/oop-exercise/index.js
kasionio: Define friends property as an empty array inside the class

submissions/kasionio/oop-exercise/index.js
kasionio: Thanks! Now I realize that my code just made the output but wasn't really adding friends. Fixed the logic. Test in the browser console. works fine now :)

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,108 @@
...
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.legs = legs;
+      this.saying = saying;
+      this.friends = [];
+  }
+
+  addFriends(...friendsInstances) {
+      this.friends.push(...friendsInstances);
```
OleksiyRudenko, line 21: JFYI
`someArray = [...someArray, ...extraItems]` adds extra items as well

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  getFriendsNames() {
+      if (this.friends.length > 0) {
+          return this.friends.map(friend => friend.name)
+      } else {
+          return this.friends = 'forever alone...'
+      }
```
OleksiyRudenko, line 29: 1. `return this.friends = 'forever alone...'`: try the following:
```
print(woman.getOutput());
print(woman.getOutput());
```
You will see the issue caused by this fragment.

2. The entire method body is a typical case for the ternary operator. Check MDN for details.
3. Will you take a challenge to master the ternary?

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,108 @@
...
+let catWoman = new CatWoman('Betty');
+
+man.addFriends(woman, cat);
+woman.addFriends(man, cat);
+dog.addFriends(man);
+catWoman.addFriends(cat);
+
+const inhabitants = [man, woman, catWoman, cat, dog];
+
+inhabitants.forEach(inhabitant => print(inhabitant.getOutput()))
```
OleksiyRudenko, line 108: Bonus:
If you rename `getOutput` to `toString` the latter will be called implictly in string context. `print` ensures string context for its arguments.
So `print(inhabitant)` will be effectively equivalent to `print(inhabitant.toString())` yet cleaner

OleksiyRudenko (changes requested):
@kasionio well done.
Yet some room for improvement.

submissions/kasionio/oop-exercise/index.js
kasionio: Nice hack! Thanks!

submissions/kasionio/oop-exercise/index.js
kasionio: 1. Everything looks as expected to me now.
It prints `PRINT: species: human; name: Kate; gender: female; legs: 2; saying: Merry Christmas!; friends: Arnold,Bobby` every time. 
2. Change to ternary

submissions/kasionio/oop-exercise/index.js
kasionio: Thanks! The spread operator looks better. 

Try to make code more ES6 looking.

submissions/kasionio/oop-exercise/index.js
```diff
@@ -0,0 +1,101 @@
...
+      this.legs = legs;
+      this.saying = saying;
+      this.friends = [];
+  }
+
+  addFriends = (...friendsInstances) => this.friends.includes(...friendsInstances) ? false : this.friends = [...this.friends, ...friendsInstances];
+
+  getFriendsNames = () => this.friends.length > 0 ? this.friends.map(friend => friend.name) : 'forever alone...';
+
+  toString() {
```
OleksiyRudenko, line 24: 1. Mixing style of methods definition is not really a good style. Please fix.

2. `addFriends` method implementation:
```
this.friends.includes(...friendsInstances) ? false : this.friends = [...this.friends, ...friendsInstances];
```
Really hectic fragment.
The method returns false if all new friends are already in the friend list. It will add friends if they are all new and also return an updated friends list. So it returns boolean under condition and an array (potentially empty) in another case. This inconsistency will make the receiving party to add extra business logic to determine if the operation was successful.

We'd rather want just add any new friends.

Consider the following example:
`list = [...list, ...(newEntries.filter(newEntry => list.includes(entry)))]`
Please comprehend it before embedding. Feel free asking questions.

3. Ternary
Any assignments within the ternary like `condition ? variable = value1 : variable = value2`(and variations thereof)  is an anti-pattern. The pattern is to assign the entire ternary result to a variable (or return, or use otherwise).
Please check the proper use of ternary operator on MDN.

OleksiyRudenko (changes requested):
@kasionio good improvement.
Yet code style consistency is important. Please also check the comment below

submissions/kasionio/oop-exercise/index.js
OleksiyRudenko: Not necessarily assign before return. Assignment can be redundant of the value returned immediately after.
The point was not to have any assignments **inside** the ternary. `... ? .... : var = value` is a an anti-pattern.

kasionio: Thank you! Ternary is a bit confusing to me. I understand that we need assign the whole ternary to variable or have some functions inside ternary. At this case if condition `this.friends.length > 0` return false can we just return string value `alone` without assigning? Is it valid?
 `return (this.friends.length > 0 ? this.friends.map(friend => friend.name) : 'alone')`

OleksiyRudenko: @kasionio 
For consistency, it is good that a function returns only certain type. So, make that in `true` case it also returns a string.
BTW, outer parenthses are not required in this example. It doesn't add to readability.

OleksiyRudenko: @kasionio Make return statement consistent in types of returnable values and we are done

kasionio: Add `toString()` method in `true` case of ternary .

Thanks a lot! It was very helpful for me.

OleksiyRudenko (approved):
@kasionio well done!
P.S. A better approach is not to rely on built-in methods. Consider using `join` explicitily with some sensible delimiter in similar scenarios

PR #407, opened 2021-01-02
OOP-exercise 

SamVal007:
A tiny js world
Please, review

[demo](https://samval007.github.io/a-tiny-JS-world/)
[codebase](https://github.com/SamVal007/a-tiny-JS-world)

Please, review.


submissions/SamVal007/Tiny-JS-OOP/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+class Cat extends Person {
+  constructor(name, gender, speach, legs = 4) {
+    super("cat", name, gender, speach, legs);
+  }
+}
+class Human extends Person {
+  constructor(name, gender, speach, friends, legs = 2, hands = 2) {
+    super("human", name, gender, speach, friends, legs, hands);
```
OleksiyRudenko, line 44: What `super` call actually does?
What wrong here and how could you fix this?

submissions/SamVal007/Tiny-JS-OOP/index.js
```diff
@@ -0,0 +1,63 @@
...
+    this.legs = legs;
```
OleksiyRudenko, line 45: Why do we need to initialize `this.legs` here?

submissions/SamVal007/Tiny-JS-OOP/index.js
```diff
@@ -0,0 +1,63 @@
...
+class Person {
+  constructor(species, gender, name, speach, legs) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.speach = speach;
+    this.legs = legs;
+  }
+
+  saySpeach() {
```
OleksiyRudenko, line 19: How this method name translates?
Can you devise a better name for it?

submissions/SamVal007/Tiny-JS-OOP/index.js
```diff
@@ -0,0 +1,63 @@
...
+class Human extends Person {
+  constructor(name, gender, speach, friends, legs = 2, hands = 2) {
+    super("human", name, gender, speach, friends, legs, hands);
+    this.legs = legs;
+    this.hands = hands;
+
+  }
+
+  saySpeach() {
+    return `${super.saySpeach()} I hvae ${this.hands} hands.`
```
OleksiyRudenko, line 51: Please fix typos.

OleksiyRudenko (changes requested):
@SamVal007 pretty nice job
However there are few questions (please answer those) and a couple of inconsistencies to fix

submissions/SamVal007/Tiny-JS-OOP/index.js
SamVal007: As I understand, when we create a child class by extending parent class, we call super() in order to initialize general fields and avoid code duplication. Custom fields from child we need initialize manually using keyword "this".

submissions/SamVal007/Tiny-JS-OOP/index.js
SamVal007: Yep, we already don't need it here! 
Because already initialized it in super("human", name, gender, speech, friends, legs, hands);

submissions/SamVal007/Tiny-JS-OOP/index.js
SamVal007: Yep! I found, that correct to write "speech" instead of "speach" :) 
but let's call this method just 'say()' - i think it's better.

SamVal007:
I've made changes. 
Please, re-review:)

submissions/SamVal007/Tiny-JS-OOP/index.js
```diff
@@ -0,0 +1,69 @@
...
+    return `${super.say()} Pets are friendly to all!`
+  }
+}
+class Cat extends Person {
+  constructor(name, gender, speech) {
+    super("cat", name, gender, speech, 4);
+  }
+  say() {
+    return `${super.say()} Pets are friendly to all!`
+  }
```
OleksiyRudenko, line 47: Code repetition. Imagine you have more kinds of pets.
Introduce the `Pet` class that will generalize shared methods from `Dog` and `Cat`.

OleksiyRudenko (changes requested):
@SamVal007 well done
Need to make code DRYer

submissions/SamVal007/Tiny-JS-OOP/index.js
SamVal007: Ok. Bit Thanks!
If I understand correct: better to make one more class Pet which will a child of Person and and will parent for Cat and Dog. Right?:)

OleksiyRudenko: Yep, the rule is whatever is shared by two classes should be generalized into their common parent class, inserted into the hyrerachy if needed.

SamVal007: Ok! I got it!
Thank You!

OleksiyRudenko (approved):

PR #408, opened 2021-01-02
memory-pair-game

LGeorge007:
### memory-pair-game

[Demo](https://lgeorge007.github.io/memory-pair-game/)
[Code base](https://github.com/LGeorge007/memory-pair-game)

Please, review.

### DIFF
--- a/submissions/LGeorge007/memory-pair-game/script.js
+++ b/submissions/LGeorge007/memory-pair-game/script.js
@@ -1,11 +1,11 @@
-let arrOfUnits = ["img/cartman.jpg",
+const arrOfUnits = ["img/cartman.jpg",
     "img/chief.jpg",
     "img/sten.jpg",
     "img/kyle.jpg",
     "img/kenny.jpg"
 ];
-let imgCover = `img/cover.jpg`;
-let gameField = document.querySelector("#gameField");
+const imgCover = `img/cover.jpg`;
+const gameField = document.querySelector("#gameField");
 let isLocked = false;
 let firstCard;
 let secondCard;
@@ -13,13 +13,13 @@ let isSecondCard = false;
 let unopenedCards;
 
 //feed him array & any even number
-let duplicateElements = function(array, times) {
+const duplicateElements = function(array, times) {
     return array.reduce((res, current) => {
         return res.concat(Array(times).fill(current));
     }, []);
 };
 
-let shuffle = function(arr) {
+const shuffle = function(arr) {
     const result = [];
     while (arr.length > 0) {
         let random = Math.floor(Math.random() * arr.length);
@@ -70,7 +70,7 @@ const checkCard = function({ target }) {
     };
 };
 
-let renderCard = function(arr) {
+const renderCard = function(arr) {
     for (let i = 0; i < arr.length; i++) {
         let card = document.createElement("div");
         card.setAttribute("id", i);
--- a/submissions/LGeorge007/memory-pair-game/script.js
+++ b/submissions/LGeorge007/memory-pair-game/script.js
@@ -1,0 +1,65 @@
+const allCards = document.querySelectorAll(".flipper");
+const arrOfCards = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19];
+let unopenedCards = arrOfCards.length;
+let firstCard;
+let secondCard;
+let clikCounter = 0;
+
+const checkCard = function (event) {
+    const checkMatch = function () {
+        if (clikCounter == 0) {
+            firstCard = event.path[1];
+            clikCounter++;
+        } else if (clikCounter == 1) {
+            secondCard = event.path[1];
+            //The same card must not be removed
+            if (firstCard.style.order != secondCard.style.order) {
+                if (firstCard.dataset.hero == secondCard.dataset.hero) {
+                    firstCard.children[1].classList.add("hiddenSide");
+                    secondCard.children[1].classList.add("hiddenSide");
+                    firstCard.classList.add("hiddenFlipper");
+                    secondCard.classList.add("hiddenFlipper");
+                    unopenedCards -= 2;
+                } else {
+                    firstCard.classList.remove("rotateCard");
+                    secondCard.classList.remove("rotateCard");
+                };
+                clikCounter--;
+                if (unopenedCards == 0) {
+                    setTimeout(() => {
+                        alert("You win!");
+                        window.location.reload();
+                    }, 1500);
+                };
+            };
+        };
+    };
+    event.path[1].classList.toggle("rotateCard");
+    setTimeout(checkMatch, 600);
+}
+
+const shuffle = function (arr) {
+    const result = [];
+    const getRandomInt = function (min, max) {
+        return Math.floor(Math.random() * (max - min + 1)) + min;
+    };
+    while (arr.length > 0) {
+        let random = getRandomInt(0, arr.length - 1);
+        let elem = arr.splice(random, 1)[0];
+        result.push(elem);
+    }
+    return result;
+}
+
+const setOrder = function (order) {
+    for (let i = 0; i < allCards.length; i++) {
+        allCards[i].style.order = order[i];
+    }
+}
+
+const onStart = function () {
+    setOrder(shuffle(arrOfCards));
+    allCards.forEach(card => card.addEventListener('click', checkCard));
+}
+
+document.addEventListener("DOMContentLoaded", onStart);
--- a/submissions/LGeorge007/memory-pair-game/script.js
+++ b/submissions/LGeorge007/memory-pair-game/script.js
@@ -1,65 +1,104 @@
-const allCards = document.querySelectorAll(".flipper");
-const arrOfCards = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19];
-let unopenedCards = arrOfCards.length;
+let arrOfUnits = ["img/cartman.jpg",
+    "img/chief.jpg",
+    "img/sten.jpg",
+    "img/kyle.jpg",
+    "img/kenny.jpg"
+];
+let imgCover = `img/cover.jpg`;
+let gameField = document.querySelector("#gameField");
+let isLocked = false;
 let firstCard;
 let secondCard;
-let clikCounter = 0;
+let isSecondCard = false;
+let unopenedCards;
 
-const checkCard = function (event) {
-    const checkMatch = function () {
-        if (clikCounter == 0) {
-            firstCard = event.path[1];
-            clikCounter++;
-        } else if (clikCounter == 1) {
-            secondCard = event.path[1];
-            //The same card must not be removed
-            if (firstCard.style.order != secondCard.style.order) {
-                if (firstCard.dataset.hero == secondCard.dataset.hero) {
-                    firstCard.children[1].classList.add("hiddenSide");
-                    secondCard.children[1].classList.add("hiddenSide");
-                    firstCard.classList.add("hiddenFlipper");
-                    secondCard.classList.add("hiddenFlipper");
-                    unopenedCards -= 2;
-                } else {
-                    firstCard.classList.remove("rotateCard");
-                    secondCard.classList.remove("rotateCard");
-                };
-                clikCounter--;
-                if (unopenedCards == 0) {
-                    setTimeout(() => {
-                        alert("You win!");
-                        window.location.reload();
-                    }, 1500);
-                };
-            };
-        };
-    };
-    event.path[1].classList.toggle("rotateCard");
-    setTimeout(checkMatch, 600);
-}
+//feed him array & any even number
+let duplicateElements = function(array, times) {
+    return array.reduce((res, current) => {
+        return res.concat(Array(times).fill(current));
+    }, []);
+};
 
-const shuffle = function (arr) {
+let shuffle = function(arr) {
     const result = [];
-    const getRandomInt = function (min, max) {
-        return Math.floor(Math.random() * (max - min + 1)) + min;
-    };
     while (arr.length > 0) {
-        let random = getRandomInt(0, arr.length - 1);
+        let random = Math.floor(Math.random() * arr.length);
         let elem = arr.splice(random, 1)[0];
         result.push(elem);
     }
     return result;
+};
+
+const checkCard = function({ target }) {
+    if (!isLocked) {
+        if (isSecondCard) {
+            isLocked = true;
+            secondCard = target.closest("div");
+            secondCard.classList.toggle("rotateCard");
+            //The same card must not be removed
+            if (firstCard.id !== secondCard.id) {
+                if (firstCard.dataset.hero === secondCard.dataset.hero) {
+                    unopenedCards -= 2;
+                    setTimeout(() => {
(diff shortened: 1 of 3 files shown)


submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
+const allCards = document.querySelectorAll(".flipper");
+const arrOfCards = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19];
+let unopenedCards = arrOfCards.length;
+let firstCard;
+let secondCard;
+let clikCounter = 0;
+
+const checkCard = function (event) {
+    const checkMatch = function () {
+        if (clikCounter == 0) {
```
yaripey, line 10: Please use `===` instead of `==` everywhere.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
+const allCards = document.querySelectorAll(".flipper");
+const arrOfCards = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19];
+let unopenedCards = arrOfCards.length;
+let firstCard;
+let secondCard;
+let clikCounter = 0;
+
+const checkCard = function (event) {
+    const checkMatch = function () {
```
yaripey, line 9: There is no point for function `checkMatch` to be declared inside function `checkCard`. Either remove the function completely and place it's code inside `checkCard`'s body, or move `checkMatch` outside of `checkCard`'s body. In your case, every time player clicks a card a new instance of `checkMatch` is created in the memory, which is not good for performance.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
...
+        if (clikCounter == 0) {
+            firstCard = event.path[1];
```
yaripey, line 11: Will only comment on this once to shorten the review, you can use `target` to get to the element you need. Also, if you use `.closest()` method you will ensure yourself with a right element, so you don't accidentally add a class to a wrong element, for example.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
...
+            clikCounter++;
+        } else if (clikCounter == 1) {
+            secondCard = event.path[1];
+            //The same card must not be removed
+            if (firstCard.style.order != secondCard.style.order) {
```
yaripey, line 16: Scripting based on CSS rules is bad in general. If you wan't to tie your html cards with javascript it is better to use `ID`s.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
...
+
+const setOrder = function (order) {
+    for (let i = 0; i < allCards.length; i++) {
+        allCards[i].style.order = order[i];
+    }
+}
+
+const onStart = function () {
+    setOrder(shuffle(arrOfCards));
+    allCards.forEach(card => card.addEventListener('click', checkCard));
```
yaripey, line 62: Adding event listeners for each card is bad for performance. It is better to add one event listener for `click` on the `document` or some sort of `container`. Your pressed card will still be in a `target` of event.

yaripey (commented):
On one hand the code is short, but on the other hand it is hardly scalable. For example, if you'll ever want to add more card types or change the amount of cards on the screen you'll have to update both HTML and JS files.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: How can I organize cards using ID? 
Using flexbox, I may apply order property to organize cards.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: == is faster than ===, isn't it?

yaripey: `===` is actually faster, because if types are different it just returns `false`. If types are different and the operator is `==`, than JavaScript has to do different type operations to be able to compare two values.

submissions/LGeorge007/memory-pair-game/script.js
yaripey: That is why your way of dealing with cards has a lot of flaws. You already learned methods for modifying the html structure using JS code. Wouldn't it be better to create a template in JavaScript and create cards layout using JavaScript array methods?

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
+const allCards = document.querySelectorAll(".flipper");
+const arrOfCards = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19];
```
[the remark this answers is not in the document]
LGeorge007: I agree with you completely.
Fixed.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
...
+let unopenedCards = arrOfCards.length;
+let firstCard;
+let secondCard;
+let clikCounter = 0;
```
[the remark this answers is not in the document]
LGeorge007: I use variable isSecondCard that has boolean type.
Fixed.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,65 @@
...
+
+const checkCard = function (event) {
```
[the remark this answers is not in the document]
LGeorge007: Fixed.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: Fixed it.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: Fixed it.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: Fixed. I use ID="flipper" isntead .flipper.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: I agree. This is a good concept. 
But, I tried to hang the handler function on the entire gameField.
And Side effect appeared - gameField started to mirror.( When you clicked  near by flipper).
I rejected this model for this task. And I hang the handle function, when I create document Fragment.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: I am creating cards using a documentFragment.
This allows  to have a consistent structure for div class = "flipper". 
And I am referring to the children of the DIV(.flipper), using the following syntax.
firstCard.children[1]
secondCard.children[1]

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: Unfortunately, I cannot use one handler for the entire GameField. 
GameField starts to behave like a "mirror"

yaripey: That's because you need an additional check. If the player clicked the card, than do the styling. If player clicked field instead - just ignore the click and do nothing.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,104 @@
...
+        backSide.classList.add("back");
+
+        card.appendChild(frontSide);
+        card.appendChild(backSide);
+        gameField.appendChild(card);
+    };
+};
+
+const init = function() {
+    arrOfUnits = shuffle(duplicateElements(arrOfUnits, 4));
```
yaripey, line 99: `arrOfUnits` is a `const`. You cannot reassign constants.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,104 @@
...
+        result.push(elem);
+    }
+    return result;
+};
+
+const checkCard = function({ target }) {
+    if (!isLocked) {
+        if (isSecondCard) {
+            isLocked = true;
+            secondCard = target.closest("div");
```
yaripey, line 36: That is why your field starts 'mirroring' if clicked. You are searching for a `div`. Your field is `div` too. `.closest()` works with any selectors, so you can search for `.flipper` instead.

yaripey (commented):
Your code doesn't work because of constant reassigning.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: Fixed.

submissions/LGeorge007/memory-pair-game/script.js
LGeorge007: fixed it.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,104 @@
...
+        let backSide = document.createElement("img");
+        backSide.setAttribute("src", arr[i]);
+        backSide.setAttribute("draggable", false);
+        backSide.classList.add("back");
+
+        card.appendChild(frontSide);
+        card.appendChild(backSide);
+        gameField.appendChild(card);
+    };
+};
```
yaripey, line 96: There's a lot to this function.

Firstly, there's no point in using `let` here. You can still mutate constants, so it's better to use them here. Remember to __always__ use constants and use variables only if you are sure that you need them.

Secondly, the html structure of the cards is not obvious with this code. Even though it's not necessary, consider using string templates, like
```js
const card = `<div class="flipper">${}</div>`
```

Third, to make your code a little cleaner you can use Array methods like `.forEach()` instead of `for` cycle.

LGeorge007: If I use StringTemplate, how can I set the draggable attribute? Ok, I'll rewrite it to StringTemplate, but I'll write the draggable right in the template.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+            isLocked = true;
+            secondCard = target.closest(".flipper");
+            secondCard.classList.toggle("rotateCard");
+            //The same card must not be removed
+            if (firstCard.id !== secondCard.id) {
+                if (firstCard.dataset.hero === secondCard.dataset.hero) {
+                    unopenedCards -= 2;
+                    setTimeout(() => {
+                        firstCard.children[1].classList.add("hiddenSide");
+                        secondCard.children[1].classList.add("hiddenSide");
```
babayK0, line 46: Don't use constructions like children[1], firstElementChild, parentNode, nextSibling, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use querySelector or closest, if in event, instead.

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+let unopenedCards;
+
+//feed him array & any even number
+const duplicateElements = function(array, times) {
+    return array.reduce((res, current) => {
+        return res.concat(Array(times).fill(current));
+    }, []);
+};
+
+const shuffle = function(arr) {
```
babayK0, line 23: you will mutate your original array so change logic of this function

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+            firstCard.classList.toggle("rotateCard");
+            isSecondCard = true;
+        };
+    };
+};
+
+const renderCard = function(arr) {
+    let htmlFragment = ``;
+    arr.forEach((el, i) => {
+        htmlFragment += `<div id=${i} class="flipper" data-hero=${el.slice(4,-4)}>
```
babayK0, line 78: don't use slice for getting data-hero. ( how you will handle this if you image format will change or one image will have .jpeg another .jpg format)

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+            };
+        } else {
+            firstCard = target.closest(".flipper");
+            firstCard.classList.toggle("rotateCard");
+            isSecondCard = true;
+        };
+    };
+};
+
+const renderCard = function(arr) {
```
babayK0, line 75:
```suggestion
const renderCards = function(arr) {
```

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+    let htmlFragment = ``;
```
babayK0, line 76: there is no need to init it with template string. `''` will enough

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+    const result = [];
+    while (arr.length > 0) {
+        let random = Math.floor(Math.random() * arr.length);
+        let elem = arr.splice(random, 1)[0];
+        result.push(elem);
+    }
+    return result;
+};
+
+const checkCard = function({ target }) {
```
babayK0, line 33: [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)

too big function. what it do ? check cards, check win condition etc.
 So it better to have smaller, more abstract functions, than complex like this one

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+    if (!target.closest(".flipper")) { return };
+    if (!isLocked) {
+        if (isSecondCard) {
+            isLocked = true;
+            secondCard = target.closest(".flipper");
+            secondCard.classList.toggle("rotateCard");
+            //The same card must not be removed
```
babayK0, line 40: redundant comment

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+    while (arr.length > 0) {
+        let random = Math.floor(Math.random() * arr.length);
+        let elem = arr.splice(random, 1)[0];
+        result.push(elem);
+    }
+    return result;
+};
+
+const checkCard = function({ target }) {
+    if (!target.closest(".flipper")) { return };
```
babayK0, line 34: curly braces is not necessary

submissions/LGeorge007/memory-pair-game/script.js
```diff
@@ -0,0 +1,93 @@
...
+    if (!isLocked) {
+        if (isSecondCard) {
```
babayK0, line 36: Avoid a lot of nesting if statements. not readable. 
Read about [guard expressions](https://medium.com/@iampika/javascript-functions-e66ed8871771) and [clauses](https://learningactors.com/javascript-guard-clauses-how-you-can-refactor-conditional-logic/)

babayK0 (changes requested):
Good job !
Watch my comments below, please !

babayK0 (approved):

PR #424, opened 2021-01-06
Friends app

f3d0t:
### Friends App

[Demo](https://f3d0t.github.io/friends-app/)
[Code base](https://github.com/f3d0t/f3d0t.github.io/tree/main/friends-app)

Please, review.

### DIFF
--- a/submissions/f3d0t/friends-app/app.js
+++ b/submissions/f3d0t/friends-app/app.js
@@ -1,0 +1,166 @@
+const MAIN = document.querySelector(".main");
+const FILTERS = document.querySelector(".filters");
+const RESET_BUTTON = document.querySelector(".filters__button");
+const DAY_NIGHT = document.querySelector(".day_night");
+const FRIENDS_COUNT = 60;
+const REQUEST_LINK = `https://randomuser.me/api/?results=${FRIENDS_COUNT}`;
+
+const fetchData = async (requestLink) => {
+	try {
+		const response = await fetch(requestLink);
+		const { results } = await response.json();
+		return results;
+	} catch (error) {
+		alert("HTTP-Error: " + error);
+	}
+};
+
+const initApp = (friendsDataArray) => {
+	const friends = new FriendsList(friendsDataArray, MAIN);
+	friends.renderCards();
+	bindEventListeners(friends);
+};
+
+const bindEventListeners = (friends) => {
+	FILTERS.addEventListener("input", ({ target }) => {
+		if (target.name == "search") {
+			friends.filterBySearch(target.value);
+		}
+		if (target.name == "gender") {
+			friends.filterByGender(target.id);
+			FILTERS.elements.search.value = "";
+			FILTERS.elements.name.forEach((radioButton) => {
+				if (radioButton.checked) {
+					friends.sortByName(radioButton.id);
+				}
+			});
+			FILTERS.elements.age.forEach((radioButton) => {
+				if (radioButton.checked) {
+					friends.sortByAge(radioButton.id);
+				}
+			});
+		}
+		if (target.name == "name") {
+			friends.sortByName(target.id);
+			FILTERS.elements.search.value = "";
+		}
+		if (target.name == "age") {
+			friends.sortByAge(target.id);
+			FILTERS.elements.search.value = "";
+		}
+	});
+	RESET_BUTTON.addEventListener("click", (e) => {
+		e.preventDefault();
+		FILTERS.querySelectorAll("input[type=radio]").forEach((radioButton) => {
+			radioButton.checked = false;
+		});
+		FILTERS.elements.search.value = "";
+		friends.resetFilters();
+	});
+	FILTERS.addEventListener("keydown", (event) => {
+		if (event.keyIdentifier == "U+000A" || event.keyIdentifier == "Enter" || event.keyCode == 13) {
+			event.preventDefault();
+		}
+	});
+};
+
+class FriendsList {
+	constructor(data, parentWrapper) {
+		this.allCards = data.map((friendData) => new FriendCard(friendData));
+		this.currentCards = [...this.allCards];
+		this.wrapper = parentWrapper;
+	}
+	renderCards(cardsArray = this.currentCards) {
+		this.wrapper.innerHTML = "";
+		this.wrapper.append(...cardsArray.map((card) => card.createCard()));
+	}
+	resetFilters() {
+		this.currentCards = [...this.allCards];
+		this.renderCards();
+	}
+	filterByGender(gender) {
+		if (gender == "male") {
+			this.currentCards = this.allCards.filter((friendCard) => friendCard["gender"] == "male");
+		}
+		if (gender == "female") {
+			this.currentCards = this.allCards.filter((friendCard) => friendCard["gender"] == "female");
+		}
+		if (gender == "all") {
+			this.currentCards = [...this.allCards];
+		}
+		this.renderCards();
+	}
+	filterBySearch(searchValue) {
+		const filteredArray = this.currentCards.filter((friendCard) => Object.values(friendCard).slice(1).join(" ").toLowerCase().indexOf(searchValue.toLowerCase()) > -1);
+		this.renderCards(filteredArray);
+	}
+	sortByName(sortType) {
+		if (sortType == "az") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "name", "ascending");
+		}
+		if (sortType == "za") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "name", "descending");
+		}
+		this.renderCards();
+	}
+	sortByAge(sortType) {
+		if (sortType == "ascending") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "ascending");
+		}
+		if (sortType == "descending") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "descending");
+		}
+		this.renderCards();
+	}
+}
+
+const sortObjectsByPropertyValue = (arrayOfObjects, key, sortOrder) => {
+	const array = arrayOfObjects;
+	const modifier = sortOrder == "ascending" ? 1 : -1;
+	return array.sort((a, b) => (a[key] > b[key] ? 1 * modifier : a[key] < b[key] ? -1 * modifier : 0));
+};
+
+class FriendCard {
+	constructor(friendData) {
+		this.gender = friendData.gender;
+		this.name = `${friendData.name.first} ${friendData.name.last}`;
+		this.age = friendData.dob.age;
+		this.photo = friendData.picture.large;
+		this.email = friendData.email;
+		this.phone = friendData.cell;
+		this.country = friendData.location.country;
+	}
+
+	createCard() {
+		const card = document.createElement("div");
+		card.classList.add("card");
+		card.innerHTML = `<img class="card__img" src="${this.photo}" alt="${this.name} photo"'>\n
+                          <h3 class="card__name">${this.name}</h3>\n
+                          <span class="card__age">Age: ${this.age}</span>\n
+                          <a class="card__phone" href="tel:+${this.phone.replace(/\D/g, "")}">${this.phone}</a>\n
+                          <a class="card__email" href="mailto:${this.email}">${this.email}</a>\n
+                          <span class="card__country">${this.country}</span>`;
+		return card;
+	}
+}
+
+const dayNightChange = () => {
+	if (!DAY_NIGHT.classList.contains("light")) {
+		document.documentElement.style.setProperty("--light", "#111");
+		document.documentElement.style.setProperty("--dark", "#eee");
+		document.documentElement.style.setProperty("--gray", "#ccc");
+		DAY_NIGHT.classList.toggle("light");
+	} else {
+		document.documentElement.style.setProperty("--light", "#eee");
+		document.documentElement.style.setProperty("--dark", "#111");
+		document.documentElement.style.setProperty("--gray", "#222");
+		DAY_NIGHT.classList.toggle("light");
+	}
+};
+
+document.addEventListener("DOMContentLoaded", () => {
+	fetchData(REQUEST_LINK).then((resultDataArray) => initApp(resultDataArray));
+	DAY_NIGHT.addEventListener("click", () => {
+		dayNightChange();
+	});
+});
--- a/submissions/f3d0t/friends-app/app.js
+++ b/submissions/f3d0t/friends-app/app.js
@@ -42,7 +42,7 @@ const bindEventListeners = (friends) => {
 			}
 			if (FILTERS.querySelector("[name=age]:checked")) {
 				friends.sortByAge(FILTERS.querySelector("[name=age]:checked").id);
-			}//this two forEach callbacks is needed to re-sort new arrays of cards, filtered by gender
+			} //this two forEach callbacks is needed to re-sort new arrays of cards, filtered by gender
 		}
 		if (target.name == "name") {
 			friends.sortByName(target.id);
@@ -172,18 +172,11 @@ class FriendCard {
 	createCard() {
 		const card = document.createElement("div");
 		card.classList.add("card");
-		card.innerHTML = `<img class="card__img" src="${this.photo}" alt="${
-			this.name
-		} photo"'>\n
+		card.innerHTML = `<img class="card__img" src="${this.photo}" alt="${this.name} photo"'>\n
                           <h3 class="card__name">${this.name}</h3>\n
                           <span class="card__age">Age: ${this.age}</span>\n
-                          <a class="card__phone" href="tel:+${this.phone.replace(
-														/\D/g,
-														""
-													)}">${this.phone}</a>\n
-                          <a class="card__email" href="mailto:${this.email}">${
-			this.email
-		}</a>\n
+                          <a class="card__phone" href="tel:+${this.phone.replace(/\D/g,"")}">${this.phone}</a>\n
+                          <a class="card__email" href="mailto:${this.email}">${this.email}</a>\n
                           <span class="card__country">${this.country}</span>`;
 		return card;
 	}
--- a/submissions/f3d0t/friends-app/app.js
+++ b/submissions/f3d0t/friends-app/app.js
@@ -1,4 +1,3 @@
-const BODY = document.body;
 const MAIN = document.querySelector(".main");
 const FILTERS = document.querySelector(".filters");
 const RESET_BUTTON = document.querySelector(".filters__button");
@@ -63,7 +62,7 @@ const bindEventListeners = (friends) => {
 		FILTERS.elements.search.value = "";
 		friends.resetFilters();
 	});
-	FILTERS.addEventListener("submit", (event) => {
+	FILTERS.addEventListener("keydown", (event) => {
 		if (event.key === "Enter") {
 			event.preventDefault();
 		}
@@ -85,14 +84,10 @@ class FriendsList {
 		this.renderCards();
 	}
 	filterByGender(gender) {
-		if (gender === "male") {
-			this.currentCards = this.allCards.filter((friendCard) => friendCard.gender === "male");
-		}
-		if (gender === "female") {
-			this.currentCards = this.allCards.filter((friendCard) => friendCard.gender === "female");
-		}
 		if (gender === "all") {
 			this.currentCards = [...this.allCards];
+		} else {
+			this.currentCards = this.allCards.filter((friendCard) => friendCard.gender === gender);
 		}
 		this.renderCards();
 	}
@@ -103,37 +98,26 @@ class FriendsList {
 		this.renderCards(filteredArray);
 	}
 	sortByName(sortType) {
-		if (sortType === "az") {
-			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "name", "ascending");
-		}
-		if (sortType === "za") {
-			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "name", "descending");
-		}
+		this.currentCards = sortObjectsByPropertyValue(this.currentCards, "name", sortType);
 		this.renderCards();
 	}
 	sortByAge(sortType) {
-		if (sortType === "ascending") {
-			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "ascending");
-		}
-		if (sortType === "descending") {
-			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "descending");
-		}
+		this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", sortType);
 		this.renderCards();
 	}
 }
 
-const sortObjectsByPropertyValue = (arrayOfObjects, key, sortOrder) => {
-	const array = arrayOfObjects;
-	array.sort((a, b) => {
+const sortObjectsByPropertyValue = (arrayOfCards, key, sortOrder) => {
+	arrayOfCards.sort((a, b) => {
 		if (a[key] > b[key]) {
 			return 1;
-		} else if (a[key] < b[key]) {
-			return -1;
-		} else {
-			return 0;
 		}
+		if (a[key] < b[key]) {
+			return -1;
+		}
+		return 0;
 	});
-	return sortOrder === "ascending" ? array : array.reverse();
+	return sortOrder === "ascending" ? arrayOfCards : arrayOfCards.reverse();
 };
 
 class FriendCard {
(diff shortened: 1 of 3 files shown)


zonzujiro:
Looks like you can sort friends by age and name "in the same" time. 🤔 

image

How it is possible? :)

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+                          <span class="card__country">${this.country}</span>`;
+		return card;
+	}
+}
+
+const dayNightChange = () => {
+	if (!DAY_NIGHT.classList.contains("light")) {
+		document.documentElement.style.setProperty("--light", "#111");
+		document.documentElement.style.setProperty("--dark", "#eee");
+		document.documentElement.style.setProperty("--gray", "#ccc");
```
zonzujiro, line 151: Why you can't do this with pure CSS?

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
+const MAIN = document.querySelector(".main");
+const FILTERS = document.querySelector(".filters");
+const RESET_BUTTON = document.querySelector(".filters__button");
+const DAY_NIGHT = document.querySelector(".day_night");
+const FRIENDS_COUNT = 60;
+const REQUEST_LINK = `https://randomuser.me/api/?results=${FRIENDS_COUNT}`;
+
+const fetchData = async (requestLink) => {
+	try {
+		const response = await fetch(requestLink);
```
zonzujiro, line 10: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+
+const bindEventListeners = (friends) => {
+	FILTERS.addEventListener("input", ({ target }) => {
+		if (target.name == "search") {
+			friends.filterBySearch(target.value);
+		}
+		if (target.name == "gender") {
+			friends.filterByGender(target.id);
+			FILTERS.elements.search.value = "";
+			FILTERS.elements.name.forEach((radioButton) => {
```
zonzujiro, line 32: You can use `querySelector` to get checked inputs

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+	RESET_BUTTON.addEventListener("click", (e) => {
+		e.preventDefault();
+		FILTERS.querySelectorAll("input[type=radio]").forEach((radioButton) => {
+			radioButton.checked = false;
+		});
+		FILTERS.elements.search.value = "";
+		friends.resetFilters();
+	});
+	FILTERS.addEventListener("keydown", (event) => {
+		if (event.keyIdentifier == "U+000A" || event.keyIdentifier == "Enter" || event.keyCode == 13) {
```
zonzujiro, line 61: And why you need so many checks? 🤔

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "descending");
+		}
+		this.renderCards();
+	}
+}
+
+const sortObjectsByPropertyValue = (arrayOfObjects, key, sortOrder) => {
+	const array = arrayOfObjects;
+	const modifier = sortOrder == "ascending" ? 1 : -1;
+	return array.sort((a, b) => (a[key] > b[key] ? 1 * modifier : a[key] < b[key] ? -1 * modifier : 0));
```
zonzujiro, line 120: Nested ternaries are hard to read. Please, don't use them

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+			FILTERS.elements.name.forEach((radioButton) => {
+				if (radioButton.checked) {
+					friends.sortByName(radioButton.id);
+				}
+			});
+			FILTERS.elements.age.forEach((radioButton) => {
+				if (radioButton.checked) {
+					friends.sortByAge(radioButton.id);
+				}
+			});//this two forEach callbacks is needed to re-sort new arrays of cards, filtered by gender
```
zonzujiro, line 41: How you can have items sorted by name and by gender in one time? 🤔

zonzujiro (changes requested):
Let's start with small steps :)

Please, check comments

submissions/f3d0t/friends-app/app.js
f3d0t: I can do It with pure CSS, but is it necessarily? To do this in pure CSS, I should declare both LightTheme and DarkTheme custom properties in :root pseudo selector, then declare --active custom properties at body{} selector with active theme styling, and then override it with changing the class of the body to "dark" with JS. It will contain really much code unlike this approach. Check option 4 in first answer in [this](https://stackoverflow.com/questions/58206867/overriding-root-css-variables-from-inner-scopes) question on stackoverflow.
It is happened because we can't override CSS custom property declared in :root pseudo-element from anywhere of other CSS. Only we can do is to override it at top parent element.

submissions/f3d0t/friends-app/app.js
f3d0t: Why not? 
Or you ask the sequence of actions?
My Gender filter really filters array - it is not sorting

submissions/f3d0t/friends-app/app.js
f3d0t: should I use pure promises syntax without await?

submissions/f3d0t/friends-app/app.js
zonzujiro: > I can do It with pure CSS, but is it necessarily?

Kind of. If something can be done with CSS - it should be done with CSS. General idea is fact that JS for logic and CSS is for styles.

> To do this in pure CSS, I should declare both LightTheme and DarkTheme custom properties in :root pseudo selector, then declare --active custom properties at body{} selector with active theme styling, and then override it with changing the class of the body to "dark" with JS. 

This is how pretty much how CSS states looks like. You declare both states in CSS and then switch them with additional class name.

>  It will contain really more code unlike current approach.

You write code for humans, not for computer. Think about this in another way:

1. It is more declarative
1. You keep everything in one place and if someone will open your file, he will see all values
1. You will keep everything related to styles in CSS
1. And pure CSS is cool! 🔥

submissions/f3d0t/friends-app/app.js
f3d0t: Fixed this)

submissions/f3d0t/friends-app/app.js
zonzujiro: No. But you should handle response status

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+		if (gender == "female") {
+			this.currentCards = this.allCards.filter((friendCard) => friendCard["gender"] == "female");
+		}
+		if (gender == "all") {
+			this.currentCards = [...this.allCards];
+		}
+		this.renderCards();
+	}
+	filterBySearch(searchValue) {
+		const filteredArray = this.currentCards.filter((friendCard) => Object.values(friendCard).slice(1).join(" ").toLowerCase().indexOf(searchValue.toLowerCase()) > -1);
```
zonzujiro, line 94: First of all, you *really* should consider using prettier - it is a hell of one liner.
And second - what is going on here? :)

submissions/f3d0t/friends-app/app.js
zonzujiro: Sry, I meant sorted by name and by age.

submissions/f3d0t/friends-app/app.js
f3d0t: Okay, I will do this =)

submissions/f3d0t/friends-app/index.html
```diff
@@ -0,0 +1,62 @@
...
+
+		<form class="filters">
+			<div class="filters__section">
+				<h3>Filters:</h3>
+			</div>
+			<div class="filters__section">
+				<input type="text" name="search" placeholder="Search by anything" class="filters__text_field" autocomplete="off" />
+			</div>
+			<hr />
+			<div class="filters__section">
```
zonzujiro, line 29: `fieldset`

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,166 @@
...
+		}
+		if (target.name == "gender") {
+			friends.filterByGender(target.id);
+			FILTERS.elements.search.value = "";
+			FILTERS.elements.name.forEach((radioButton) => {
+				if (radioButton.checked) {
+					friends.sortByName(radioButton.id);
+				}
+			});
+			FILTERS.elements.age.forEach((radioButton) => {
```
zonzujiro, line 37: If I understand correctly, you are using `NodeList#forEach` here in order to get "one" selected radio input, right?

If so, wrap them with `fieldset` and then just get checked input, because right now it is confusing.

submissions/f3d0t/friends-app/app.js
f3d0t: Some people can have the same age and in this case they will be sorted

submissions/f3d0t/friends-app/app.js
f3d0t: Already using this, but the line width was set to 999 =(( fixing this

submissions/f3d0t/friends-app/app.js
zonzujiro: And if they not? :) What use case for this?

Imagine UX:

1. I am pressing sort by age - `0 > 99` sorted
1. I am pressing sort by name - 'a > z' sorted

And age sort, basically, ruined. But your application will show, that you sorting by age and by name. What the point?

submissions/f3d0t/friends-app/app.js
f3d0t: Ok, thanks=)

submissions/f3d0t/friends-app/app.js
f3d0t: This creates new, filtered array, which is implemented to not to modify sorted array, so you can search from sorted cards.

submissions/f3d0t/friends-app/app.js
f3d0t: I will reset filtering by age after pressing name input for now.

submissions/f3d0t/friends-app/app.js
zonzujiro: I see. Please, split this logic into two steps - creation of array and then filtering them. This will improve readability

f3d0t: it gets a one FriendCard, get the array from its values by Object.values, then slices by the first (because the first is gender, and I dont want to search by gender), then it joins array to a string by " " and then find that string contains searched value by indexOf method and LowerCasing them both =)

f3d0t:
@zonzujiro Ivan, thank you for review, pls check, I have fixed those bugs

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+			this.email
+		}</a>\n
+                          <span class="card__country">${this.country}</span>`;
+		return card;
+	}
+}
+
+const dayNightChange = () => {
+	BODY.classList.toggle("light");
+	DAY_NIGHT.classList.toggle("light");
```
zonzujiro, line 187: `DAY_NIGHT` inside `body`, right? :) Why you need separate class on it?

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
+const BODY = document.querySelector("body");
```
zonzujiro, line 1: Why not address to it with `document.body`? 🤔

zonzujiro (changes requested):
Ok. Next round

submissions/f3d0t/friends-app/app.js
zonzujiro: Ok, in such a case, if someone will reorder keys in object, your slice will fail.
Please, extract `gender` using its key.
Second, would be cool to see some variable named `allValues` or `searchTarget`. Just for readability.

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+                          <h3 class="card__name">${this.name}</h3>\n
+                          <span class="card__age">Age: ${this.age}</span>\n
+                          <a class="card__phone" href="tel:+${this.phone.replace(
+														/\D/g,
+														""
+													)}">${this.phone}</a>\n
+                          <a class="card__email" href="mailto:${this.email}">${
+			this.email
+		}</a>\n
+                          <span class="card__country">${this.country}</span>`;
```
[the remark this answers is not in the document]
zonzujiro: You have changed code, but my opinion stays the same. Just in case

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+	});
+	RESET_BUTTON.addEventListener("click", (e) => {
+		e.preventDefault();
+		FILTERS.querySelectorAll("input[type=radio]:checked").forEach((radioButton) => {
+			radioButton.checked = false;
+		});
+		FILTERS.elements.search.value = "";
+		friends.resetFilters();
+	});
+	FILTERS.addEventListener("keydown", (event) => {
```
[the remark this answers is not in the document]
f3d0t: it presses the button on enter anyway, which led to reset all form)

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+
+const bindEventListeners = (friends) => {
+	FILTERS.addEventListener("input", ({ target }) => {
+		if (target.name == "search") {
+			friends.filterBySearch(target.value);
+		}
+		if (target.name == "gender") {
+			friends.filterByGender(target.id);
+			FILTERS.elements.search.value = "";
+			if (FILTERS.querySelector("[name=name]:checked")) {
```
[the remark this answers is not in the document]
f3d0t: Yes, because firstly we check if there is a checked input, and if it is, we pass this inputs Id to sorting method of the object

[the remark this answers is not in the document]
zonzujiro: If so, you can save `FILTERS.querySelector("[name=name]:checked")` into variable prior checking id :)

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+		this.wrapper.append(...cardsArray.map((card) => card.createCard()));
+	}
+	resetFilters() {
+		this.currentCards = [...this.allCards];
+		this.renderCards();
+	}
+	filterByGender(gender) {
+		if (gender == "male") {
+			this.currentCards = this.allCards.filter(
+				(friendCard) => friendCard["gender"] == "male"
```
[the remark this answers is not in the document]
f3d0t: but they are equivalent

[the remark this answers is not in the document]
zonzujiro: Not sure what you mean. Can you describe the problem in another way?

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+			);
+		}
+		if (gender == "female") {
+			this.currentCards = this.allCards.filter(
+				(friendCard) => friendCard["gender"] == "female"
+			);
+		}
```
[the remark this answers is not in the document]
f3d0t: Yes, replaced

submissions/f3d0t/friends-app/app.js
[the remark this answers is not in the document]
f3d0t: I have replaced them, but I don't understand why dot notation is preferable?

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+		this.renderCards();
+	}
+	filterBySearch(searchValue) {
+		const filteredArray = this.currentCards.filter(
+			(friendCard) =>
+				Object.values(friendCard)
+					.slice(1)
+					.join(" ")
+					.toLowerCase()
```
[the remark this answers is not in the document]
f3d0t: now I have created new searchTargetValues property like in your suggestion above

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,202 @@
...
+					.indexOf(searchValue.toLowerCase()) > -1
```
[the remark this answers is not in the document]
f3d0t: cool)

submissions/f3d0t/friends-app/app.js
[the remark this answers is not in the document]
f3d0t: > image
> 
> You don't need `\n` in multiline string. And in general, if you want reorder items into columns or row, prefer CSS. So please, use flexboxes here

You mean that inside the card I need to make a display: flex?

submissions/f3d0t/friends-app/app.js
f3d0t: just to avoid weird CSS selectors, fixed that

submissions/f3d0t/friends-app/app.js
f3d0t: oops)

f3d0t:
@zonzujiro Ivan round two done, except that flexbox question

submissions/f3d0t/friends-app/app.js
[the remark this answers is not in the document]
zonzujiro: As an example, yes.

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
+const BODY = document.body;
```
zonzujiro, line 1: Ok. Why you need constant for this? 🤔

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+	RESET_BUTTON.addEventListener("click", (e) => {
+		e.preventDefault();
+		FILTERS.querySelectorAll("input[type=radio]:checked").forEach((radioButton) => {
+			radioButton.checked = false;
+		});
+		FILTERS.elements.search.value = "";
+		friends.resetFilters();
+	});
+	FILTERS.addEventListener("submit", (event) => {
+		if (event.key === "Enter") {
```
zonzujiro, line 67: You don't need this check when you preventing submit event, am I right?

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+	filterByGender(gender) {
+		if (gender === "male") {
+			this.currentCards = this.allCards.filter((friendCard) => friendCard.gender === "male");
+		}
+		if (gender === "female") {
+			this.currentCards = this.allCards.filter((friendCard) => friendCard.gender === "female");
+		}
+		if (gender === "all") {
+			this.currentCards = [...this.allCards];
+		}
```
zonzujiro, line 96:
```suggestion
		if (gender === "all") {
			this.currentCards = [...this.allCards];
		} else {
		        this.currentCards = this.allCards.filter((friendCard) => friendCard.gender === gender );
		}
```

zonzujiro: How big value of tab spaces you have? Looks like more than 4. If so, please use one of the two most commonly used - 2 or 4.

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+	}
+	filterBySearch(searchValue) {
+		const filteredArray = this.currentCards.filter((friendCard) =>
+			friendCard.searchTarget.toLowerCase().includes(searchValue.toLowerCase())
+		);
+		this.renderCards(filteredArray);
+	}
+	sortByName(sortType) {
+		if (sortType === "az") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "name", "ascending");
```
zonzujiro, line 107: Unnamed string variables :) Remove them. Place them into variables, as an example.

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+		}
+		this.renderCards();
+	}
+	sortByAge(sortType) {
+		if (sortType === "ascending") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "ascending");
+		}
+		if (sortType === "descending") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "descending");
+		}
```
zonzujiro, line 120:
```suggestion
		this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", sortType);
```

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+		this.renderCards();
+	}
+	filterBySearch(searchValue) {
+		const filteredArray = this.currentCards.filter((friendCard) =>
+			friendCard.searchTarget.toLowerCase().includes(searchValue.toLowerCase())
+		);
+		this.renderCards(filteredArray);
+	}
+	sortByName(sortType) {
+		if (sortType === "az") {
```
zonzujiro, line 106: Unify this `sortType` too sort type of age - `"ascending"` and `"descending"`

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+		}
+		if (sortType === "descending") {
+			this.currentCards = sortObjectsByPropertyValue(this.currentCards, "age", "descending");
+		}
+		this.renderCards();
+	}
+}
+
+const sortObjectsByPropertyValue = (arrayOfObjects, key, sortOrder) => {
+	const array = arrayOfObjects;
```
zonzujiro, line 126: First: 
```suggestion
const sortObjectsByPropertyValue = (array , key, sortOrder) => {
```

And second - `array` and `arrayOfObjects` both not the best names. Use something more descriptive - what objects?

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+const dayNightChange = () => {
+	BODY.classList.toggle("light");
+};
+
+document.addEventListener("DOMContentLoaded", () => {
+	fetchData(REQUEST_LINK).then((resultDataArray) => initApp(resultDataArray));
+	DAY_NIGHT.addEventListener("click", () => {
+		dayNightChange();
+	});
+});
```
zonzujiro, line 173:
```suggestion
document.addEventListener("DOMContentLoaded", () => {
	fetchData(REQUEST_LINK).then(initApp);
	DAY_NIGHT.addEventListener("click", () => {
		BODY.classList.toggle("light");
	});
});
```

submissions/f3d0t/friends-app/app.js
```diff
@@ -0,0 +1,173 @@
...
+const sortObjectsByPropertyValue = (arrayOfObjects, key, sortOrder) => {
+	const array = arrayOfObjects;
+	array.sort((a, b) => {
+		if (a[key] > b[key]) {
+			return 1;
+		} else if (a[key] < b[key]) {
+			return -1;
+		} else {
+			return 0;
+		}
```
zonzujiro, line 134: Looks cleaner, IMO :)

```suggestion
		if (a[key] > b[key]) {
			return 1;
		} 
		
		if (a[key] < b[key]) {
			return -1;
		}
		
		return 0;
```

zonzujiro (changes requested):

submissions/f3d0t/friends-app/app.js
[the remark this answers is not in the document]
zonzujiro: Because array notation usually used for computed keys or variables, like `friendCard[gender]` or `friendCard[i * 2]`.

submissions/f3d0t/friends-app/app.js
f3d0t: removed.)

submissions/f3d0t/friends-app/app.js
f3d0t: oops, I was testing the submit event, forgot to change back.

submissions/f3d0t/friends-app/app.js
f3d0t: I have 2 tab space width in my VS code

submissions/f3d0t/friends-app/app.js
f3d0t: wooow, It calls initApp already with fetched data as a parameter!! Cool

f3d0t:
@zonzujiro Ivan, thank you for your cool review, very informative and educative! Fixed this

submissions/f3d0t/friends-app/app.js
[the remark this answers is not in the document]
f3d0t: I understand now=)

submissions/f3d0t/friends-app/app.js
zonzujiro: Do you mean that you use tabs, not spaces? Because indentation larger than 2 spaces.

zonzujiro:
Ok, LGTM

PR #437, opened 2021-01-24
Friends App

yaripey:
### Friends-app

[Demo](https://yaripey.github.io/friends-app/)
[Code base](https://github.com/yaripey/friends-app)

Please, review.

### DIFF
--- a/submissions/yaripey/friends-app/js/userService.js
+++ b/submissions/yaripey/friends-app/js/userService.js
@@ -3,37 +3,33 @@ import { settings } from './settings.js'
 let fetchingTries = 0
 const maxTriesAmount = 5;
 
-function handleErrors(response) {
-  if (!response.ok) {
-    throw Error(response.statusText);
+async function fetchUsers(amount) {
+  const usersURL = `https://randomuser.me/api/?results=${amount}`
+  try {
+    const fetchedResult = await fetch(usersURL)
+    const fetchedUsers = await fetchedResult.json()
+    console.log(fetchedUsers.results)
+    return fetchedUsers.results
+  } catch (err) {
+    console.log(err, 'Trying to fetch again.')
+    if (fetchingTries < maxTriesAmount) {
+      fetchingTries++;
+      return fetchUsers(amount)
+    } else {
+      console.log('Too many fetching attempts. Check connection or try again later.')
+      fetchingTries = 0;
+    }
   }
-  return response;
 }
 
-export function loadUsers(amount) {
-  const usersURL = `https://randomuser.me/api/?results=${amount}`
-  fetch(usersURL)
-    .then(handleErrors)
-    .then((response) => response.json())
-    .then((responseResultsObject) => {
-      const usersArray = responseResultsObject.results
-      usersArray.forEach(loadedUser => {
-        const id = loadedUsers.length
-        const newUser = new User({ ...loadedUser, id })
-        loadedUsers.push(newUser)
-      })
-      updateUsersToShow()
-    })
-    .catch(function () {
-      console.log('Error occured, trying again.')
-      if (fetchingTries < maxTriesAmount) {
-        fetchingTries++;
-        loadUsers(amount)
-      } else {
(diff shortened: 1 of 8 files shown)


zonzujiro:
I see common mistakes. Please check this list:

submissions/yaripey/friends-app/js/userService.js
```diff
@@ -0,0 +1,142 @@
...
+import { settings } from './settings.js'
+let fetchingTries = 0
+const maxTriesAmount = 5;
+
+async function fetchUsers(amount) {
+  const usersURL = `https://randomuser.me/api/?results=${amount}`
+  try {
+    const fetchedResult = await fetch(usersURL)
+    const fetchedUsers = await fetchedResult.json()
+    console.log(fetchedUsers.results)
```
zonzujiro, line 11: Please remove code that is necessary for debugging from PR :)

submissions/yaripey/friends-app/js/userService.js
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+      console.log('Too many fetching attempts. Check connection or try again later.')
+      fetchingTries = 0;
+    }
+  }
+}
+
+export async function loadUsers(amount) {
+  const fetchedUsers = await fetchUsers(amount)
+  fetchedUsers.forEach(loadedUser => {
```
zonzujiro, line 27: You will have `undefined` if all attempts will fail, no? 🤔

submissions/yaripey/friends-app/js/userService.js
```diff
@@ -0,0 +1,142 @@
...
+    const newUser = new User({...loadedUser, id})
+    loadedUsers.push(newUser)
+  })
+  updateUsersToShow()
+}
+
+export function peekUser({ target }) {
+  const targetElement = target.closest('.user')
+  if (targetElement) {
+    const id = target.closest('.user').id
```
zonzujiro, line 38:
```suggestion
    const { id } = target.closest('.user')
```

submissions/yaripey/friends-app/js/userService.js
```diff
@@ -0,0 +1,142 @@
...
+      if (!(loadedUser.age === settings.ageFilter)) {
+        return false
+      }
+    }
+    return true
+  })
+  return filteredUsers;
+}
+
+function alphaSort(usersToSort, ascending) {
```
zonzujiro, line 72: What with naming? :) Do you have `betaSort`?

submissions/yaripey/friends-app/js/userService.js
```diff
@@ -0,0 +1,142 @@
...
+    })
+  } else {
+    return usersToSort.sort(function (a, b) {
+      if (a.firstName < b.firstName) {
+        return 1;
+      }
+      if (a.firstName > b.firstName) {
+        return -1;
+      }
+      return 0;
```
zonzujiro, line 91: Same function as in `ascending` ;)

submissions/yaripey/friends-app/js/userService.js
```diff
@@ -0,0 +1,142 @@
...
+    return usersToSort.sort(function (a, b) {
+      return b.age - a.age;
+    })
+  }
+}
+
+function sortUsers(usersToSort, sortType) {
+  switch(sortType) {
+    case 'A-Z':
+      return alphaSort(usersToSort, true);
```
zonzujiro, line 111: Please, don't use `boolean` arguments. When someone else will check your code he will not understand the meaning of `true` or `false`.

Also, sort functions should understand by themself order of the sort.

zonzujiro (changes requested):

submissions/yaripey/friends-app/js/userService.js
yaripey: It's a short for alphabetical sort :D But I see the problem here, will rename.

zonzujiro (approved):

zonzujiro:
LGTM

PR #442, opened 2021-01-25
Friends-app

xT0PERx:
### Friends-app

[Demo](https://xt0perx.github.io/friends-app/)
[Code base](https://github.com/xT0PERx/friends-app)

Please, review.

### DIFF
--- a/submissions/xT0PERx/Friends app/js/index.js
+++ b/submissions/xT0PERx/Friends app/js/index.js
@@ -1,16 +1,15 @@
-const allFriends = {
-  currentList: [],
-  changeList: [],
-  numberOfFriends: 72,
+const friendsList = {
+  current: [],
+  changed: [],
+  count: 72,
 };
 
-const REQUEST_URL = "https://randomuser.me/api/1.3/";
-const ALL_USERS_CARDS_FILTERS = allFriends.changeList;
-const ALL_USERS_CARDS = allFriends.currentList;
+const ALL_USERS_CARDS_FILTERS = friendsList.changed;
+const ALL_USERS_CARDS = friendsList.current;
 const USERS_LIST = document.getElementById("usersList");
 const FILTERS_MENU = document.getElementById("filters_menu");
-const ABC_BUTTON_FILTER = document.getElementById("Abc");
-const ZYX_BUTTON_FILTER = document.getElementById("Zyx");
+const NAME_ASCEND_BUTTON_FILTER = document.getElementById("name-ascend");
+const NAME_DESCEND_BUTTON_FILTER = document.getElementById("name-descend");
 const MALE_BUTTON_FILTER = document.getElementById("male");
 const FEMALE_BUTTON_FILTER = document.getElementById("female");
 const YOUNGER_FILTER_BUTTON = document.getElementById("younger-ftr");
@@ -19,25 +18,24 @@ const RESET_BUTTON = document.getElementById("reset");
 const SEARCH_INPUT = document.querySelector(".searcher");
 
 function initApp() {
-  getUsers();
+  getUsers().then((users) => {
+    saveUsers(users);
+    createUsersCards(ALL_USERS_CARDS);
+  });
 }
 
 initApp();
 
 function getUsers() {
-  const url = `${REQUEST_URL}?results=${allFriends.numberOfFriends}`;
-  fetch(url)
+  const url = `https://randomuser.me/api/1.3/?results=${friendsList.count}`;
+
+  return fetch(url)
     .then((res) => {
       if (res.ok) {
         return res.json();
-      } else {
-        throw new Error("Что-то не так с сервером...");
       }
     })
-    .then((data) => {
-      saveUsers(data.results);
-      createUsersCards(ALL_USERS_CARDS);
-    })
+    .then((data) => data.results)
     .catch((err) => {
       console.error(err.message);
     });
@@ -68,32 +66,32 @@ function createUsersCards(users) {
 
 FILTERS_MENU.addEventListener("click", function ({ target }) {
   switch (target) {
-    case ABC_BUTTON_FILTER:
-      sortingByABC();
+    case NAME_ASCEND_BUTTON_FILTER:
+      sortByAscendName();
       break;
-    case ZYX_BUTTON_FILTER:
-      sortingByZYX();
+    case NAME_DESCEND_BUTTON_FILTER:
+      sortByDescendName();
       break;
     case YOUNGER_FILTER_BUTTON:
-      sortAscending();
+      sortAscend();
       break;
     case SENIOR_FILTER_BUTTON:
-      descendingSort();
+      descendSort();
       break;
     case MALE_BUTTON_FILTER:
-      sortingMaleOnly();
+      sortByMale();
       break;
     case FEMALE_BUTTON_FILTER:
-      sortingFemaleOnly();
+      sortByFemale();
       break;
     case RESET_BUTTON:
-      resettingFilters();
+      resetFilters();
       break;
   }
 });
 
-function sortingByABC() {
-  const abcSearch = ALL_USERS_CARDS_FILTERS.sort((a, b) => {
+function sortByAscendName() {
+  const increaseSearch = ALL_USERS_CARDS_FILTERS.sort((a, b) => {
     if (a.name.first > b.name.first) {
       return 1;
     }
@@ -104,11 +102,11 @@ function sortingByABC() {
   });
 
   resetCards();
-  createUsersCards(abcSearch);
+  createUsersCards(increaseSearch);
 }
 
-function sortingByZYX() {
-  const zyxSearch = ALL_USERS_CARDS_FILTERS.sort((a, b) => {
+function sortByDescendName() {
+  const decreaseSearch = ALL_USERS_CARDS_FILTERS.sort((a, b) => {
     if (a.name.first < b.name.first) {
       return 1;
     }
@@ -118,10 +116,10 @@ function sortingByZYX() {
     return 0;
   });
   resetCards();
-  createUsersCards(zyxSearch);
+  createUsersCards(decreaseSearch);
 }
 
-function sortAscending() {
+function sortAscend() {
   const youngers = ALL_USERS_CARDS_FILTERS.sort(
     (a, b) => a.dob.age - b.dob.age
   );
@@ -129,13 +127,13 @@ function sortAscending() {
   createUsersCards(youngers);
 }
 
-function descendingSort() {
+function descendSort() {
   const seniors = ALL_USERS_CARDS_FILTERS.sort((a, b) => b.dob.age - a.dob.age);
   resetCards();
   createUsersCards(seniors);
 }
 
-function sortingMaleOnly() {
+function sortByMale() {
   const filterMale = ALL_USERS_CARDS_FILTERS.filter(
     (elem) => elem.gender === "male"
   );
@@ -143,7 +141,7 @@ function sortingMaleOnly() {
   createUsersCards(filterMale);
 }
 
-function sortingFemaleOnly() {
+function sortByFemale() {
   const filterFemale = ALL_USERS_CARDS_FILTERS.filter(
     (elem) => elem.gender === "female"
   );
@@ -151,7 +149,7 @@ function sortingFemaleOnly() {
   createUsersCards(filterFemale);
 }
 
-function resettingFilters() {
+function resetFilters() {
   resetCards();
   createUsersCards(ALL_USERS_CARDS);
 }
@@ -164,14 +162,16 @@ function search(people) {
   resetCards();
   const search = SEARCH_INPUT.value.toLowerCase();
 
-  people = people.filter(
+  const newPeople = people.filter(
     (elem) =>
       elem.name.first.toLowerCase().match(search) ||
       elem.name.last.toLowerCase().match(search)
   );
-  createUsersCards(people);
+  createUsersCards(newPeople);
 }
 
 function resetCards() {
   USERS_LIST.innerHTML = "";
 }
+
+// function sortWithAppliedFilters() {}
--- a/submissions/xT0PERx/Friends app/js/index.js
+++ b/submissions/xT0PERx/Friends app/js/index.js
@@ -1,0 +1,177 @@
+const allFriends = {
+  currentList: [],
+  changeList: [],
+  numberOfFriends: 72,
+};
+
+const REQUEST_URL = "https://randomuser.me/api/1.3/";
+const ALL_USERS_CARDS_FILTERS = allFriends.changeList;
+const ALL_USERS_CARDS = allFriends.currentList;
+const USERS_LIST = document.getElementById("usersList");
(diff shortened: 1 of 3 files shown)


submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+    .then((res) => {
+      if (res.ok) {
+        return res.json();
+      } else {
+        throw new Error("Что-то не так с сервером...");
+      }
+    })
+    .then((data) => {
+      saveUsers(data.results);
+      createUsersCards(ALL_USERS_CARDS);
```
zonzujiro, line 39: Single responsibility - your functions should do only one job. As an example function in which you fetch users should only fetch them and not render, transform or process them in other ways. All these things should be done in another place, outside your function

zonzujiro (changes requested):
OK , I see common mistakes.  Please, go thru this check list:  

- Single responsibility - your functions should do only one job. As an example function in which you fetch users should only fetch them and not render, transform or process them in other ways. All this things should be done in another place, outside your function. Same applied to sort function, which usually do all types of sorting ;)
- Don’t use names as `arr`, `array`, `data` or etc. They to generic. Don’t use things like `sortingByABC` and `sortingByZYX` - it’s unclear and just weird. `friendsArr` also bad idea - you show data type in variable name and in JS it’s redundant
- When you need to decide which sort condition were selected, prefer `value` of input to “check class name of pressed element”, or “compare pressed element to saved element” etc. You can have value “female”, get value and then just use it

submissions/xT0PERx/Friends app/js/index.js
xT0PERx: Could you explain in more detail how I can process data without using functions in __fetch__. Perhaps I should save the data to a variable and then work with it, is this considered the best way?

zonzujiro: No, the global variable not the best way. Your get function should return the result of fetch, which is a promise with needed data and then you handle this promise as usual.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+function createUsersCards(users) {
+  const lists = document.createDocumentFragment();
+
+  users.forEach(function (el) {
+    const card = document.createElement("li");
+    card.classList.add("userCard");
+    card.innerHTML = `<span class="name">${el.name.first} ${el.name.last}</span>
+    <img src="${el.picture.large}">
+    <span class="age">Age: ${el.dob.age}</span>
+    <span class="location">Location: ${el.location.city}</span>`;
```
xT0PERx, line 62: Do I understand this point of the comment correctly? 
___Prefer template strings to “createElement” API___
This means that it is better for me to create elements not through a piece of  HTML code, but using a command such as __document.createElement()__? and then __innerHTML="bla-bla"__?

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+const ALL_USERS_CARDS = allFriends.currentList;
+const USERS_LIST = document.getElementById("usersList");
+const FILTERS_MENU = document.getElementById("filters_menu");
+const ABC_BUTTON_FILTER = document.getElementById("Abc");
+const ZYX_BUTTON_FILTER = document.getElementById("Zyx");
+const MALE_BUTTON_FILTER = document.getElementById("male");
+const FEMALE_BUTTON_FILTER = document.getElementById("female");
+const YOUNGER_FILTER_BUTTON = document.getElementById("younger-ftr");
+const SENIOR_FILTER_BUTTON = document.getElementById("senior-ftr");
+const RESET_BUTTON = document.getElementById("reset");
```
xT0PERx, line 18: Do I understand this point of the comment correctly?
___When you need to decide which sort condition were selected, prefer value of input to “check class name of pressed element”, or “compare pressed element to saved element” etc. You can have value “female”, get value and then just use it___
It is better to assign a __value__ to the __button__ element and use it in my code than to assign to the element __id__?

zonzujiro: > value of input

vs 

> value to the button

:)

And WDYM under?

>  to assign to the element id

`value` will allow you to understand what element were clicked

zonzujiro: > value of input

vs 

> value to the button

:)

And WDYM under?

>  to assign to the element id

`value` will allow you to understand what element was clicked

submissions/xT0PERx/Friends app/js/index.js
zonzujiro: No, vice-versa. Template strings should be used

submissions/xT0PERx/Friends app/js/index.js
xT0PERx: To understand which element was clicked, I use  __\<button id="name-ascend"__. My question is as follows. For elements which, according to the specification, provide such properties as __"value"__ it is better to use them, and for elements whitch dont have __"value"__  - i use __id__(if I want to apply to them JS code). 
I.e __\<button value="name-ascend"___ instead __\<button id="name-ascend"__. 
I seem to have caught up, but I still want to hear an expert's opinion :)

zonzujiro: > My question is as follows. For elements which, according to the specification, provide such properties as "value" it is better to use them

Yes.

> and for elements whitch dont have "value" - i use id(if I want to apply to them JS code).

No. `id` does not replace `value`. It is just a unique identificator and it can be used here and there. Generally speaking, it has it's own "do's and don'ts".

You don't need `button`, you need `radio` because you have two buttons and both buttons can have two states - checked and unchecked. It is a list of radio buttons.

So what you should do is replace `button` with `input type='radio'` and then get the value of input from the `change` event.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+  );
+  resetCards();
+  createUsersCards(youngers);
+}
+
+function descendSort() {
+  const seniors = ALL_USERS_CARDS_FILTERS.sort((a, b) => b.dob.age - a.dob.age);
+  resetCards();
+  createUsersCards(seniors);
+}
```
zonzujiro, line 134: Same functions. You can use one instead. Only order of arguments matters

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+  createUsersCards(filterMale);
+}
+
+function sortByFemale() {
+  const filterFemale = ALL_USERS_CARDS_FILTERS.filter(
+    (elem) => elem.gender === "female"
+  );
+  resetCards();
+  createUsersCards(filterFemale);
+}
```
zonzujiro, line 150: Same with same functions

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+      elem.name.last.toLowerCase().match(search)
+  );
+  createUsersCards(newPeople);
+}
+
+function resetCards() {
+  USERS_LIST.innerHTML = "";
+}
+
+// function sortWithAppliedFilters() {}
```
zonzujiro, line 177: Please remove unused code from PR

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+      return 1;
+    }
+    if (a.name.first > b.name.first) {
+      return -1;
+    }
+    return 0;
+  });
+  resetCards();
+  createUsersCards(decreaseSearch);
+}
```
zonzujiro, line 120: Same functions

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,177 @@
...
+      sortByMale();
+      break;
+    case FEMALE_BUTTON_FILTER:
+      sortByFemale();
+      break;
+    case RESET_BUTTON:
+      resetFilters();
+      break;
+  }
+});
```
zonzujiro, line 91: When you need to decide which sort of condition was selected, prefer value of input instead “check classname of pressed element”, or “compare pressed element to saved element” etc. You can have value female, get value and then just use it

zonzujiro (changes requested):

zonzujiro:
Please, resolve outdated :)

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,175 @@
...
+  if (WHAT_FILTER_CHOSEN === "younger") {
+    ageSort.sort(runSorting);
+  } else {
+    ageSort.sort((a, b) => runSorting(b, a));
+  }
+  resetCards();
+  createUsersCards(ageSort);
+}
+
+function sortByGender() {
```
zonzujiro, line 129:
```jsx
function sortByGender(gender) {
```

And then just

```jsx
    return ALL_USERS_CARDS_FILTERS.filter(
      (elem) => elem.gender === gender
    );
```

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,175 @@
...
+      elem.name.last.toLowerCase().match(search)
+  );
+  createUsersCards(newPeople);
+}
+
+function resetCards() {
+  USERS_LIST.innerHTML = "";
+}
+
+function offChecked() {
```
zonzujiro, line 169: `resetInputs`

zonzujiro (changes requested):

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,175 @@
...
+  });
+}
+
+function createUsersCards(users) {
+  const lists = document.createDocumentFragment();
+
+  users.forEach(function (el) {
+    const card = document.createElement("li");
+    card.classList.add("userCard");
+    template = `<span class="name">${el.name.first} ${el.name.last}</span>
```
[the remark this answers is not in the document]
xT0PERx: No, I haven't worked with Python. It's all about my carelessness - I'll fix it.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,175 @@
...
+     <img src="${el.picture.large}">
+     <span class="age">Age: ${el.dob.age}</span>
+     <span class="location">Location: ${el.location.city}</span>`;
+    card.insertAdjacentHTML("beforeEnd", template);
```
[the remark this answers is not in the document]
xT0PERx: I read in the specification that this method works faster than **innerHtml**, I decided to use it.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,175 @@
...
+    lists.append(card);
+  });
+
+  USERS_LIST.appendChild(lists);
+}
+
+FILTERS_MENU.addEventListener("click", function ({ target }) {
+  if (target.nodeName === "INPUT") {
```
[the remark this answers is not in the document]
xT0PERx: Since my __input__ with the __label__ is considered __one target__, nothing else comes to my mind than this check. I would be glad to your advice. Thank.

zonzujiro:
Resolve outdated :)

xT0PERx:
I corrected everything except the check __if (target.nodeName === "INPUT")__ - but unfortunately I don't see another way out - I would be grateful for advice on this matter. If I suddenly missed something else - I apologize - but reviewed it several times - in my opinion, I corrected all the previous comments.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+  deleteAllShowedUsers();
+  const featuredUsers = search(allSavedUsers);
+  createUsersCards(featuredUsers);
+});
+
+function search(people) {
+  const search = searchInput.value.toLowerCase();
+
+  const peopleSearch = people.filter(
```
zonzujiro, line 141: Just return, without variable.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,158 @@
...
+searchInput.addEventListener("keydown", () => {
+  if (searchInput.value === "") {
+    return;
+  }
+  deleteAllShowedUsers();
+  const featuredUsers = search(allSavedUsers);
+  createUsersCards(featuredUsers);
+});
+
+function search(people) {
```
zonzujiro, line 138: Verb

zonzujiro (changes requested):

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,137 @@
...
+
+allFilters.addEventListener("change", function ({ target }) {
+  let filteredUsers = null;
+
+  if (target.value === "reset") {
+    resetFilters();
+    return;
+  }
+
+  if (target.name === "search") {
```
zonzujiro, line 67: Can be a good idea to separate form into two fieldsets.

One with sorters and another with a search filter. Then, you will don't need this check. So this event listener will don't know anything about the existence of search input.

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,137 @@
...
+    return;
+  }
+
+  const sortedCases = {
+    sort_by_alphabet: (users) => nameSorters[target.value](users),
```
zonzujiro, line 72: I don't like `snake_cases`. In JS you should use `camelCase`. So you can rename `name` :)

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,137 @@
...
+  firstUsers.dob.age - secondUsers.dob.age;
+
+const nameSorters = {
+  nameAscend: (users) => users.sort(compareNames),
+  nameDescend: (users) => users.sort((a, b) => compareNames(b, a)),
+};
+
+const ageSorters = {
+  younger: (users) => users.sort(compareAge),
+  senior: (users) => users.sort((a, b) => compareAge(b, a)),
```
zonzujiro, line 104: `older` :)

submissions/xT0PERx/Friends app/js/index.js
```diff
@@ -0,0 +1,137 @@
...
+  deleteAllShowedUsers();
+  createUsersCards(allSavedUsers.current);
+}
+searchInput.addEventListener("keydown", () => {
+  deleteAllShowedUsers();
+  const featuredUsers = seekNames(allSavedUsers.changed);
+  createUsersCards(featuredUsers);
+});
+
+function seekNames(people) {
```
zonzujiro, line 120: `findByName`?

submissions/xT0PERx/Friends app/js/index.js
zonzujiro: And about the reset button also ;)

zonzujiro (approved):
LGTM.

Comments just FYI

PR #443, opened 2021-01-27
Friends app

ik-web:
### Friends App

[Demo](https://ik-web.github.io/friends-app/)
[Code base](https://github.com/ik-web/friends-app)

Please, review.


zonzujiro (changes requested):
I see common mistakes. Please check this list:

submissions/ik-web/friends-app/js/script.js
```diff
@@ -0,0 +1,132 @@
...
+                </div>
+            </div>
+        `;
+    }
+    
+    function getUsersInfo() {
+        return fetch(url)
+        .then(response => response.json())
+        .then(response => response.results)
+        .catch(err => window.location.reload());
```
zonzujiro, line 32: Why you need reload in case of error?

zonzujiro (changes requested):
Please, go thru the list one more time and fix all issues from it

submissions/ik-web/friends-app/js/script.js
ik-web: To re-request data from the server.  Now, it seems to me that this is bad practice ... It is better to display a message like "Something went wrong... Please reload this page" ?

zonzujiro: Nope. Don't make users solve your problems. If you believe that reloading of data will help - you should do it by yourself.

ik-web: In the current task, in my opinion, this is the correct solution, what do you think about this? can I leave it as it is?

zonzujiro: Can I hear reasons why you think this is a good solution?

submissions/ik-web/friends-app/js/script.js
```diff
@@ -0,0 +1,132 @@
...
+    // ---------------------------------------------------------------
+    function doBurgerMenu() {
+        document.querySelector('.burger').addEventListener('click', function (e) {
+            e.preventDefault();
+            this.classList.toggle('active');
+            document.querySelector('.aside').classList.toggle('active');
+            document.querySelector('.body').classList.toggle('blocked');
+        });
+    }
+    // ---------------------------------------------------------------
```
zonzujiro, line 58: Please, read a list in the comments ;)

submissions/ik-web/friends-app/js/script.js
```diff
@@ -0,0 +1,132 @@
...
+    function addUserCardsOnPage(usersInfo) {
+        let userCards = '';
+        usersInfo.forEach(user => {
+            userCards += createUserCard(
+                `${user.name.first} ${user.name.last}`,
+                user.picture.large, 
+                user.gender, 
+                user.dob.age, 
+                user.phone,
+                user.email
```
zonzujiro, line 44: Too many arguments. Usually, function should have not more than 3 arguments

ik-web: I understood. I am sorry for this mistake ... ((
I changed the parameters of this function.

submissions/ik-web/friends-app/js/script.js
ik-web: If you mean ugly comments, sorry... I didn't finish watching... I deleted them.

submissions/ik-web/friends-app/js/script.js
ik-web: In this case, I see the problem the server doesn't always return the requested data, but when the page reloads, the problem is solved. As you sayd above, the user doesn't have to solve our problems, and it would be bad to ask him to reload the page if it is possible to do an automatic reload in case of this error.

zonzujiro: Yes, but I meant that you can send a request, wait for a result, if there is an empty response or error - fetch data again.

From the user's perspective loading process will take a slightly bigger time. Even if he will mention this difference

ik-web: I can't understand how I need to write this... It should be something like a loop request, recursion?

zonzujiro: Yeap. It can be async loop or recursion.

Other students already implemented this logic, you can find it in closed PRs

zonzujiro: Yeap. It can be async loop or recursion.

Other students already implemented this logic, you can find it in closed PRs

ik-web: Thanks a lot! I did it by recursion

zonzujiro (approved):
LGTM

PR #446, opened 2021-01-31
friends-app by LGeorge007

LGeorge007:
### friends-app

[Demo](https://lgeorge007.github.io/friends-app/)
[Code base](https://github.com/LGeorge007/friends-app)

Please, review.

### DIFF
--- a/submissions/LGeorge007/friends-app/script.js
+++ b/submissions/LGeorge007/friends-app/script.js
@@ -1,0 +1,149 @@
+const content = document.querySelector(".content");
+const navPanel = document.querySelector(".sidebar");
+const button = document.querySelector(".menu-button");
+const form = document.querySelector('.filters');
+const searchInput = document.getElementById('search-input');
+const resetButton = document.getElementById('reset-button');
+const url = 'https://randomuser.me/api/?results=120&inc=gender,name,location,email,phone,picture,dob&appid=YI09-4F4Y-E9IS-OTSH';
+const state = {
+    gotResult: [],
+    filterBySex: "",
+    filterByName: "",
+    sortByAge: "",
+    sortByName: "",
+    toDefault(arr) {
+        this.filterBySex = "all";
+        this.filterByName = "",
+        this.sortByAge = "ascending";
+        this.sortByName = "none";
+        this.gotResult = arr;
+    }
+};
+let result = [];
+
+const sortByAge = (a, b) => a.dob.age <= b.dob.age ? -1 : 1;
+const sortByName = (a, b) => a.name.first.toLowerCase() >= b.name.first.toLowerCase() ? 1 : -1;
+
+const slideMenu = function() {
+    navPanel.classList.toggle("sidebar-toggle");
+    content.classList.toggle("content-toggle");
+}
+
+const controlToDefault = function() {
+    form.elements[1].checked = "true";
+    form.elements[9].checked = "true";
+    searchInput.value = "";
+}
+
+const handleInput = function () {
+    state.filterByName = this.value.trim().toLowerCase();
+    processCard(state);
+};
+
+const reset = function() {
+    state.toDefault();
+    controlToDefault();
+    processCard(state);
+};
+
+const processForm = function({currentTarget}){
+    state.sortByName = "none";
+    state.sortByAge = "none";
+    let selectedSex;
+    let elements = currentTarget.elements;
+    let selectedChoice = Array.from(elements.sortBy).filter(({checked}) => checked);
+    switch(selectedChoice[0].id) {
+        case "age-ascending":
+          state.sortByAge = "ascending";
+          break;
+        case "age-descending":
+          state.sortByAge = "descending";
+          break;
+        case "name-ascending":
+          state.sortByName = "ascending";
+          break;
+        case "name-descending":
+          state.sortByName = "descending";
+          break;
+    };
+    selectedSex = Array.from(currentTarget.elements.gender).filter(({checked}) => checked);
+    state.filterBySex = selectedSex[0].dataset.gender;
+    processCard(state);
+};
+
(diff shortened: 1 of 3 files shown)


submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+    controlToDefault();
+    processCard(state);
+};
+
+const processForm = function({currentTarget}){
+    state.sortByName = "none";
+    state.sortByAge = "none";
+    let selectedSex;
+    let elements = currentTarget.elements;
+    let selectedChoice = Array.from(elements.sortBy).filter(({checked}) => checked);
```
zonzujiro, line 67: Can be replaced with query selector

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+    switch(selectedChoice[0].id) {
```
zonzujiro, line 68: Prefer property `value` with 'ascending`, etc. values

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+const processCard = function (state){
+    if (state.filterBySex === "all"){
+        state.gotResult = result;
+    };
+    if (state.filterBySex === "male"){
+        state.gotResult = result.filter( unit => unit.gender == "male");
+    };
+    if (state.filterBySex === "female"){
+        state.gotResult = result.filter( unit => unit.gender == "female");
+    };
```
zonzujiro, line 96: `state.gotResult = result.filter( unit => unit.gender == state.filterBySex);`

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+
+    if (state.filterByName !== "") {
+        state.gotResult = state.gotResult.filter( elem => elem.name.first.toLowerCase().startsWith(state.filterByName) );
+    }
+
+    if (state.sortByAge === "ascending"){
+        state.gotResult.sort((a, b) => sortByAge(a, b));
```
zonzujiro, line 103:
```suggestion
        state.gotResult.sort(sortByAge);
```

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+};
+let result = [];
+
+const sortByAge = function(a, b) {
+    return a.dob.age == b.dob.age
+                ? 0
+                : a.dob.age <= b.dob.age
+                        ? -1
+                        : 1;
```
zonzujiro, line 29: Don't use nested ternaries. They hard to read

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+    filterByName: "",
+    sortByAge: "",
+    sortByName: "",
+    toDefault(arr) {
+        this.filterBySex = "all";
+        this.filterByName = "",
+        this.sortByAge = "ascending";
+        this.sortByName = "none";
+        this.gotResult = arr;
+    }
```
zonzujiro, line 20: Usually state should not have any functions

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+};
+let result = [];
```
zonzujiro, line 22: Why not in state? 🤔

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+}
+
+const slideMenu = function() {
+    navPanel.classList.toggle("sidebar-toggle");
+    content.classList.toggle("content-toggle");
+}
+
+const controlToDefault = function() {
+    form.elements[1].checked = "true";
+    form.elements[9].checked = "true";
```
zonzujiro, line 47: Direct access to DOM

zonzujiro (changes requested):
Leaved couple comments.

Also, I see common mistakes. Please check this list:

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,162 @@
...
+                        : 1;
+}
+
+const sortByName = function(a, b) {
+    return a.name.first.toLowerCase() == b.name.first.toLowerCase()
+                ? 0
+                : a.name.first.toLowerCase() >= b.name.first.toLowerCase()
+                        ? 1
+                        : -1;
+}
```
mustbefail, line 38: You can use .localeCompare(), the function returns exactly 0 1 -1. And you don't need .toLoweCase()
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,152 @@
...
+    state.sortedUsers = this.unsortedUsers;
+}
+
+const sortByAge = function (a, b) {
+    return a.dob.age - b.dob.age;
+}
+
+const sortByName = function (a, b) {
+    return a.name.first.localeCompare(b.name.first);
+}
```
zonzujiro, line 30: They not sorting, they comparing ;)

submissions/LGeorge007/friends-app/script.js
```diff
@@ -0,0 +1,152 @@
...
+
+const slideMenu = function () {
+    navPanel.classList.toggle("sidebar-toggle");
+    content.classList.toggle("content-toggle");
+}
+
+const reset = function () {
```
zonzujiro, line 37: Reset what? :)

zonzujiro (approved):
LGTM

PR #449, opened 2021-02-03
Friends App

zalevskyi:
### Friends App

[Demo](https://zalevskyi.github.io/kottans-frontend/practice/friends-app/)
[Code base](https://github.com/zalevskyi/kottans-frontend/tree/main/practice/friends-app)

Please, review.


submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+  fetch(`${apiUrl}?results=${USERS_QUANTITY}&inc=${fields.join(',')}`, {headers: {'Content-Type': 'application/json'}})
+    .then(response => {
+      if (!response.ok) {
+        throw Error(response.statusText)}
+      return response})
+    .then(response => response.json())
+    .then(responseJSON => {
+      if (responseJSON.hasOwnProperty('results')) {
+        pushUserData(responseJSON.results)
+        onReady()}
```
zonzujiro, line 26: No need for this `onReady`. You can return result of `fetch`, which is promise and then use `then` on it.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+       country: user.location.country,
+       city: user.location.city,
+       phone: user.phone,
+       image128px: user.picture.large}))
+}
+function showResults() {
+  RESULTS.innerHTML = getUserDataToShow(USERS_DATA, REQUEST_PARAMETERS).map(user => getUserCard(user)).join('')
+  SORT.innerHTML = [`<div class='sort_line'>Sort by: ${getSortLinks(REQUEST_PARAMETERS.sort)}</div>`,
+                  `<div class='sort_line'> Order: ${getSortOrderLinks(REQUEST_PARAMETERS.order)}</div>`].join('')
+}
```
zonzujiro, line 45: Please, use `prettier` :)

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+  let apiUrl = 'https://randomuser.me/api/1.3/'
+  let fields = ['name', 'gender', 'location', 'dob', 'phone', 'picture']
+  fetch(`${apiUrl}?results=${USERS_QUANTITY}&inc=${fields.join(',')}`, {headers: {'Content-Type': 'application/json'}})
+    .then(response => {
+      if (!response.ok) {
+        throw Error(response.statusText)}
+      return response})
+    .then(response => response.json())
+    .then(responseJSON => {
+      if (responseJSON.hasOwnProperty('results')) {
```
zonzujiro, line 24: In what case this check will be `falsy`?

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+  SORT.innerHTML = [`<div class='sort_line'>Sort by: ${getSortLinks(REQUEST_PARAMETERS.sort)}</div>`,
+                  `<div class='sort_line'> Order: ${getSortOrderLinks(REQUEST_PARAMETERS.order)}</div>`].join('')
+}
+function getUserDataToShow(userData, parameters) {
+  return sort(filter(userData, parameters), parameters)
+}
+function sort(userData, parameters) {
+  if (parameters.hasOwnProperty('sort')) {
+    let sortedData = sortAscending(userData, parameters.sort)
+    if (parameters.hasOwnProperty('order') && parameters['order']==='descending') sortedData.reverse()
```
zonzujiro, line 52: Please, use curly brackets everywhere. It is not consistent when you use them in some place and in some place don't.

Also, one time you can forget to place it and you will have tricky bug ;)

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+    return sortedData
+  }
+  else return userData
+}
+function sortAscending(userData, propertyToSort) {
+  if (propertyToSort === 'name') return sortAscendingFullName(userData, 'name', 'surname')
+  else if (propertyToSort === 'surname') return sortAscendingFullName(userData, 'surname', 'name')
```
zonzujiro, line 59: `'surname', 'name'` - don't use this approach. It is JS and you can easily missprint name of the property.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+  else if (propertyToSort === 'age') return sortAscendingNumeric(userData, 'age')
+  else return sortAscendingString(userData, propertyToSort)
+}
+function sortAscendingFullName(userData, firstToSort, secondToSort) {
+  let sortedData = userData.slice()
+  sortedData.sort((a,b) => {return `${a[firstToSort]} ${a[secondToSort]}`.localeCompare(`${b[firstToSort]} ${b[secondToSort]}`)})
```
zonzujiro, line 65: If you want one-liner, you don't need `return` here.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+}
+function getUserCard(userData) {
+  return [`<div class='card'>`,
+            `<img class='photo' alt='${userData.name} ${userData.surname} photo' src='${userData.image128px}'>`,
+            `<p class='name'>${userData.name} ${userData.surname}</p>`,
+            `<p class='demographics card_text'>${userData.gender}, ${userData.age} years</p>`,
+            `<p class='city card_text'>${userData.city}</p>`,
+            `<p class='country card_text'>${userData.country}</p>`,
+            `<p class='phone card_text'>${userData.phone}</p>`,
+          `</div>`].join('')
```
zonzujiro, line 102: Instead of array you can use template string ;)

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+function pushUserData(randomUserResults) {
+  randomUserResults.forEach(user => USERS_DATA.push(
+      {name: user.name.first,
+       surname: user.name.last,
+       gender: user.gender,
+       age: user.dob.age,
+       country: user.location.country,
+       city: user.location.city,
+       phone: user.phone,
+       image128px: user.picture.large}))
```
zonzujiro, line 39: Just `image` will be enough

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+function getSortOrderLinks(currentOrder) {
+  return ['ascending', 'descending'].map(orderType => {
+    if (orderType === currentOrder) return `<span class='sorting_selected'>${orderType}</span>`
+    else return `<a href='#order=${orderType}' class='sorting_option'>${orderType}</a>`
+  }).join(' ')
+}
+
+SORT.addEventListener('click', evt => {
+  if (evt.target.hash!==undefined) {
+    [property, value] = evt.target.hash.slice(1).split('=')
```
zonzujiro, line 119: And where declaration operator?

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+}
+function getSortOrderLinks(currentOrder) {
+  return ['ascending', 'descending'].map(orderType => {
+    if (orderType === currentOrder) return `<span class='sorting_selected'>${orderType}</span>`
+    else return `<a href='#order=${orderType}' class='sorting_option'>${orderType}</a>`
+  }).join(' ')
+}
+
+SORT.addEventListener('click', evt => {
+  if (evt.target.hash!==undefined) {
```
zonzujiro, line 118: Why you need this check?

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+  else if (propertyToSort === 'surname') return sortAscendingFullName(userData, 'surname', 'name')
+  else if (propertyToSort === 'age') return sortAscendingNumeric(userData, 'age')
+  else return sortAscendingString(userData, propertyToSort)
+}
+function sortAscendingFullName(userData, firstToSort, secondToSort) {
+  let sortedData = userData.slice()
+  sortedData.sort((a,b) => {return `${a[firstToSort]} ${a[secondToSort]}`.localeCompare(`${b[firstToSort]} ${b[secondToSort]}`)})
+  return sortedData
+}
+function sortAscendingNumeric(userData, propertyToSort) {
```
zonzujiro, line 68: Looks like you don't have functions for descending sort.

Please, test your application :)

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+    REQUEST_PARAMETERS[property] = value
+    showResults()
+  }
+  evt.preventDefault()
+})
+FILTER.addEventListener('input', evt => {
+  ['name', 'surname', 'age_from', 'age_till', 'city', 'country'].forEach(key => {
+    FILTER[key].value.length>0 ? REQUEST_PARAMETERS[key] = FILTER[key].value.toLowerCase() : delete REQUEST_PARAMETERS[key]
+  })
+  FILTER.gender.value!=='all' ? REQUEST_PARAMETERS.gender = FILTER.gender.value : delete REQUEST_PARAMETERS.gender
```
zonzujiro, line 129: Don't use ternaries in such a way :) Replace it with `if ... else`

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+function showResults() {
+  RESULTS.innerHTML = getUserDataToShow(USERS_DATA, REQUEST_PARAMETERS).map(user => getUserCard(user)).join('')
+  SORT.innerHTML = [`<div class='sort_line'>Sort by: ${getSortLinks(REQUEST_PARAMETERS.sort)}</div>`,
+                  `<div class='sort_line'> Order: ${getSortOrderLinks(REQUEST_PARAMETERS.order)}</div>`].join('')
+}
+function getUserDataToShow(userData, parameters) {
+  return sort(filter(userData, parameters), parameters)
+}
+function sort(userData, parameters) {
+  if (parameters.hasOwnProperty('sort')) {
```
zonzujiro, line 50: `hasOwnProperty` is redundany here. You can just use `parameters.sort`

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+            `<p class='demographics card_text'>${userData.gender}, ${userData.age} years</p>`,
+            `<p class='city card_text'>${userData.city}</p>`,
+            `<p class='country card_text'>${userData.country}</p>`,
+            `<p class='phone card_text'>${userData.phone}</p>`,
+          `</div>`].join('')
+}
+function getSortLinks(currentSort) {
+  return ['name', 'surname', 'age', 'gender', 'city', 'country'].map(property => {
+    if (property === currentSort) return `<span class='sorting_selected'>${property}</span>`
+    else return `<a href='#sort=${property}' class='sorting_option'>${property}</a>`
```
zonzujiro, line 107: Why it should be link?

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,133 @@
...
+function filter(userData, parameters) {
+  Object.keys(parameters).forEach(key => {
+    if (['name', 'surname', 'country', 'city'].includes(key)) {
+      userData = userData.filter(user => user[key].toLowerCase().startsWith(parameters[key]))}
+    else if (key==='gender') {
+      userData = userData.filter(user => user.gender.toLowerCase()===parameters.gender)}
+    else if (key==='age_from') {
+      userData = userData.filter(user => user.age >= parameters.age_from)}
+    else if (key==='age_till') {
+      userData = userData.filter(user => user.age <= parameters.age_till)}
```
zonzujiro, line 90: `userData.filter(user => user.age <= parameters[key])`

zonzujiro (changes requested):
I gave a short look and looks like you don't have the functionality of `descending` sort. Please, implement it and test your app prior to sending it to review.

Steps to reproduce:
- filter all users into some age boundries
- try to change sort order

Also, couple of small things:
- Use `prettier` and curly brackets - it will make your code easier to review

submissions/zalevskyi/friends-app/app.js
zalevskyi: redone with select (for property to sort) and radio (for order)

submissions/zalevskyi/friends-app/app.js
zalevskyi: In case if API will be changed suddenly and JSON will have different structure
Some kind of distrust to API
Removed

submissions/zalevskyi/friends-app/app.js
zalevskyi: overcautious
removed

zalevskyi:
Thank you for the review!
Reformatted with prettier and [tested](https://share.clickup.com/clip/p/t4624973/f8315ef2-4744-43fb-bb3c-705ea1358958/screen-recording-2021-02-12-23:23.webm)
Please review once more.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
...
+  RESULTS.innerHTML = `<div class='connection_message'>Please wait. Connecting to server. Attemp: ${connectionAttemps}</div>`
+  connectionAttemps++
+  fetchUserData(connectToServer)
+}
+function fetchUserData(onError) {
+  let apiUrl = 'https://randomuser.me/api/1.3/'
+  let fields = ['name', 'gender', 'location', 'dob', 'phone', 'picture']
+  fetch(`${apiUrl}?results=${USERS_QUANTITY}&inc=${fields.join(',')}`, {
+    headers: { 'Content-Type': 'application/json' },
+  })
```
zonzujiro, line 19: Just return `fetch` and then you will have a possibility to write `fetchUserData().catch()`

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
...
+    orderValue = -1
+  }
+  sortedData.sort((a, b) => {
+    if (a[propertyToSort] > b[propertyToSort]) {
+      return 1 * orderValue
+    } else if (a[propertyToSort] < b[propertyToSort]) {
+      return -1 * orderValue
+    } else {
+      return 0
+    }
```
zonzujiro, line 112: Looks like overengineering. You can have only `sortByString` function and then pass arguments in the different order.

Please, check merged PRs

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
...
+    (a, b) =>
+      `${a[firstToSort]} ${a[secondToSort]}`.localeCompare(
+        `${b[firstToSort]} ${b[secondToSort]}`
+      ) * orderValue
+  )
+  return sortedData
+}
+function sortNumeric(userData, propertyToSort, order) {
+  let sortedData = userData.slice()
+  if (order === 'ascending') {
```
zonzujiro, line 91: Same here - no need in this check in this function

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
...
+function getUserDataToShow(userData, parameters) {
+  if (parameters.sort) {
+    return sort(filter(userData, parameters), parameters.sort, parameters.order)
+  } else {
+    return filter(userData, parameters)
+  }
+}
+function sort(userData, fieldToSort, order) {
+  if (fieldToSort === 'name') {
+    return sortFullName(userData, 'name', order)
```
zonzujiro, line 61: Passing the string key is a bad idea - you can misprint and you will have a bug. Instead, your sorting function should accept concrete values from two objects - friend A and friend B.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
+const USERS_DATA = []
+const USERS_QUANTITY = 30
+const RESULTS = document.getElementById('results')
+const SORT = document.getElementById('sort')
+const FILTER = document.getElementById('filter')
+const REQUEST_PARAMETERS = {}
+let connectionAttemps = 1
```
zonzujiro, line 7: Why it is global? Is it used anywhere except the fetching function?

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
...
+
+function connectToServer() {
+  RESULTS.innerHTML = `<div class='connection_message'>Please wait. Connecting to server. Attemp: ${connectionAttemps}</div>`
+  connectionAttemps++
```
zonzujiro, line 11: Ok. This logic should be inside fetching function

zonzujiro (changes requested):

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
+const USERS_DATA = []
+const USERS_QUANTITY = 30
```
zonzujiro, line 2: Why it is global also?

submissions/zalevskyi/friends-app/app.js
zalevskyi: Thank you for the suggestion. It helped me to think over it once again and improve the code.
But I disagree with your request in several points:
1. Bad naming request. 

> You can have only `sortByString` function and then pass arguments in the different order

I have checked the [last merged PR](https://github.com/kottans/frontend-2021-homeworks/pull/455/files) and found this sample `sortedArr.sort((a, b) => sortByName(a, b))`
Well, in English it literally means: "I want to sort array by **_sorting_** age" while it should be "I want to sort array by **_comparing_** age". You can not sort by sorting, you can sort by comparing different values
Due to this naming (`sortByString`) It was very hard for me to understand what does it mean:

> pass arguments in the different order

Sorting by its meaning should not depend on argument order, as order is output of sorting, but not input. While comparing does depends on order.

2. No clear reason for such comparison function as it is already built-in into language

Again in the [same merged PR](https://github.com/kottans/frontend-2021-homeworks/pull/455/files) I found such comparison function
```
const sortByName = (a, b) => {
  return a.name.first.localeCompare(b.name.first);
}
```
But it adds nothing to built-in function, just:
-  renames it (and does it in bad way by naming sort what is actually compare)
-  changes the way String instance method call to static-like call

3.  Error prone

As you recommended for other code snippet:
`sortAscendingFullName(userData, 'name', 'surname')`
> 'surname', 'name' - don't use this approach. It is JS and you can easily misprint name of the property.

As for me it is even more easily to misprint just one letter or order of letters in multiple places
Just compare my new [implementation with flipping sign of compare result](https://gist.github.com/zalevskyi/6dfea994f6e5ea8401d49cf7984f0d46) vs. the same [function but changing arguments order in compare functions as suggested](https://gist.github.com/zalevskyi/efd923f3fc0cbd28dda26c921a18b936)
I don't think that it is overengineering. There is no space for error of misprinting order of `a` or `b`. There is even no need to think about it and check it. Instead it is clearly stated: flip compare result if needed by `orderValue`.

submissions/zalevskyi/friends-app/app.js
zalevskyi: `const USERS_QUNTITY` was global to understand easily this parameter of implementation in future.
It was easy to find answer to possible question "Why there is only 30 users shown?"
Now I have moved it into `function connectToServer()`. Or I can move it even lower to `function fetchUserData()`. And in this way it becomes more like hard coded magic number

zalevskyi:
Thank you for the review!
Resolved is corrected.
But I disagree with some sort recommendations.
Please review once more.

submissions/zalevskyi/friends-app/app.js
zonzujiro: Sorry for being confusing :)

The main idea was next:

You don't need things like `sortAscendingFullName` and `sortDescendingFullName`, because you can have only one function, which will sort by name. For this, you can concatenate the name and full name into one value somewhere.

Then, you don't need `sortAscending` and `sortDescending` functions.

> ⚠️ Please, consider next more like pseudocode and not like the source of copy-paste. You should find a better solution on your own.

> Regarding naming - feel free to use `compareNames` and `compareSurnames` if you like.

It is much simpler code when you will write something like this:

```jsx
const compareAge = (firstFriend, secondFriend) => {
  return firstFriend.dob.age - secondFriend.dob.age;
};
```

Then, also as an example, you can have:

```jsx
const ageSorters = {
  descending: () => {
    friends.sort((a, b) => compareAge(a, b));
  },
  ascending: () => {
    friends.sort(compareAge);
  }
};
```

And use it like this:

```jsx
myForm.addEventListener("change", ({ target: radioButton }) => {
  if (radioButton.name === AGE) {
    ageSorters[radioButton.value]();
  }

  if (radioButton.name === NAME) {
    nameSorters[radioButton.value]();
  }
});
```

Which, possibly, can be simplified to:

```jsx
myForm.addEventListener("change", ({ target: radioButton }) => {
  const sorter =
    radioButton.name === AGE
      ? ageSorters[radioButton.value]
      : nameSorters[radioButton.value];

  sorter();
});
```

And that's all :)

1. You have separated the logic of sorting into two separate domains - sorting by age and sorting by name and moved everything into separately mapped handlers. So you don't need to write endless `if ... else` checks. Instead, you are using declarative code.
1. Because of this you don't have a place for "misprinting"
1. You don't pass any strings as arguments anywhere, which is antipattern in general. And with object keys in particular.

Now compare it with code from your examples.

> Actually it is rather under engineering. Potentially in case of very frequent comparison in extremely data savvy application flipping sing of comparison (multiplying by 1 or -1) will negatively impact performance to some extent and cost money.

If I got a penny every time... just read about premature optimization. As an example [here](https://wiki.c2.com/?PrematureOptimization)

TLDR:

The following set of heuristics is a correction:
1. Make it work.
1. Make it right - the code is readable and every idea is expressed.
1. Make everything work.
1. Make everything right.
1. Use the system and find performance bottlenecks.
1. Use a profiler in those bottlenecks to determine what needs to be optimized. See ProfileBeforeOptimizing.
1. Make it fast. You maintained unit tests, right? Then you can refactor the code mercilessly in order to improve the performance.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,187 @@
...
+function connectToServer() {
+  RESULTS.innerHTML = `<div class='connection_message'>Please wait. Connecting to server. Attemp: ${connectionAttemps}</div>`
+  connectionAttemps++
+  fetchUserData(connectToServer)
+}
+function fetchUserData(onError) {
+  let apiUrl = 'https://randomuser.me/api/1.3/'
+  let fields = ['name', 'gender', 'location', 'dob', 'phone', 'picture']
+  fetch(`${apiUrl}?results=${USERS_QUANTITY}&inc=${fields.join(',')}`, {
+    headers: { 'Content-Type': 'application/json' },
```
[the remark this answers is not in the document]
zalevskyi: Well, I guess I copied it from POST request example of fetch()
Removed

submissions/zalevskyi/friends-app/app.js
zalevskyi: Thank you very much! It was really helpful to understand.
Effectiveness of simply changing order of function arguments, possibility to use functions as object properties and antipattern of strings as arguments was not obvious to me.
Thank you for the article about optimization. That is going to be next step for me.

zalevskyi:
Thank you for all your comments and time!
Please review once more.

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,109 @@
...
+    country: (a, b) => a.country.localeCompare(b.country),
+    city: (a, b) => a.city.localeCompare(b.city),
+  }
+}
+function connectToServer() {
+  let connectionAttemps = 1
+  fetchUserData(connectionAttemps)
+    .then((responseJSON) => {
+      responseJSON.results.forEach((user) => USERS_DATA.push(new User(user)))
+      users_data_filtered = [...USERS_DATA]
```
zonzujiro, line 33: `camelCase` please. Always ;)

submissions/zalevskyi/friends-app/app.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+SORT.addEventListener('input', (evt) => {
+  if (SORT.field.value) {
+    showResults(sorter[SORT.order.value](users_data_filtered, SORT.field.value))
+  }
+})
+FILTER.addEventListener('input', ({ target: { name, value } }) => {
+  if (value) FILTER_PARAMETERS[name] = value.toLowerCase()
+  else delete FILTER_PARAMETERS[name]
```
zonzujiro, line 104: Brackets

zonzujiro (approved):
LGTM.

Comments just FYI

PR #450, opened 2021-02-03
Friends App

Harnytskyi:
### Friends App

[Demo](https://harnytskyi.github.io/kottans-frontend/task_friends_app/)
[Code base](https://github.com/Harnytskyi/kottans-frontend/tree/main/task_friends_app)

Please, review.

### DIFF
--- a/submissions/Harnytskyi/friends-app/index.html
+++ b/submissions/Harnytskyi/friends-app/index.html
@@ -38,9 +38,9 @@
             </div>  
             <div id="filter-gender" class="filter-item">
                 <p class="filters-header"><b>Gender</b></p>
-                <label><input type="checkbox" id="gender-male" value="male" checked>Man</label>
+                <label><input type="checkbox" id="gender-male" class="gender-checkbox" value="male" checked>Man</label>
                 <br>
-                <label><input type="checkbox" id="gender-female" value="female" checked>Woman</label>
+                <label><input type="checkbox" id="gender-female" class="gender-checkbox" value="female" checked>Woman</label>
             </div>
             <div id="filter-age" class="filter-item">
                 <p class="filters-header"><b>Age</b></p>
--- a/submissions/Harnytskyi/friends-app/index.html
+++ b/submissions/Harnytskyi/friends-app/index.html
@@ -12,23 +12,35 @@
 <body>
     <header class="header">
             <div class="header-title"><b>Friends</b>App</div>
-        <input id="search" class="search" onkeyup="search()" placeholder="Search" onsearch="search()" type="search">
+        <input id="search" class="search" onkeyup="searchEvent()" placeholder="Search" onsearch="searchEvent()" type="search">
     </header>
     <main>
         <div class="filters">
-            <div class="filter-item">
+            <div id="filter-sort" class="filter-item">
                 <p class="filters-header"><b>Sort</b></p>
-                <input type="button" class="button-sort" id="a-z" value="A-Z">
-                <input type="button" class="button-sort" id="z-a" value="Z-A">
+                <label class="radio-sort-container" value="">
+                    <input type="radio" name="sort-alphabet" id="a-z" value="alphabet-ascend">
+                    <div class="radio-sort">A-Z</div>
+                </label>
+                <label class="radio-sort-container">
+                    <input type="radio" name="sort-alphabet" id="z-a" value="alphabet-descend">
+                    <div class="radio-sort">Z-A</div>
+                </label>
                 <br>
-                <input type="button" class="button-sort" id="0-9" value="0-9">
-                <input type="button" class="button-sort" id="9-0" value="9-0">
+                <label class="radio-sort-container">
+                    <input type="radio" name="sort-age" id="0-9" value="age-ascend">
+                    <div class="radio-sort">0-9</div>
+                </label>
+                <label class="radio-sort-container">
+                    <input type="radio" name="sort-age" id="9-0" value="age-descend">
+                    <div class="radio-sort">9-0</div>
+                </label>
             </div>  
-            <div class="filter-item">
+            <div id="filter-gender" class="filter-item">
                 <p class="filters-header"><b>Gender</b></p>
-                <label><input type="checkbox" id="gender-male"  checked>Man</label>
+                <label><input type="checkbox" id="gender-male" value="male" checked>Man</label>
                 <br>
-                <label><input type="checkbox" id="gender-female" checked>Woman</label>
+                <label><input type="checkbox" id="gender-female" value="female" checked>Woman</label>
             </div>
         </div>
         <div id = "contentField" class="content-field">
--- a/submissions/Harnytskyi/friends-app/index.html
+++ b/submissions/Harnytskyi/friends-app/index.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <title>Friends App</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=DM+Sans:ital,wght@0,500;1,400&display=swap" rel="stylesheet">
+    <link rel="stylesheet" type="text/css" media="screen" href="css/style.css" />
+</head>
+<body>
+    <header class="header">
+            <div class="header-title"><b>Friends</b>App</div>
+        <input id="search" class="search" onkeyup="search()" placeholder="Search" onsearch="search()" type="search">
+    </header>
+    <main>
+        <div class="filters">
+            <div class="filter-item">
+                <p class="filters-header"><b>Sort</b></p>
+                <input type="button" class="button-sort" id="a-z" value="A-Z">
+                <input type="button" class="button-sort" id="z-a" value="Z-A">
+                <br>
+                <input type="button" class="button-sort" id="0-9" value="0-9">
+                <input type="button" class="button-sort" id="9-0" value="9-0">
+            </div>  
+            <div class="filter-item">
+                <p class="filters-header"><b>Gender</b></p>
+                <label><input type="checkbox" id="gender-male"  checked>Man</label>
+                <br>
+                <label><input type="checkbox" id="gender-female" checked>Woman</label>
+            </div>
+        </div>
+        <div id = "contentField" class="content-field">
+        <div id = "allFriends" class="allFriends"></div>
+    </div>
+    </main>
+    <script src="js/script.js"></script>
+</body>
+</html>
--- a/submissions/Harnytskyi/friends-app/index.html
+++ b/submissions/Harnytskyi/friends-app/index.html
@@ -12,27 +12,27 @@
 <body>
     <header class="header">
             <div class="header-title"><b>Friends</b>App</div>
-        <input id="search" class="search" onkeyup="searchEvent()" placeholder="Search" onsearch="searchEvent()" type="search">
+        <input id="search" class="search" onkeyup="selectFriends()" placeholder="Search" onsearch="selectFriends()" type="search">
     </header>
     <main>
         <div class="filters">
             <div id="filter-sort" class="filter-item">
                 <p class="filters-header"><b>Sort</b></p>
                 <label class="radio-sort-container" value="">
-                    <input type="radio" name="sort-alphabet" id="a-z" value="alphabet-ascend">
+                    <input type="radio" name="sort" id="a-z" value="alphabet-ascend">
                     <div class="radio-sort">A-Z</div>
                 </label>
                 <label class="radio-sort-container">
-                    <input type="radio" name="sort-alphabet" id="z-a" value="alphabet-descend">
+                    <input type="radio" name="sort" id="z-a" value="alphabet-descend">
                     <div class="radio-sort">Z-A</div>
                 </label>
                 <br>
                 <label class="radio-sort-container">
-                    <input type="radio" name="sort-age" id="0-9" value="age-ascend">
+                    <input type="radio" name="sort" id="0-9" value="age-ascend">
                     <div class="radio-sort">0-9</div>
                 </label>
                 <label class="radio-sort-container">
-                    <input type="radio" name="sort-age" id="9-0" value="age-descend">
+                    <input type="radio" name="sort" id="9-0" value="age-descend">
                     <div class="radio-sort">9-0</div>
                 </label>
             </div>  
@@ -42,10 +42,16 @@
                 <br>
                 <label><input type="checkbox" id="gender-female" value="female" checked>Woman</label>
             </div>
+            <div id="filter-age" class="filter-item">
+                <p class="filters-header"><b>Age</b></p>
+                <input type="number" id="minAge" min="0" title="Min age" class="number-age">
+                <input type="number" id="maxAge" min="0" title="Max age" class="number-age">
+            </div>
         </div>
         <div id = "contentField" class="content-field">
+        <div id = "errorMessage" class="error-text"></div>
         <div id = "allFriends" class="allFriends"></div>
-    </div>
+        </div>
     </main>
     <script src="js/script.js"></script>
 </body>
--- a/submissions/Harnytskyi/friends-app/js/script.js
+++ b/submissions/Harnytskyi/friends-app/js/script.js
@@ -8,12 +8,8 @@ const MAX_AGE = document.getElementById("maxAge");
 const FILTER_AGE = document.getElementById("filter-age");
 let friends = [];
 let minAge, maxAge;
-let gender = {
-    male: true,
-    female: true
-}
-
 let selectedFriends;
+
 function saveUsers(usersArray) {
     friends = usersArray.map(user => {
         return {
@@ -27,14 +23,9 @@ function saveUsers(usersArray) {
     })
 }
 function filterByGender() {
-    if (gender.male == true && gender.female == true)
-        selectedFriends = friends;
-    else if (gender.male == true && gender.female == false)
-        selectedFriends = friends.filter(item => item.gender == "male")
-    else if (gender.male == false && gender.female == true)
-        selectedFriends = friends.filter(item => item.gender == "female")
-    else
-        selectedFriends = [];
+    let checkboxes = Array.from(document.querySelectorAll(".gender-checkbox:checked"));
+    selectedFriends = friends.filter(item => 
+        checkboxes.some((gender) => item.gender === gender.value));
 }
 function filterByAge() {
     selectedFriends = selectedFriends.filter(friend => (friend.age >= minAge && friend.age <= maxAge));
@@ -55,9 +46,6 @@ function displayFriends() {
     }).join("");
 }
 
-function changeFilterStatus(checkbox) {
-    gender[checkbox.value] = !gender[checkbox.value];
-}
 function selectFriends() {
     filterByGender();
     filterByAge();
@@ -130,7 +118,6 @@ function initApp() {
         });
 }
 
-
 function showNotFoundMessage() {
     ERROR_MESSAGE.innerHTML = `Not Found`;
 }
@@ -152,7 +139,6 @@ function checkAgeRangeValidity() {
 
 FILTER_GENDER.addEventListener('change', (event) => {
     const target = event.target;
-    changeFilterStatus(target);
     selectFriends();
 })
 FILTER_SORT.addEventListener('change', (event) => {
--- a/submissions/Harnytskyi/friends-app/js/script.js
+++ b/submissions/Harnytskyi/friends-app/js/script.js
@@ -1,0 +1,151 @@
+const GRID = document.getElementById("allFriends");
+const CHECKBOX_MAN = document.getElementById("gender-male");
+const CHECKBOX_WOMAN = document.getElementById("gender-female");
+const SORT_AZ = document.getElementById("a-z");
+const SORT_ZA = document.getElementById("z-a");
+const SEARCH_INPUT = document.getElementById('search');
+const SORT_09 = document.getElementById("0-9");
+const SORT_90 = document.getElementById("9-0");
+let friends = [];
+
+let selectedFriends = [...friends];
+function filter() {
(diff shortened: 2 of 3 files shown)


submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
+const GRID = document.getElementById("allFriends");
+const CHECKBOX_MAN = document.getElementById("gender-male");
+const CHECKBOX_WOMAN = document.getElementById("gender-female");
+const SORT_AZ = document.getElementById("a-z");
+const SORT_ZA = document.getElementById("z-a");
+const SEARCH_INPUT = document.getElementById('search');
+const SORT_09 = document.getElementById("0-9");
+const SORT_90 = document.getElementById("9-0");
```
zonzujiro, line 8: I also can suggest `SORT_60` because of the speed limit and `SORT_88` just because the cool number.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+let friends = [];
+
+let selectedFriends = [...friends];
+function filter() {
```
zonzujiro, line 12: `filterByGender`

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+            <div class="email">
+                <a href="${item.email}">
+                    <button class="button-message">SEND MESSAGE</button>
+                </a>
+            </div>
+        <div>`
+    });
+    search();
+}
+let gender = {
```
zonzujiro, line 37: `let` will not hoist and you have defined it's *after* you have used it.

Bad practice.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+function displayFriends() {
+    GRID.innerHTML = '';
+    selectedFriends.forEach(item => {
+        GRID.innerHTML +=  `<div id="${item.name}" class="card">
+            <img class="photo" src="${item.photo}">
+            <p class="name">${item.name}</p>
+            <p class="age">Age: ${item.age}</p>
+            <a class="phone" href="${item.phone}">${item.phone}</a>
+            <div class="email">
+                <a href="${item.email}">
```
zonzujiro, line 29: Broken link. `href` argument for an email have special format

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+                CHECKBOX_WOMAN.click();
+            }
+        }
+        else if (sex == "female") {
+            if (gender.male == false) {
+                gender.male = true;
+                CHECKBOX_MAN.click();
+            }
+        }
+        gender[sex] = false;
```
zonzujiro, line 58: You can simplify it.

1. Get the value of checkbox and then check the status of gender with `gender[checkox.value]`
2. Change state in`gender`

submissions/Harnytskyi/friends-app/index.html
```diff
@@ -0,0 +1,40 @@
...
+    </header>
+    <main>
+        <div class="filters">
+            <div class="filter-item">
+                <p class="filters-header"><b>Sort</b></p>
+                <input type="button" class="button-sort" id="a-z" value="A-Z">
+                <input type="button" class="button-sort" id="z-a" value="Z-A">
+                <br>
+                <input type="button" class="button-sort" id="0-9" value="0-9">
+                <input type="button" class="button-sort" id="9-0" value="9-0">
```
zonzujiro, line 25: Just make it as list of radio buttons. Then HTML will hold a state of current selected input and you will have a possibility just to get `change` event with `value` of checked element.

submissions/Harnytskyi/friends-app/index.html
```diff
@@ -0,0 +1,40 @@
...
+            </div>  
+            <div class="filter-item">
+                <p class="filters-header"><b>Gender</b></p>
+                <label><input type="checkbox" id="gender-male"  checked>Man</label>
```
zonzujiro, line 29: Redundant spaces before `checked`

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+    changeFilterStatus(CHECKBOX_MAN, "male");
+});
+CHECKBOX_WOMAN.addEventListener('change', function () {
+    changeFilterStatus(CHECKBOX_WOMAN, "female");
+});
+SORT_AZ.addEventListener('click', function () {
+    sortFriends("name");
+})
+SORT_ZA.addEventListener('click', function () {
+    sortFriends("name", "descending");
```
zonzujiro, line 100: Don't send name of properties to objects as strings. In JS you can easily missprint and get tricky error.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+    filter();
+    displayFriends();
+}
+
+CHECKBOX_MAN.addEventListener('change', function () {
+    changeFilterStatus(CHECKBOX_MAN, "male");
+});
+CHECKBOX_WOMAN.addEventListener('change', function () {
+    changeFilterStatus(CHECKBOX_WOMAN, "female");
+});
```
zonzujiro, line 95: Event delegation

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+})
+SORT_ZA.addEventListener('click', function () {
+    sortFriends("name", "descending");
+})
+SORT_09.addEventListener('click', function () {
+    sortFriends("age");
+})
+
+SORT_90.addEventListener('click', function () {
+    sortFriends("age", "descending");
```
zonzujiro, line 107: Event delegation

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+})
+
+function fetchUsers(){
+    let targetUrl  = "https://randomuser.me/api/?results=30&nat=au,ca,ch,de,dk,es,fr,gb,ie,no,nl,nz,us"
+
+    fetch(targetUrl, {metod: "get"})
+    .then((response) => {
+      let json = response.json();
+      if (response.status >= 200 && response.status < 300) {
```
zonzujiro, line 116: just `response.ok`

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+    })
+    .catch(function(error) {
+      showErrorMessage(error);
+    });
+}
+
+function saveUsers(usersArray){
+    usersArray.forEach(user => {
+        let friend ={
+            name: user.name.first +" "+ user.name.last,
```
zonzujiro, line 133: You can use template strings

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+    filter();
+    displayFriends();
+}
+
+function search() {
+    let searchRequest = SEARCH_INPUT.value.toUpperCase();
+    selectedFriends.forEach(item => {
+        const FRIEND = document.getElementById(item.name)
+        FRIEND.classList.remove('hidden');
+        if (item.name.toUpperCase().indexOf(searchRequest) == -1) {
```
zonzujiro, line 69: `String#includes`

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,151 @@
...
+            FRIEND.classList.add('hidden');
```
zonzujiro, line 70: It's a bad idea.

You will search in DOM every item every time when you need to hide it. Also, hiding items one-by-one will hit performance, because browser will do reflow/repaint operations every time when you appply `hidden`.

It is easier to complitely rerender necessary items.

zonzujiro (changes requested):

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: What does "will not hoist" mean?
Do I just need to move this object higher?

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: What is wrong? Can't use Event delegation?

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: How should be done in this case?
Declare global variables, assign property names to them, and send them?

zonzujiro:
Please, resolve outdated

submissions/Harnytskyi/friends-app/js/script.js
zonzujiro: https://developer.mozilla.org/en-US/docs/Glossary/Hoisting

`const` and `let` usually will not be visible prior to their declaration. In your case, your function will work fine, because it will get the actual value from the variable during runtime.

So it will work anyway. But this is a good general rule - declaring variables prior they were used :)

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: renamed `SORT_90` to `SORT_AGE_DESCEND`

submissions/Harnytskyi/friends-app/index.html
Harnytskyi: replaced these `button` with `radio`

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: Redesigned the `sortFriends` function, the strings are no longer send

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: united two event listeners into a common one `FILTER_GENDER`

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: united two event listeners into a common one `FILTER_SORT`

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: changed `indexOf(searchRequest) == -1` to
`includes(searchRequest)`

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: Reworked. Necessary items are rerendered completely

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: added `mailto:`

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+    filterByGender();
+    displayFriends();
+}
+
+FILTER_GENDER.addEventListener('change', (event) => {
+    let target = event.target;
+    changeFilterStatus(target);
+})
+FILTER_SORT.addEventListener('change', (event) => {
+    let target = event.target;
```
zonzujiro, line 118:
```suggestion
    const { value } = event.target;
```

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+    sortFriends(target.value);
+})
+
+function fetchUsers() {
+    let targetUrl = "https://randomuser.me/api/?results=30&nat=au,ca,ch,de,dk,es,fr,gb,ie,no,nl,nz,us"
+
+    fetch(targetUrl, { metod: "get" })
```
zonzujiro, line 125: No need for `method`. `fetch` by default will send `get` request

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+        .then((response) => {
+            let json = response.json();
+            if (response.ok) {
```
zonzujiro, line 128: Do you really need parsed response if status not ok?

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+                return json;
+            }
+            else {
+                return error;
+            }
+        }).then((json) => {
+            saveUsers(json.results);
+        })
```
zonzujiro, line 136: Single responsibility - your functions should do only one job. As an example function in which you fetch users should only fetch them and not render, transform or process them in other ways. All these things should be done in another place, outside your function. The same applied to the sort function, which usually does all types of sorting 😉

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+    searchedFriends = [];
+    selectedFriends.forEach(item => {
+        if (item.name.toUpperCase().includes(searchRequest)) {
+            searchedFriends.push(item)
+        }
+        selectedFriends = searchedFriends;
+    })
+}
+
+function sortFriends(value) {
```
zonzujiro, line 80: You also filter here, so the name is confusing.

Move filtering outside this function

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+const FILTER_SORT = document.getElementById("filter-sort");
+const FILTER_GENDER = document.getElementById("filter-gender");
+let friends = [];
+let gender = {
+    male: true,
+    female: true
+}
+
+let selectedFriends = [...friends];
+function filterByGender() {
```
zonzujiro, line 18: If you will pass `value` of selected radio button, you will have a possibility to write:

```jsx
selectedFriends = friends.filter(item => item.gender == value)
```

Instead of all these checks. And because of this, you will remove your `let gender` as redundant, because the state of selected filters kept in HTML right now ;)

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
+const GRID = document.getElementById("allFriends");
+const CHECKBOX_MAN = document.getElementById("gender-male");
+const CHECKBOX_WOMAN = document.getElementById("gender-female");
+const SORT_ALPHABET_ASCEND = document.getElementById("a-z");
+const SORT_ALPHABET_DESCEND = document.getElementById("z-a");
+const SEARCH_INPUT = document.getElementById('search');
+const SORT_AGE_ASCEND = document.getElementById("0-9");
+const SORT_AGE_DESCEND = document.getElementById("9-0");
+const FILTER_SORT = document.getElementById("filter-sort");
+const FILTER_GENDER = document.getElementById("filter-gender");
```
zonzujiro, line 10: Don’t overuse global variables, keep data that you are using closest as possible to the place where you using it.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+    friends.sort((a, b) => {
+        if (order == "descending") {
+            let c = a;
+            a = b;
+            b = c;
+        }
+        if (a[criterion] > b[criterion]) return 1;
+        if (a[criterion] == b[criterion]) return 0;
+        if (a[criterion] < b[criterion]) return -1;
```
zonzujiro, line 107: Please, go thru other PRs and find inspiration for the simplest implementation of the sorting function.
As an example, you don't need `order`.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+
+let selectedFriends = [...friends];
+function filterByGender() {
+    if (gender.male == true && gender.female == true)
+        selectedFriends = friends;
+    else if (gender.male == true && gender.female == false)
+        selectedFriends = friends.filter(item => item.gender == "male")
+    else if (gender.male == false && gender.female == true)
+        selectedFriends = friends.filter(item => item.gender == "female")
+    search();
```
zonzujiro, line 25: Same as with your `sortFunction`. You should not call search from `filterByGender` it creates unpredictable side effects.

You should move this call outside of this function.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+                gender.male = true;
+                CHECKBOX_MAN.click();
+            }
+        }
+        gender[checkbox.value] = false;
+    }
+    filterByGender();
+    displayFriends();
+}
+function searchEvent() {
```
zonzujiro, line 64: Function names should be verbs

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+    filterByGender();
+    displayFriends();
+}
+
+function search() {
```
zonzujiro, line 69: Verb

zonzujiro (changes requested):

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: I agree with you that the use of `radio` in this case simplifies the program. 
I only need three radio buttons: "Man", "Woman" and "All". 
But I decided to use this solution because there are only two checkboxes and you can see which ones are selected at once, and it can be scaled and used in tasks with many filters.

submissions/Harnytskyi/friends-app/js/script.js
```diff
@@ -0,0 +1,163 @@
...
+            if (gender.male == false) {
+                gender.male = true;
+                CHECKBOX_MAN.click();
+            }
+        }
+        gender[checkbox.value] = false;
+    }
+    filterByGender();
+    displayFriends();
+}
```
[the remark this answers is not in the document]
Harnytskyi: My decision is that at least one option will always be shown. I decided to use it here because if the user disabling the checkboxes "man" and "woman" at the same time, he will be shown a blank page, which makes no sense.

As for the "custom clicking" remark, I just didn't guess the call necessary functionality directly. I will fix it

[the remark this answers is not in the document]
Harnytskyi: now I have noticed that I can use two conditions in one `if`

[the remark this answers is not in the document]
zonzujiro: > My decision is that at least one option will always be shown. I decided to use it here because if the user disabling the checkboxes "man" and "woman" at the same time, he will be shown a blank page, which makes no sense.

Not sure what you mean, TBH :)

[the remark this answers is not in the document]
Harnytskyi: It seems that I did not understand you because your code does not work completely. I propose the following solution:
```suggestion
function changeFilterStatus(checkbox) {
    if (!checkbox.checked){
        if (checkbox.value == "male" && gender.female == false) {
                CHECKBOX_WOMAN.checked = !CHECKBOX_WOMAN.checked;
                changeFilterStatus(CHECKBOX_WOMAN);
        }
        else if (checkbox.value == "female" && gender.male == false) {
                CHECKBOX_MAN.checked = !CHECKBOX_MAN.checked;
                changeFilterStatus(CHECKBOX_MAN);
        }
    }
    gender[checkbox.value] = !gender[checkbox.value];
    filterByGender();
    displayFriends();
}
```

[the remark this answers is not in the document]
zonzujiro: IMO, looks like overengineering. `changeFilterStatus(CHECKBOX_WOMAN);` - why do you need recursion here?

Also, when I click on `male` you will change the state of `female` checkbox? Wuuuut? :)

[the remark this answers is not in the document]
Harnytskyi: There are only two categories ("male" and "female"). 
If the user does not search in one category, he must search in another.
To do this, I change the status of another checkbox.
male-female-checkbox

Of course, I can make that no category be selected but I think this solution more interesting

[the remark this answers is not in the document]
zonzujiro: Ok, you have implemented your own custom controller for this. Now let's think from the user's perspective - is this controller behaves as expected or it's confusing? Usual behaviour for the checkbox is "when I click on `checkbox` it will change its own state". And not the state of the neighbour. So I believe it is confusing.

Also, you have two categories, but you anyway have 3 states:
1. female
1. male
1. all genders

And you can have "only one state selected".

So I see two ways:

1. Give the user a possibility to remove all checkboxes - which is bad for reasons you have mentioned.
1. Or implement controller which will show to the user that he can select 1 of 3 choices. Which is quite good for `radio` ;)

WDYT?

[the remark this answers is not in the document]
Harnytskyi: Thank you for the meaningful explanation. 
You have convinced me that from a UX my decision is inexpedient. 
I decided to choose the first way and give a possibility to remove all checkboxes.

submissions/Harnytskyi/friends-app/js/script.js
Harnytskyi: Does it have to look like this?
`.then((response) => response.json())`

submissions/Harnytskyi/friends-app/js/script.js
zonzujiro: Please, do not resolve conversations where you have shared your opinion and did not receive an answer. :)

First thing - you will not have a case where you will need to scale this functionality - requirements will never change.

Second, your solution is very imperative and not scalable. You will need to check the state of all checkboxes with your `if ... else`  logic which will lead to very concrete, not abstract, code. Such code very hard to maintain, because on every change you will add more and more `if ... else` lines. And this will lead to another changes, which also will lead to other changes.

So, if you *really* want to keep your checkboxes, find a more abstract way to do this.
Or, considering that requirements will never change, just use `radio` buttons.

zonzujiro (changes requested):

zonzujiro:
Good job!

PR #434, opened 2021-01-20
add friends-app

TaniaBondarenko:
Friends App

[Demo](https://taniabondarenko.github.io/Friends-App/)
[Code base](https://github.com/TaniaBondarenko/Friends-App/blob/main/app.js)

The code is submitted in a dedicated feature branch.
Please, review.

PR #452, opened 2021-02-14
Friends App

vakulinina:
### Friends App

[Demo](https://vakulinina.github.io/kottans-friends-app/)
[Code base](https://github.com/vakulinina/kottans-friends-app)

Please, review.


submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,173 @@
...
+        filter.nameSort = 'desc';
+        sortByNameButton.classList.remove('asc')
+        sortByNameButton.classList.add('desc')
+      } else {
+        filter.nameSort = 'asc';
+        sortByNameButton.classList.remove('desc')
+        sortByNameButton.classList.add('asc')
+      }
+      sortByAgeButton.classList.remove('asc', 'desc')
+      filter.ageSort = null;
```
zonzujiro, line 130: Move this code into separate functions

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,173 @@
+const minAge = 0;
+const maxAge = 99;
```
zonzujiro, line 2: `MIN_AGE` and `MAX_AGE`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,173 @@
...
+      break;
+    case 'age-to':
+      filter.ageTo = target.value || maxAge;
+      break;
+    case 'search-field':
+      filter.searchValue = target.value.toLowerCase();
+      break;
+    case 'gender-male':
+    case 'gender-female':
+    case 'gender-any':
```
zonzujiro, line 156: `default`?

zonzujiro (changes requested):

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,178 @@
...
+    filteredUsers = doSearch(filteredUsers);
+  }
+  renderUsers(filteredUsers)
+};
+
+const setNameSortType = () => {
+  if (filter.nameSort === 'asc') {
+    filter.nameSort = 'desc';
+    sortByNameButton.classList.remove('asc')
+    sortByNameButton.classList.add('desc')
```
zonzujiro, line 120: `classList.replace`

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,178 @@
...
+  } else {
+    filter.nameSort = 'asc';
```
zonzujiro, line 122: `asc` and `desc` should be in constants.

Also, please use full names, not shortcuts.

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,178 @@
...
+      break;
+    case 'age-to':
+      filter.ageTo = target.value || MAX_AGE;
+      break;
+    case 'search-field':
+      filter.searchValue = target.value.toLowerCase();
+      break;
+    default:
+      filter.gender = target.value;
+  }
```
zonzujiro, line 163: Try to replace `switch` with map.

Something like https://github.com/kottans/frontend-2021-homeworks/pull/434#discussion_r576145474

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,178 @@
...
+  filterUsers()
+};
+
+const initialize = async () => {
+  const filtersBar = document.querySelector('.filters');
+  await getUsers()
+  renderUsers(allUsers)
+  filtersBar.oninput = handleInput;
+  document.querySelector('.sort-buttons').onclick = handleInput;
+  document.querySelector('.filters-button').onclick = () => {
```
zonzujiro, line 173: Why not `addEventListener`?

zonzujiro (changes requested):

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,178 @@
...
+    if (!response.ok) {
+      throw new Error(response.status)
+    }
+    const json = await response.json();
+    allUsers = json.results;
+  } catch (error) {
+    const errorNotification = document.createElement('p');
+    errorNotification.classList.add('error-notification')
+    errorNotification.innerHTML = `Sorry, couldn't get the data <br> ${error}`;
+    document.querySelector('main').prepend(errorNotification)
```
zonzujiro, line 27: Creation and rendering of DOM elements should be in a separate function.

submissions/vakulinina/friends-app/main.js
vakulinina: Since I didn't see the difference in this case (i only need to handle one event for every element), I decided to stick with the event handlers, they just look prettier to me :) 
Is something wrong with them?

zonzujiro: Well... it is not something critical. But `onclick` will delete all other listeners and considered as old API.

So when you can choose between old API with drawbacks and new API without them, you should always prefer the last option ;)

vakulinina:
Thank you for the review! 
Please check the updates, I'm not sure if I did everything exactly the way you meant it though )

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,174 @@
...
+};
+
+const setNameSortOrder = () => {
+  if (filter.nameSort === ASCENDING_ORDER) {
+    filter.nameSort = DESCENDING_ORDER;
+    sortByNameButton.classList.replace(ASCENDING_ORDER, DESCENDING_ORDER)
+  } else {
+    filter.nameSort = ASCENDING_ORDER;
+    sortByNameButton.classList.remove(DESCENDING_ORDER)
+    sortByNameButton.classList.add(ASCENDING_ORDER)
```
zonzujiro, line 129: Khm

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,174 @@
...
+const handleInput = (target) => {
+  const filters = {
+    'name-sort': setNameSortOrder,
+    'age-sort': setAgeSortOrder,
+    'age-from': () => filter.ageFrom = target.value || MIN_AGE,
+    'age-to': () => filter.ageTo = target.value || MAX_AGE,
+    'search-field': () => filter.searchValue = target.value.toLowerCase()
+  };
+
+  filters[target.id] ? filters[target.id]() : filter.gender = target.value;
```
zonzujiro, line 157: Don't use ternary operators like this.

Also, instead of `id` you can use [`data-*`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*). Just because you need this data for utility purposes and for searching or styling.

submissions/vakulinina/friends-app/main.js
```diff
@@ -0,0 +1,174 @@
...
+  filterUsers();
+};
+
+const initialize = async () => {
+  const filtersBar = document.querySelector('.filters');
+  await getUsers()
+  renderUsers(allUsers)
+  filtersBar.addEventListener('input', ({ target }) => handleInput(target));
+  document.querySelector('.sort-buttons').addEventListener('click', ({ target }) => {
```
zonzujiro, line 166: Buttons can be inputs and then they will fire `change` event and you will have a possibility to remove check at the next line

zonzujiro (changes requested):

submissions/vakulinina/friends-app/main.js
vakulinina: I didn't replace this with `replace` because this block is also applied in a case when `sortByNameButton` has neither of classes. So 'replace' would return `false` and new class would not be added.

submissions/vakulinina/friends-app/main.js
vakulinina: I'm stuck here. Do you mean `<input type="button">`? How can I make it fire `change` event? It doesn't if I click it.

zonzujiro: As an example https://markheath.net/post/customize-radio-button-css

zonzujiro (approved):
LGTM

PR #459, opened 2021-02-18
Friends-app

VBystrov:
### Friends-app

[Demo](https://vbystrov.github.io/friends-app/)
[Code base](https://github.com/VBystrov/friends-app)

Please, review.

### DIFF
--- a/submissions/VBystrov/friends-app/js/index.js
+++ b/submissions/VBystrov/friends-app/js/index.js
@@ -1,0 +1,326 @@
+const endpoint = 'https://randomuser.me/api/';
+const includeFields = `inc=${[
+  'gender',
+  'name',
+  'location',
+  'picture',
+  'dob',
+  'email',
+].join(',')}`;
+const amountFriends = `results=${50}`;
+const options = `?${[includeFields, amountFriends].join('&')}`;
+const requestFriendsUrl = `${endpoint}${options}`;
+const initialSearchParameters = {
+  name: null,
+  age: null,
+  gender: 'all',
+  sorting: 'nameAZ',
+};
+
+class FriendsApp {
+  constructor() {
+    this.init(requestFriendsUrl);
+  }
+
+  init(url) {
+    fetch(url)
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error('Network response was not ok');
+        }
+        return response.json();
+      })
+      .then((friendsData) => {
+        const friendlist = new Friendlist(friendsData.results);
+        const search = new Search(friendlist);
+      })
+      .catch((error) => {
+        console.error(
+          'There has been a problem with your fetch operation:',
+          error
+        );
+      });
+  }
+}
+
+class Friendlist {
+  constructor(friendsData) {
+    this.container = document.getElementById('friendlist');
+    this.createCards(friendsData);
+  }
+
+  createCards(friendlistData) {
+    this.cards = [];
+    friendlistData.forEach((cardData) => {
+      this.cards.push(new Card(cardData));
+    });
+  }
+
+  applySearch(search) {
+    this.applyFilters(search);
+    this.applySorting(search.sorting);
+  }
+
+  applyFilters(search) {
+    this.filteredCards = [...this.cards];
+
+    if (search.age) {
+      this.filteredCards = this.filteredCards.filter(
+        (card) => card.age === search.age
+      );
+    }
+
+    if (search.name) {
+      this.filteredCards = this.filteredCards.filter((card) =>
+        `${card.firstName} ${card.lastName}`.includes(search.name)
+      );
+    }
+
+    if (search.gender !== 'all') {
+      this.filteredCards = this.filteredCards.filter(
+        (card) => card.gender === search.gender
+      );
+    }
+  }
+
+  applySorting(sortingOrder) {
+    let sortFunction;
+    switch (sortingOrder) {
+      case 'nameAZ':
+      case 'nameZA':
+        sortFunction = (function () {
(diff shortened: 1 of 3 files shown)


submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,326 @@
...
+      this.cards.push(new Card(cardData));
+    });
+  }
+
+  applySearch(search) {
+    this.applyFilters(search);
+    this.applySorting(search.sorting);
+  }
+
+  applyFilters(search) {
```
zonzujiro, line 64: `search` quite opaque naming, TBH. Looks more like a state of some controllers, etc.

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,326 @@
...
+            let order = orderSign;
+            if (fullName1 < fullName2) {
+              return order * -1;
+            }
+            if (fullName1 > fullName2) {
+              return order * 1;
+            }
+            if (fullName1 === fullName2) {
+              return 0;
+            }
```
zonzujiro, line 105: Move this function into some variables.

And then map them.

Also, no need in part with `order`. You can do it simpler by changing the order of passed arguments.

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,326 @@
...
+    this.inputAge.addEventListener('input', this.changeSearchAge.bind(this));
+    this.genderContainer.addEventListener(
+      'click',
+      this.changeSearchGender.bind(this)
+    );
+  }
+
+  parseUrl() {
+    const searchParameters = { ...initialSearchParameters };
+    this.url = new URL(document.location.href);
```
zonzujiro, line 161: 👍

zonzujiro (changes requested):
Minor comments :)

submissions/VBystrov/friends-app/js/index.js
VBystrov: I can`t change order of passed arguments because i pass sortFunction to Array.sort(), so I need to change passed function. 

> Move this function into some variables.

> And then map them.

I don`t quite understand, what do you mean, but I tried to improve this part. 
I changed my mind, maybe its not worth to have one function that checks which is order every time when comparing array elements, but better to have two simpler but similar functions. What do you think?

VBystrov:
Changed name search to case insensitive and little improvement layout for small screens.

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,337 @@
...
+          if (fullName1 < fullName2) {
+            return -1;
+          }
+          if (fullName1 > fullName2) {
+            return 1;
+          }
+          if (fullName1 === fullName2) {
+            return 0;
+          }
+        };
```
zonzujiro, line 101: This function should be declared outside `applySorting`.

Separate responsibilities - `applySorting` should not know anything about the internals of the sorter function.

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,337 @@
...
```
zonzujiro, line 117: Same

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    if (searchParameters.gender !== 'all') {
+      this.filteredCards = this.filteredCards.filter(
+        (card) => card.gender === searchParameters.gender
+      );
+    }
+  }
+
+  applySorting(sortingOrder) {
```
zonzujiro, line 85: In general, this function looks redundant. You can have mapped sorters. Something like:

```jsx
const ageSorters = {
  ascending: ...
}
```
And then, you will just call them using `name` and `value` from inputs. You can check other PRs ;)

zonzujiro (changes requested):

submissions/VBystrov/friends-app/js/index.js
zonzujiro: Left comment. You did a good job. But I think we have some place for improvements :)

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,328 @@
...
+    this.container.addEventListener('click', this.changeSorting.bind(this));
+    this.inputName.addEventListener('input', this.changeSearchName.bind(this));
+    this.inputAge.addEventListener('input', this.changeSearchAge.bind(this));
+    this.genderContainer.addEventListener(
+      'click',
+      this.changeSearchGender.bind(this)
+    );
+  }
+
+  parseUrl() {
```
zonzujiro, line 124: Move URL-related stuff into a separate entity. I don't think that `Search` should be responsible for handling interactions with URL

submissions/VBystrov/friends-app/js/index.js
```diff
@@ -0,0 +1,328 @@
...
+        <p class="friendcard__age">${this.age}</p>
+      </div>
+    </div>
+  </div>`;
+  }
+}
+
+class Sorters {
+  constructor() {
+    this.nameAZ = function (card1, card2) {
```
zonzujiro, line 291: Don’t use names as arr, array, data or etc. They to generic. Don’t use things like sortingByABC and sortingByZYX - it’s unclear and just weird. user1 and user2 - don't use numbers in names. Just don't :) friendsArr also a bad idea - you show data type in the variable name and in JS it’s redundant

zonzujiro (changes requested):

zonzujiro (approved):

PR #461, opened 2021-02-19
Friends App is ready for review

m-ruslan:
### Friends App

[Demo](https://m-ruslan.github.io/kottans-freinds-app/)
[Code base](https://github.com/m-ruslan/kottans-freinds-app)

Please, review.

### DIFF
--- a/submissions/m-ruslan/friends-app/src/list.js
+++ b/submissions/m-ruslan/friends-app/src/list.js
@@ -1,0 +1,46 @@
+const makeUserHtml = (user) => {
+  const li = document.createElement("li");
+  li.classList.add("user-card");
+
+  const name = `${user.name.first} ${user.name.last}`;
+  const email = `${user.email}`;
+  const city = `${user.location.city}`;
+
+  const maxLengthOfName = 15;
+  const maxLengthOfEmail = 25;
+  const maxLengthOfCity = 18;
+
+  const ending = "...";
+
+  const truncatedName = truncateField(name, maxLengthOfName, ending);
+  const truncatedEmail = truncateField(email, maxLengthOfEmail, ending);
+  const truncatedCity = truncateField(city, maxLengthOfCity, ending);
+
+  li.innerHTML = `
+  <h3 class="user-card__user-name">${truncatedName}</h3>
+  <img class="user-card__user-photo" src="${user.picture.large}" alt="photo of ${name}"}>
+  <ul class="user-card__info-list">
+    <li >Gender: ${user.gender}</li>
+    <li>Age: ${user.dob.age}</li>
+    <li class="user-card__user-email"><a href="mailto:${user.email}" tabindex="-1">${truncatedEmail}</a></li>
+    <li class="user-card__user-tel"><a href="tel:${user.cell}" tabindex="-1">${user.cell}</a></li>
+    <li>${truncatedCity}</li>
+  </ul>
+  `;
+
+  return li;
+};
+
+const truncateField = (field, maxLenghOfField, ending) => {
+  return field.length > maxLenghOfField
+    ? field.substring(0, maxLenghOfField - ending.length) + ending
+    : field;
+};
+
+export const makeUsersList = (users) => {
+  const list = document.createElement("ul");
+  list.classList.add("users-list");
+  const domElements = users.map(makeUserHtml);
+  list.append(...domElements);
+  return list;
+};
--- a/submissions/m-ruslan/friends-app/src/list.js
+++ b/submissions/m-ruslan/friends-app/src/list.js
@@ -2,29 +2,17 @@ const makeUserHtml = (user) => {
   const li = document.createElement("li");
   li.classList.add("user-card");
 
-  const name = `${user.name.first} ${user.name.last}`;
-  const email = `${user.email}`;
-  const city = `${user.location.city}`;
-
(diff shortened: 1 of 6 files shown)


submissions/m-ruslan/friends-app/src/list.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const maxLengthOfName = 15;
+  const maxLengthOfEmail = 25;
+  const maxLengthOfCity = 18;
+
+  const ending = "...";
+
+  const truncatedName = truncateField(name, maxLengthOfName, ending);
+  const truncatedEmail = truncateField(email, maxLengthOfEmail, ending);
+  const truncatedCity = truncateField(city, maxLengthOfCity, ending);
```
zonzujiro, line 17: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow ;)

Please, use CSS instead

submissions/m-ruslan/friends-app/src/sorter.js
```diff
@@ -0,0 +1,147 @@
...
+  filterGenderBy: "",
+  searchByName: "",
+  ageFrom: "",
+  ageTo: "",
+};
+
+let drawResult = null;
+
+const getDatasetValue = (elements, key) => {
+  const checked = Array.from(elements).find(({ checked }) => checked);
```
zonzujiro, line 18: You can replace this with a query selector.

And when you will do, you probably will understand that there is no need for a separate function.

submissions/m-ruslan/friends-app/src/sorter.js
```diff
@@ -0,0 +1,147 @@
...
+  );
+  state.sortOrder = selectedOrder;
+  state.sortBy = selectedSorting;
+  state.filterGenderBy = filteredGender;
+
+  updateUsersList(users);
+  drawResult(filtered);
+};
+
+const nameInputListener = ({ currentTarget }) => {
```
zonzujiro, line 108: Verb :)

submissions/m-ruslan/friends-app/src/sorter.js
```diff
@@ -0,0 +1,147 @@
...
+  const searchByName = currentTarget.value.toLowerCase();
+  state.searchByName = searchByName;
+  updateUsersList(users);
+  drawResult(filtered);
+};
+
+const ageInputListener = ({ currentTarget }) => {
+  if (currentTarget.id === "search-by-age-from") {
```
zonzujiro, line 116: If you will use [`data-*`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) instead of id, then you will have a possibility to write something like this:

```jsx
state[currentTarget.dataset.ageType] = currentTarget.value
```

Datasets will convert `kebab-case` into `camelCase`. This is not very obvious, but fun :D

submissions/m-ruslan/friends-app/src/sorter.js
```diff
@@ -0,0 +1,147 @@
...
+};
+
+const findUsersByName = (users) => {
+  if (state.searchByName === "") return users;
+
+  return users.filter(
+    (user) =>
+      `${user.name.first} ${user.name.last}`
+        .toLowerCase()
+        .indexOf(state.searchByName) !== -1
```
zonzujiro, line 67: `String#includes`

zonzujiro (changes requested):
Looks good in general. Please check comments

m-ruslan:
@zonzujiro thank you for the review. It seems that I've fixed all issues you've written about. Could you please check it one more time?

submissions/m-ruslan/friends-app/src/list.js
```diff
@@ -0,0 +1,34 @@
...
+  </ul>
+  `;
+
+  return li;
+};
+
+const truncateField = (field, maxLenghOfField, ending) => {
+  return field.length > maxLenghOfField
+    ? field.substring(0, maxLenghOfField - ending.length) + ending
+    : field;
```
zonzujiro, line 25: Looks like not used variable

submissions/m-ruslan/friends-app/src/sorter.js
```diff
@@ -0,0 +1,143 @@
...
+
+let drawResult = null;
+
+const getValueOrDatasetValue = (parentElement, inputGroupName, datasetKey) => {
+  const checked = parentElement.querySelector(
+    `input[name=${inputGroupName}]:checked`
+  );
+  if (checked === null) return "";
+  return datasetKey === undefined ? checked.value : checked.dataset[datasetKey];
+};
```
zonzujiro, line 23: IMO, you should know when you will get `value` and when `data-*`. For now, this code looks like "I don't know what I want, just get me something" :)

You should avoid such stuff. because they *can* create unclear data flow.

zonzujiro (approved):
LGTM.

Comments FYI

submissions/m-ruslan/friends-app/src/sorter.js
m-ruslan: @zonzujiro thank you, I will take it into account in future

submissions/m-ruslan/friends-app/src/list.js
m-ruslan: @zonzujiro thank you, you are right)

PR #467, opened 2021-04-04
Create FUNDING.yml

suchov:
Think about creating a funding yml

_Upd_: removed default PR comment contents

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,7 +1,7 @@
 # These are supported funding model platforms
 
-github: # [suchov] Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
-patreon: # Replace with a single Patreon username
+github: [suchov] # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: kottans # Replace with a single Patreon username
 open_collective: # Replace with a single Open Collective username
 ko_fi: # Replace with a single Ko-fi username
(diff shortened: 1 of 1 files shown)


.github/FUNDING.yml
```diff
@@ -0,0 +1,12 @@
+# These are supported funding model platforms
+
+github: # [suchov] Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
```
suchov, line 3: added my surname just to make the PR valid, we can think about the options that we want to put

suchov:
Just want to suggest us to think about adding the Funding.yml to our most popular repos...

.github/FUNDING.yml
OleksiyRudenko, line 3:
```suggestion
github: [suchov] # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
```

Uncommented a list of names.

How does a user know if they are sponsors-enabled? Anyway, I think one or two would be sufficient

.github/FUNDING.yml
```diff
@@ -0,0 +1,12 @@
...
+patreon: # Replace with a single Patreon username
```
OleksiyRudenko, line 4:
```suggestion
patreon: kottans # Replace with a single Patreon username
```

OleksiyRudenko (changes requested):
LGTM in general. Just need to finalize the lists.

(Boy, that many funding entry point services around. Of those only heard of github, patreon and [opencollective](https://opencollective.com/pricing))

zonzujiro (approved):

